Issue #639, opened 2023-04-18
Improve Unlock Wallet User Experience


piotrwitek:
On the vault unlock page when the user successfully entered the password we should add an unlock animation, this would minimize the perceived delay needed to decrypt the vault.

Design: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Wallet-%E2%80%93-Extension?node-id=5539%3A12166&t=kMDtHazTXP1H6vlH-1
Animation file: https://lottiefiles.com/63686-lock-animation

adammake:
Added to Jira https://make-software.atlassian.net/browse/WALLET-50

adammake:
Need animation positioning and playback details from design team.

PR #640, opened 2023-04-18
feat: add unlock wallet animation

piotrwitek:

## Description

Added animations lib and Lottie dependency
Added new unlock animation in the unlock vault view

## Linked tickets

#639 

## Checklist


src/apps/popup/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'none'; object-src 'none'; base-uri 'none'; form-action 'none'; script-src 'self'; style-src 'unsafe-inline'; img-src https: data:; connect-src https://event-store-api-clarity-testnet.make.services https://event-store-api-clarity-mainnet.make.services"
+      content="default-src 'none'; object-src 'none'; base-uri 'none'; form-action 'none'; script-src 'self'; style-src 'unsafe-inline'; img-src https: data:; connect-src http: ws:"
```
ost-ptk, line 7: this is just for a test?

piotrwitek: good catch, yes I'll rollback that

src/libs/layout/unlock-vault/index.tsx
```diff
@@ -150,6 +153,32 @@ export function UnlockVaultPageContent() {
...
   }
 
+  if (isLoading) {
+    return (
+      <ContentContainer>
+        <CenteredFlexColumn style={{}}>
+          <Player
+            speed={1}
+            autoplay
+            keepLastFrame
```
ost-ptk, line 163: What do you think if we add `loop` props?
If the unlocking process is long, it will animate all time. Now it only animates once and stops.
Or this will decrease unlocking performance?

piotrwitek: It'll not have a performance impact, but It depends on the animation itself if it will look good when looped, I'm still waiting for the final animation but it's a valid idea :)

ost-ptk: I set the unlocking timeout for 10 sec and it looks weird when the animation stops and nothing changes

piotrwitek: It shouldn't take more than 3 sec even on older laptops. There will be a low percentage of pretty obsolete machines but there is nothing we can do about that.

Issue #636, opened 2023-04-18
Identicon/avatar UI improvements


adammake:
- The account image should be rounded.
- Texts on the buttons are too thin. Make them according to design.
- Connected (on account card) should be centred in the green container, but now it’s not.

![image.png](https://images.zenhubusercontent.com/61e1ac21d1fded72df402cc1/4e27cec4-1faa-48b2-8df7-6d18d018b0ea)

https://makellc.slack.com/archives/C026E4N640N/p1681681736336879?thread_ts=1681681444.886859&cid=C026E4N640N

ost-ptk:
> * Texts on the buttons are too thin. Make them according to design.

I checked on 4 browsers (Chrome, Firefox, Safari, Edge), and in each, the buttons are bold as in Figma

@adammake

karatnyk:
Moving to 'Done' as successfully implemented in build 1.3.0

PR #641, opened 2023-04-19
feature: add ui improvements for home page

ost-ptk:

## Description

- The account image is rounded.
- The connected status (on account card) is centered in the container.

## Linked tickets

#636 

## Checklist

- [ ] If the PR adds any new text to the UI, make sure they are localized

- [ ] When this PR affects architecture changes wait for review from Piotr before merging

image


piotrwitek (approved):

src/libs/layout/containers.ts
```diff
@@ -64,7 +67,7 @@ const getGapSize = (gap?: SpacingSize) => {
       return '8px';
     case SpacingSize.Medium:
       return '12px';
-    case SpacingSize.Big:
+    case SpacingSize.ExtraLarge:
```
piotrwitek: I think this should be Large in the current order, leave `ExtraLarge` for an additional bigger size

src/libs/ui/components/list/list.tsx
```diff
@@ -77,12 +79,14 @@ export function List<ListRow extends ListRowBase>({
   renderFooter,
   headerLabel,
   headerAction,
-  marginLeftForItemSeparatorLine
+  marginLeftForItemSeparatorLine,
+  headerLabelSpaceSize = SpacingSize.ExtraLarge,
```
piotrwitek: This should be renamed to top, as there is functionally no difference from existing `top` prop on the existing `VerticalSpaceContainer`, consistency should be the most valuable priority.

src/libs/ui/components/list/list.tsx
```diff
@@ -77,12 +79,14 @@ export function List<ListRow extends ListRowBase>({
...
+  contentSpaceSize = SpacingSize.ExtraLarge
```
piotrwitek: this change seems unnecessary as all props are changed to use the same value Small, so it seems to me like a simple refactor instead of a component design change.

piotrwitek (changes requested):
@ost-ptk review is ready

piotrwitek:
@ost-ptk I don't know what happened with that PR but it changed probably because of a forced pushed, could you apply comments to the relevant PR so I don't have to do the review again, please?

ost-ptk:
> @ost-ptk I don't know what happened with that PR but it changed content probably because of reverted commits, could you apply comments done here to the relevant PR so I don't have to do the review process again, please?

I reverted that changes and create a separate PR as these changes belong to another ticket
https://github.com/make-software/casper-wallet/pull/645

src/libs/ui/components/list/list.tsx
ost-ptk: Not all. I only passed these props, `contentSpaceSize` and `headerLabelSpaceSize`, a few times where it's needed by design. These default values are for all other cases when we didn't pass these props.

src/libs/ui/components/list/list.tsx
piotrwitek: cool make sense 👍

src/libs/ui/components/list/list.tsx
piotrwitek: In that case, seems like we have most 3 variations, in my opinion adding variations makes more sense than adding unnecessary complexity to the component API

ost-ptk:
@piotrwitek Changes that are related to the List component, were merged in this [PR](https://github.com/make-software/casper-wallet/pull/645)

piotrwitek:
> > @ost-ptk I don't know what happened with that PR but it changed content probably because of reverted commits, could you apply comments done here to the relevant PR so I don't have to do the review process again, please?
> 
> I reverted that changes and create a separate PR as these changes belong to another ticket #645

Yes I saw it, thanks for letting me know

piotrwitek (approved):

Issue #630, opened 2023-04-13
Implement handling for the SDK getter methods to provide handling for cases like "wallet is locked" or "site is unapproved"


piotrwitek:
Currently, when calling SDK getter methods it's impossible to distinguish if the wallet is locked or the site is unapproved making it difficult to handle some use cases by integrators.
Implement handling for the SDK getter methods to provide handling for cases like "wallet is locked" or "site is unapproved"

piotrwitek:
Resolved by #647

Issue #629, opened 2023-04-13
Get active key SDK method is returning public address when not connected


piotrwitek:
Get active key should return undefined when not connected to site, currently it's returning active key even when it's not approved

@karatnyk This can be tested on the playground app by clicking on a new get active key button when the wallet is disconnected/locked.

PR #647, opened 2023-04-20
feature: improved SDK getter methods to correctly handle state access and provide better DX by providing extended error handling in promises

piotrwitek:

## Description

- Updated CasperWalletState to return undefined instead of null value when cannot access `isConnected` or `activeKey` state values because the wallet is in the locked state which is more appropriate to represent a non-accessible state.
- Extended `isConnected` and `getActivePublicKey` methods to throw errors when the wallet is locked and when the key is not approved to be shared with the asking site, this will help integrators to better handle edge cases around those scenarios.
- Added SdkError class that will provide a consistent error object to consumers containing error message and err code for easy identification.

## Linked tickets

#629
#630 

## Checklist

- [ ] Include a screenshot or recording if implementing significant UI or user flow change


src/apps/popup/hooks/use-account-actions-with-events.ts
```diff
@@ -56,16 +56,15 @@ export function useAccountManager() {
   const isLocked = useSelector(selectVaultIsLocked);
   const activeAccount = useSelector(selectVaultActiveAccount);
   const accounts = useSelector(selectVaultAccounts);
-  const accountNamesByOriginDict = useSelector(
-    selectVaultAccountNamesByOriginDict
-  );
+  const accountNamesByOriginDict = useSelector(selectAccountNamesByOriginDict);
```
ost-ptk, line 59: We can do this here and remove it in other places
```suggestion
  const accountNamesByOriginDict = useSelector(selectAccountNamesByOriginDict) || [];
```

ost-ptk (approved):

src/apps/popup/hooks/use-account-actions-with-events.ts
piotrwitek: No, it's a dictionary, not an array.

Issue #615, opened 2023-04-04
Casper Signer | 'Signature request' screen | 'Signature request' screen becomes empty after CW gets locked


karatnyk:
**Steps to reproduce:**
1. Initiate a token transaction on cspr.live website
2. On 'Sign' step click on 'Sign with Casper Signer' button
3. Observe that 'Signature Request' window appeared
4. Wait for CW to get locked due to timeout
5. Unlock the Wallet

**Expected result:** 'Signature request' screen is displayed exactly the same prior to timeout lock
**Actual result:** 'Signature request' screen is empty
![chrome_iDO52BzzGo.png](https://images.zenhubusercontent.com/6241d5bd38a85b3df1624280/cb9d8b13-9284-4f22-a496-7cf1eb297455)
![chrome_PKUtqraUGl.gif](https://images.zenhubusercontent.com/6241d5bd38a85b3df1624280/e63d145e-068c-4c04-ad07-dc9166abb11a)

PR #652, opened 2023-04-24
fix: deploy details is lost when wallet is locked during sign flow causing empty screen when unlocked

piotrwitek:

## Description

- Moved deploys data so it's encrypted as part of the vault state on lock, so it can be recovered after unlocking and fixing the empty state bug.

- Removed obsolete deploys when a new deploy is received from SDK so the state is free from outdated data and prevents memory leaks.

## Linked tickets

#615

## Checklist


ost-ptk (approved):

piotrwitek:
Will be merged after 1.3.0 release as it's targeted at 1.3.1

PR #1342, opened 2026-06-12
feat: EIP-712 signature requests (WALLET-1251)

ost-ptk:
## WALLET-1251 — EIP-712 signature requests

Jira: https://make-software.atlassian.net/browse/WALLET-1251 (epic WALLET-1250, release CW 2.5.0)

Adds the EIP-712 typed-data signing flow: a dapp calls `provider.signTypedData(params, publicKey)`, the user sees a window with Domain/Message blocks and signs with a software key, and the dapp receives the signature. All domain logic lives in `casper-wallet-core` (`eip712Repository`); the wallet renders the prepared `domainRows`/`messageRows` and signs.

### How it works
```
dapp.signTypedData(params, pubKey)
  → content sdk.ts → signTypedDataRequest
  → background: dispatch typedData into redux (vault slice) + openWindow(SignatureRequestEip712)
  → SignEip712Page: reads typedData from redux → useFetchDataForEip712Request → eip712Repository.prepareSignatureRequest
  → renders Domain/Message rows; handleSign → eip712Repository.signTypedData → signTypedDataResponse
  → content/index.ts proxies the response to the page → dapp receives SignTypedDataResult
```

### Changes

**Dependencies / compilation**
- `casper-js-sdk` 5.0.10 → 5.0.12 (deduplicates with the core copy), `date-fns` 2.30 → ^4 (fixes the `enUS` import in core)
- Deep type imports `casper-js-sdk/dist/...` moved to the package root (5.0.12 no longer ships `.js` at those paths)

**SDK / protocol** (`src/content/`)
- `sdk.ts`: `signTypedData(params, signingPublicKeyHex)` method
- `sdk-method.ts`: `signTypedDataRequest`/`Response`/`Error` actions
- `sdk-types.ts`: `SignTypedDataParams`, `SignTypedDataResult`, `EIP712HashArtifacts` (contract aligned with the playground)
- `index.ts`: whitelist response/error in the content proxy (otherwise the call hangs)

**Background / redux**
- vault slice: `eip712ById` field + `eip712PayloadReceived` + `selectEip712JsonById` (follows the deploy `jsonById` pattern)
- `WindowApp.SignatureRequestEip712` + `RouterPath.SignEip712`
- `signTypedDataRequest` handler in `background/index.ts`

**UI** (`src/apps/signature-request/pages/sign-eip712/`)
- `useFetchDataForEip712Request` hook
- `Eip712DisplayRow` — presentation → component mapper (`AccountInfoRow` / `ContractInfoRow` / `Hash` / text), all rows lay out label↔value via `SpaceBetweenFlexRow`
- `Eip712NetworkRow` — Network with icon (reuses `NetworkIconColor` + the resolver from sign-transaction)
- `SignEip712Content` — Domain + Message blocks
- `SignEip712Page` — fetch + sign + cancel, route wired up

**Ledger guard**
- Ledger accounts are rejected on the EIP-712 signing page (mirrors the existing decrypt-message pattern): an early `signingAccount.hardware === HardwareWalletType.Ledger` check sends `signTypedDataError` back to the dapp and throws, so the user sees the ErrorBoundary screen. New `ErrorMessages.signTypedData.LEDGER_NOT_SUPPORTED` constant.

### Testing
- `npm run ci-check` — green (format + lint + tsc + test)
- `npm run build:chrome` — successful
- Manual e2e via the Casper Wallet playground: sign/cancel, Domain/Message rendering, `ContractInfoRow` for a real package hash, and the Ledger-account rejection path

### Notes
- **Ledger**: EIP-712 typed-data signing is intentionally not supported on Ledger hardware wallets (confirmed with the feature owner). Software-key signing only; Ledger accounts are rejected before reaching the signing step.
- **Package hash when the package is unresolved**: when core cannot resolve a contract package, the row arrives as `presentation: 'string'` and the hash is shown in full. Recommendation for core — force `presentation: 'hash'` for `contract_package_hash`. With a real package hash the row renders as `ContractInfoRow` correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/background/redux/vault/actions.ts
```diff
@@ -69,6 +69,11 @@ export const deployPayloadReceived = createAction('DEPLOY_PAYLOAD_RECEIVED')<{
   json: string;
 }>();
 
+export const eip712PayloadReceived = createAction('EIP712_PAYLOAD_RECEIVED')<{
```
Comp0te, line 72: **`eip712ById` is never persisted to the vault cipher → request can hang after a service-worker restart.**

In `src/background/redux/sagas/vault-sagas.ts` the `updateVaultCipher` trigger list includes `deployPayloadReceived` but not `eip712PayloadReceived`. `updateVaultCipher` encrypts the whole vault (including both `jsonById` and `eip712ById`) into the only thing persisted to storage.

Under MV3 the background service worker can be terminated between this action being dispatched and the user signing in the popup. On restart the store rehydrates from the stale cipher, `eip712ById[requestId]` is gone, `typedData` becomes `undefined`, the Sign button stays disabled and no `signTypedDataError` is sent — the dapp's `signTypedData()` promise hangs. Deploys avoid this precisely because `deployPayloadReceived` is in the trigger list.

Suggestion: add `getType(eip712PayloadReceived)` to the `updateVaultCipher` `takeLatest` array, next to `deployPayloadReceived`.

src/apps/signature-request/pages/sign-eip712/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+    // EIP-712 supports software-key signing only; Ledger accounts are rejected earlier.
+    const publicKey = PublicKey.fromHex(signingAccount.publicKey);
+    const privateKey = PrivateKey.fromHex(
+      getPrivateKeyHexFromSecretKey(
+        Conversions.base64to16(signingAccount.secretKey)
+      ),
+      publicKey.cryptoAlg
+    );
+
+    const result = eip712Repository.signTypedData({
```
Comp0te: **`handleSign` performs key reconstruction + signing with no try/catch — a throw leaves the dapp hanging.**

`PublicKey.fromHex` / `PrivateKey.fromHex` / `Conversions.base64to16` / `eip712Repository.signTypedData` all run before `responseSentRef.current = true`, and none are wrapped. If any of them throws (malformed/empty secret key, unsupported crypto alg, or a typed-data edge that passed `prepareSignatureRequest` but fails at sign time), no `signTypedDataResponse`/`signTypedDataError` is sent and the window isn't closed via this path, so the dapp promise hangs until timeout.

This is inconsistent with `useFetchDataForEip712Request`, which *does* catch and send `signTypedDataError`. Suggestion: wrap the key construction + sign in try/catch and send `sdkMethod.signTypedDataError` on failure (mirroring the missing-account / Ledger guard paths).

src/apps/signature-request/pages/sign-eip712/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+    shallowEqual
+  );
+
+  const { typedData, options } = useMemo<{
+    typedData?: IEIP712TypedData;
+    options?: IEIP712SignTypedDataOptions;
+  }>(() => {
+    const raw = eip712JsonById[requestId];
+
+    return raw ? JSON.parse(raw) : {};
```
Comp0te: **Unguarded `JSON.parse` in this `useMemo` throws during render before any error path is wired up.**

This runs before `useFetchDataForEip712Request` (whose catch sends `signTypedDataError`) and before the `beforeunload`→`handleCancel` effect is registered. If the stored payload is malformed/truncated JSON, `JSON.parse` throws during render, the component crashes to the error boundary, and no SDK response is ever sent — the dapp hangs. The sibling pages read a plain string param and never parse, so this surface is unique here.

Suggestion: wrap the parse in try/catch and send `signTypedDataError` (or treat as missing payload) instead of letting it throw.

src/content/sdk.ts
```diff
@@ -202,6 +203,25 @@ export const CasperWalletProvider = (options?: CasperWalletProviderOptions) => {
         };
       });
     },
+    signTypedData: (
```
Comp0te, line 206: **`signTypedData` is not advertised in `CasperWalletSupports`.**

No `signTypedData` member was added to the `CasperWalletSupports` enum (`src/content/sdk-types.ts`) and the supports arrays in `getActiveAccountSupports` (`src/utils.ts`) were not updated. Dapps that feature-detect via `getActivePublicKeySupports()` before calling — the established pattern for every other method — will never see EIP-712 and will treat it as unsupported.

Suggestion: add a `signTypedData` capability and include it in the non-Ledger branch of `getActiveAccountSupports` (software-only, consistent with the Ledger rejection).

src/apps/signature-request/pages/sign-eip712/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+          }
+        />
+      )}
+      renderContent={() => {
+        if (shouldTryToConnectAccount) {
+          return (
+            <ApproveConnectionContent
+              account={signingAccount}
+              accountLiquidBalance={
+                accountsBalances?.[signingAccountHash].liquidBalance ?? '0'
```
Comp0te: **Optional chaining stops at the dict — `TypeError` if the balance entry is missing.**

```ts
accountsBalances?.[signingAccountHash].liquidBalance ?? '0'
```
If `accountsBalances` resolves to a defined object that has no entry for `signingAccountHash` (account with no on-chain activity / partial API response), `.liquidBalance` throws and the `?? '0'` fallback never applies. This is copied verbatim from `sign-message`, so it's a pre-existing pattern rather than a regression — but worth fixing here: `accountsBalances?.[signingAccountHash]?.liquidBalance ?? '0'`.

src/apps/signature-request/pages/sign-eip712/index.tsx
```diff
@@ -0,0 +1,305 @@
...
+    );
+    sendSdkResponseToSpecificTab(
+      sdkMethod.signTypedDataError(error, { requestId }),
+      requestTabId
+    );
+    throw error;
+  }
+
+  // Ledger hardware wallets do not support EIP-712 typed data signing.
+  if (signingAccount.hardware === HardwareWalletType.Ledger) {
```
Comp0te, line 134: **Eligibility guard only rejects Ledger — watch-only accounts (empty `secretKey`, no hardware flag) fall through to software signing.**

A watch-only account (`secretKey: ''`, created via add-watch-account) passes this `hardware === Ledger` check, reaches `handleSign`, and `PrivateKey.fromHex('')` throws deep in signing. Combined with the missing try/catch in `handleSign`, the user gets a confusing late `cancelled: true` instead of a clear error.

Suggestion: add an explicit `if (!signingAccount.secretKey)` guard that sends `signTypedDataError`, matching the `SIGNING_ACCOUNT_MISSING` pattern just above.

src/libs/ui/utils/formatters.ts
```diff
@@ -3,7 +3,8 @@ import Big from 'big.js';
 // eslint-disable-next-line import/no-duplicates
 import { formatDistanceToNowStrict } from 'date-fns';
 // eslint-disable-next-line import/no-duplicates
-import en from 'date-fns/locale/en-US';
+import { enUS as en } from 'date-fns/locale/en-US';
+import type { FormatDistanceToken } from 'date-fns/locale/types';
```
Comp0te: **`date-fns/locale/types` isn't part of date-fns v4's `exports` map — fragile import.**

In date-fns 4.x the `exports` field only declares `.`, `./locale`, and `./locale/<lang>` — there is no `./locale/types` entry and no wildcard. This line compiles today only because the repo uses `moduleResolution: "node"` (classic), which ignores the `exports` map and resolves the physical `node_modules/date-fns/locale/types.d.ts`. It will break with `TS2307` under a future move to `moduleResolution` `bundler`/`node16`/`nodenext`.

`FormatDistanceToken` is re-exported from the package root in v4 (`index.d.ts` → `export type * from "./types.ts"` → `export type * from "./locale/types.ts"`), so the supported import is:

```ts
import type { FormatDistanceToken } from 'date-fns';
```

or folded into the existing root import:

```ts
import { formatDistanceToNowStrict, type FormatDistanceToken } from 'date-fns';
```

src/utils.ts
```diff
@@ -211,6 +211,7 @@ export const getActiveAccountSupports = (activeAccount: Account) => {
     : [
         CasperWalletSupports.signDeploy,
         CasperWalletSupports.signMessage,
+        CasperWalletSupports.signTypedData,
```
Comp0te, line 214: Follow-up on the `signTypedData` capability (review feedback addressed here ✅).

There is a **second** supports list that wasn't updated: `changedActiveAccountSupportsEvent` in `src/apps/popup/hooks/use-account-actions-with-events.ts` (~L327-332) hardcodes its own array:

```ts
activeKeySupports: [
  CasperWalletSupports.signDeploy,
  CasperWalletSupports.signMessage,
  ...(supportsTransactionV1 ? [CasperWalletSupports.signTransactionV1] : [])
]
```

This omits `signTypedData` (and already omits `messageEncryption`/`messageDecryption`), so the two paths disagree:
- **Pull** — `getActivePublicKeySupports()` uses `getActiveAccountSupports()` → now correctly returns `signTypedData`. ✅
- **Push** — the `changedActiveAccountSupportsEvent` fired on active-account change reports the stale subset, so a dapp that caches supports from that event won't see EIP-712 until it re-pulls.

This inconsistency pre-dates this PR (hence low severity), but since we're advertising a new capability it's worth aligning. Suggestion: replace the hardcoded array with `getActiveAccountSupports(activeAccount)` so both paths share one source of truth.

Comp0te (approved):
👍

src/utils.ts
ost-ptk: Checked this before changing it — I think the current code is correct for EIP-712, and the literal swap would regress the tx-v1 path:

1. `changedActiveAccountSupportsEvent` only fires from the Ledger signing flow. `changeActiveAccountSupportsWithEvent` is wired as `supportsTransactionV1Cb`, which is invoked exclusively when `activeAccount.hardware === Ledger` (`signTx` in `deployer-service/index.ts:59`, and `ledger.ts:295/317`). For software accounts it never runs. So this event always reports a *Ledger* account, and `signTypedData` is intentionally unsupported on Ledger (EIP-712 rejects hardware accounts — the Ledger branch of `getActiveAccountSupports` omits it too). Omitting it here is therefore correct, not a gap.

2. Replacing the array with `getActiveAccountSupports(activeAccount)` would change the `signTransactionV1` source. The hardcoded array uses the freshly-detected `supportsTransactionV1` param (the Ledger app version capability detected at sign time). `getActiveAccountSupports` derives it from `activeAccount.supports` in the closure, which is still the pre-detection value at emit time — it only gets updated by the `activeAccountSupportsChanged` dispatch right below (L337). The whole point of this event is to push the just-detected tx-v1 capability, so reading the stale value would defeat it.

The `messageEncryption` omission you noted is a real (pre-existing, low-severity) inconsistency, but it is unrelated to EIP-712. Leaving this as-is for the PR; happy to align the Ledger supports subset separately if we want one source of truth.

PR #1366, opened 2026-06-24
fix(eip712): signature-request display fixes + core address Key decoding

ost-ptk:
## What

EIP-712 signature-request display fixes (WALLET-1251) from QA re-test, plus the `casper-wallet-core` bump that carries the address Key-decoding and date logic.

## Changes

- `formatPrimaryTypeLabel` splits the camelCase `primaryType` so the message header reads "TRANSFER WITH AUTHORIZATION" instead of an unreadable run-on word.
- Hash values (nonce, raw package hash) render blue with a copy icon.
- Bump `casper-wallet-core` to pick up: EIP-712 `address` values decoded as Casper Keys (a contract package address is no longer shown as a "Public key"), timestamp dates + sentinels ("Always" / "No expiry"), and CAIP-2 chain names.

## Notes

- Display correctness is driven by core; the wallet renders the prepared rows (package -> contract row, account -> account row, dates -> text). No additional wallet UI code was needed for the address fix.
- Based on `eip712-naming-alignment` (PR #1360); retarget to `develop` once that merges.

## Depends on

- casper-wallet-core PR make-software/casper-wallet-core#51 (the bumped commit lives on that branch; re-bump to the merge commit after it lands).

### DIFF
--- a/src/libs/ui/utils/formatters.ts
+++ b/src/libs/ui/utils/formatters.ts
@@ -179,6 +179,7 @@ export function formatPrimaryTypeLabel(value: string): string {
     .replace(/\s+/g, ' ')
     .trim()
     .split(' ')
+    .filter(Boolean)
     .map(word => capitalizeString(word))
     .join(' ');
 }
--- a/src/libs/ui/utils/formatters.ts
+++ b/src/libs/ui/utils/formatters.ts
@@ -169,6 +169,20 @@ export function capitalizeString(str: string): string {
   return `${str[0].toUpperCase()}${str.slice(1)}`;
 }
 
+export function formatPrimaryTypeLabel(value: string): string {
+  if (!value) {
+    return value;
+  }
+  return value
+    .replace(/[_-]+/g, ' ')
+    .replace(/([a-z0-9])([A-Z])/g, '$1 $2')
+    .replace(/\s+/g, ' ')
+    .trim()
+    .split(' ')
+    .map(word => capitalizeString(word))
+    .join(' ');
+}
+
 export const formatCurrency = (
   value: string,
   code: string,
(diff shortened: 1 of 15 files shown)


src/libs/ui/utils/formatters.ts
```diff
@@ -169,6 +169,20 @@ export function capitalizeString(str: string): string {
...
+  if (!value) {
+    return value;
+  }
+  return value
+    .replace(/[_-]+/g, ' ')
+    .replace(/([a-z0-9])([A-Z])/g, '$1 $2')
+    .replace(/\s+/g, ' ')
+    .trim()
+    .split(' ')
+    .map(word => capitalizeString(word))
```
Comp0te, line 183: `formatPrimaryTypeLabel` throws a `TypeError` on separator-only input.

After `.replace(/[_-]+/g, ' ')…trim()`, an input made up only of separators/whitespace reduces to `''`, and `''.split(' ')` yields `['']` — so `capitalizeString('')` runs `''[0].toUpperCase()` → `undefined.toUpperCase()` and throws. The early `if (!value)` guard only catches empty/undefined, not `'_'` / `'__'` / `'-'` / `'   '`.

```
"_"   => THROWS: Cannot read properties of undefined (reading 'toUpperCase')
"__"  => THROWS: ...
"   " => THROWS: ...
```

This is reachable from untrusted input: `primaryType` is a dapp-supplied string (`SignTypedDataParams.primaryType: string`) passed straight to `formatPrimaryTypeLabel(signatureRequest.primaryType)` in `sign-eip712-content.tsx` with no wallet-side validation. `_` is a valid EIP-712 / Solidity type identifier, so a dapp can send `primaryType: "_"` and crash the signature-request render.

Suggested fix — drop empty segments after the split:

```ts
    .split(' ')
    .filter(Boolean)
    .map(word => capitalizeString(word))
    .join(' ');
```

ost-ptk: Good catch — fixed in 8580c2d5. Added `.filter(Boolean)` after the split to drop empty segments, plus a regression test covering `"_"`, `"__"`, `"-"` and `"   "` (all now return `""` instead of throwing).

Comp0te (approved):

PR #1377, opened 2026-06-30
fix(deps): vulnerability remediation (WALLET-1328 / DEP-2)

ost-ptk:
## DEP-2 — Vulnerability remediation

**Jira:** https://make-software.atlassian.net/browse/WALLET-1328

Stacked on **DEP-1** (Node 18 → 22).

`npm audit` reports many findings, but **only one reaches the shipped bundle**:
`i18next-http-backend` (path traversal / URL injection via unsanitised `lng`/`ns`). It is
upgraded outright. The bundler's serializer can poison the bundle, so it is treated as
production-impacting and pinned via `overrides`. Everything else is dev/build-toolchain and
accepted dev-only (verified: `npm audit --omit=dev --audit-level=high` → **0**).

### Changes
- `i18next-http-backend` `2.5.0` → `^3.0.5` — v3 uses global `fetch`. **Zero source change**:
  `src/libs/i18n/i18n.ts` calls `.use(Backend)` with no `backend:{}` options block, so nothing
  from the removed v2 API is touched.
- `overrides` += `serialize-javascript@^7.0.5` (CVE-2026-34043), `tmp@^0.2.7` (CVE-2026-44705).
  The `serialize-javascript` override is **mandatory** — the webpack plugins resolve it to a
  vulnerable `6.x` with no patched 6.x release.
- New `audit:ci` script (`npm audit --omit=dev --audit-level=high`) wired into `ci-check.yml`:
  a **blocking** runtime gate plus a **non-blocking** full-tree audit for dev-tool visibility.
- `package-lock.json` regenerated from scratch under Node 22 (Dependabot/CI cannot regenerate it
  due to the private `git+ssh` dep).

### About the large `package-lock.json` diff
The diff is big but the dependency **versions barely moved** — most of it is npm 10 re-deduping
the tree (e.g. `@redux-devtools/cli` 4.0.3→4.0.4, `@apollo/server` 4.12.2→4.13.0;
`socketcluster-client` / `electron` unchanged). A clean regen was chosen deliberately because it
also floats real in-range security patches that a minimal in-place install misses — most notably
**`undici` 7.24.1 → 7.28.0** (7 advisories incl. one high). With the minimal lock, `audit:ci`
**fails** on `undici`; with the full regen it is **0**. `lockfileVersion` stays at 3 and CI
`npm ci` verifies integrity.

> Note: `undici` only counts as "runtime" because `i18next-parser` (a build-time CLI for
> `locale:extract_pot`, never imported in `src/`) sits in `dependencies` rather than
> `devDependencies`. It is not bundled. Reclassifying it is a minor follow-up, out of scope here.

### Verification
- `npm install` (Node 22.23.1) ✅
- `npm run audit:ci` → **0 vulnerabilities** (runtime) ✅
- `npm run ci-check` → 0 errors, 28 tests passed ✅
- `npm run build:chrome` + `grep -cE 'socketcluster|remotedev|composeWithDevTools' build/chrome/*.js` → **0** on every bundle ✅
- Firefox + Safari webpack builds succeed with the `serialize-javascript@^7` override active ✅
- i18n (structural): locales present under `build/chrome/locales/` (10 languages), no `backend:{}` block ✅

### Residual (full `npm audit`): 16 dev-only, **0 critical** (2 low / 7 moderate / 7 high)
All under tooling that is **never bundled**:
- `@redux-devtools/cli` chain (`devtools:redux`) → electron / tar / cacache / node-gyp / sqlite3 / make-fetch-happen / @apollo/server — accept dev-only (consider dropping for `@redux-devtools/extension` later).
- `@redux-devtools/remote` + `webpack-dev-server` → socketcluster-client / sockjs → uuid (no fix) — accept dev-only.
- The `shell-quote` (no-fix) and nested `ajv` highs anticipated earlier are **already gone** after Node 22 + `web-ext@10`.

### Out of scope (deferred)
- Dependabot → Renovate migration (needs the Renovate GitHub App; separate ticket).
- Optional `uuid` override and the `composeWithDevTools` dynamic-import hardening (already tree-shaken).
- Reclassifying `i18next-parser` as a devDependency.

### Manual smoke (recommended before merge)
- Switch UI language across the supported set; confirm locale JSON still loads via http-backend v3 `fetch` with no console errors.
- `npm run start:chrome` → Redux remote devtools still connect on `localhost:8000`.


Comp0te:
Suggestion: move `i18next-parser` from `dependencies` to `devDependencies` (`package.json`).

It is a build-time-only CLI — used solely by the `locale:extract_pot` script and never imported anywhere in `src/`, so it is never part of the shipped extension bundle.

Because it currently sits in `dependencies`, the new `audit:ci` gate (`npm audit --omit=dev`) treats its entire subtree (`undici`, `node-fetch`, `cross-fetch`, …) as runtime. That is the only reason `undici` had to be floated to 7.28.0 to keep the gate green — nothing about it actually ships. More importantly, any future advisory in that tree would fail the runtime security gate as a false positive.

Moving it to `devDependencies` is a one-line change that makes `audit:ci` audit genuinely-shipped code only. The PR body already flags this as a follow-up; folding it in here would make the very gate this PR introduces correct from the start.

ost-ptk:
Good catch — moved `i18next-parser` to `devDependencies` in d429e4e3 and regenerated the lockfile.

Confirmed:
- `npm run audit:ci` (`npm audit --omit=dev --audit-level=high`) now reports 0 vulnerabilities.
- `undici` is now correctly marked `dev: true` in the lockfile (was the sole reason it had to be floated to 7.28.0).
- `node-fetch`/`cross-fetch` stay in the runtime tree, but that's expected — `casper-js-sdk` also depends on `node-fetch` directly, so they're not false positives.
- `tsc` and the full test suite still pass; `locale:extract_pot` still resolves the `i18next` CLI fine from `devDependencies`.

Comp0te (approved):

PR #1387, opened 2026-07-02
build(deps): DEP-10 — styled-components 6 (WALLET-1337)

ost-ptk:
## Summary

Upgrades `styled-components` 5.3.11 → 6.4.3. v6 removes automatic DOM-prop filtering (v5 silently dropped non-HTML custom props like `color`, `active`, `loading` before they reached the DOM; v6 forwards everything, triggering React warnings and invalid DOM attributes).

- Restores v5 semantics globally via a new `CspStyleSheetManager` wrapper (`src/libs/ui/style-sheet-manager.tsx`, using `@emotion/is-prop-valid`), mounted in all 5 UI app entry points, instead of renaming ~313 custom-prop usages across ~140 files.
- Fixes v6 type/API breaks across the UI library: `typography.tsx`, `svg-icon.tsx` (both had local `withConfig({shouldForwardProp})` that needed the v6 signature update — see note below), `list.tsx` (removed `FlattenInterpolation`/`ThemeProps` types), `global-style.ts`, `containers.ts`, `avatar.tsx` (assorted v6 type-inference tightenings, discovered via a full `tsc --noEmit` pass, not caught by the original grep-based audit).
- **Real bug found via manual browser QA, not caught by tsc/lint/tests:** styled-components v6 resolves `shouldForwardProp` as `componentLevel || contextLevel` — a component's own `.withConfig({shouldForwardProp})` fully replaces (does not compose with) the global `StyleSheetManager` filter. The 3 `withConfig` sites in this codebase each only excluded their one local prop, so every other custom prop on those 3 components was still leaking to the DOM. Fixed by composing `isPropValid` into all 3 local callbacks.

## Scope note

This ticket originally also planned to replace `'unsafe-inline'` with a per-build CSP nonce in `style-src` (a separate security-hardening item). That work was implemented, then reverted after manual QA found `@lottiefiles/react-lottie-player` injects inline `<style>` tags with zero nonce support (no config hook, vendored `style-inject` utility) — under a nonce-only `style-src` this broke rendering. **CSP-nonce hardening is deferred to a follow-up ticket**, pending a decision on the lottie-player situation (CSP hash for its static injected CSS vs. swapping to `lottie-web` directly). `style-src` stays `'unsafe-inline'`, unchanged from before this PR. The `CspStyleSheetManager` wrapper this PR adds needs no rework when that follow-up lands — it already passes `process.env.CSP_NONCE` through as a harmless no-op.

Jira: https://make-software.atlassian.net/browse/WALLET-1337

## Test plan

- [x] `npm run ci-check` (format:check + lint + tsc + test) — all green
- [x] `npm run build:chrome` / `build:firefox` — both succeed; verified generated CSP unchanged (`'unsafe-inline'`, no nonce)
- [x] `npm run build:safari` — webpack/JS compile phase succeeds; Xcode PACK step fails on a pre-existing, unrelated issue (stale hardcoded chunk filenames in `project.pbxproj`, tracked separately, coordinated with DEP-11)
- [x] Manual QA in Chrome (popup, light/dark) — zero "React does not recognize the X prop" warnings, zero CSP violations, confirmed after fixing the `shouldForwardProp` composition bug
- [x] `extension-manifest-reviewer` subagent — 0 findings; confirmed the CSP-nonce introduction+revert is a byte-exact no-op on `webpack.config.js`/`src/utils.ts`, all 3 manifest JSON files untouched
- [x] Final whole-branch code review — 0 Critical/Important findings, 1 Minor (addressed)


Comp0te (approved):
👍

PR #1388, opened 2026-07-03
build(deps): DEP-11 — UI component libraries (react-player 3, react-virtualized removal, Safari chunk fix) (WALLET-1336)

ost-ptk:
## Summary

DEP-11 — UI component library modernization: 4 bump-only packages, react-player 2→3 (NFT media preview rewrite), react-virtualized removal, and a Safari `.pbxproj` build-robustness fix.

Jira: https://make-software.atlassian.net/browse/WALLET-1336

### Bump-only (zero source changes)
- `big.js` `^6.2.1` → `^7.0.1` (+`@types/big.js` → `^7.0.0`)
- `qrcode.react` `^3.1.0` → `^4.2.0`
- `react-inlinesvg` `3.0.3` → `^4.5.0`
- `react-infinite-scroll-hook` `^4.1.1` → `^5.0.2` (not v6 — needs React 19)

### react-player 2 → 3 (`nft-details/content.tsx`)
Full rewrite onto native `HTMLMediaElement`: `url`→`src`, removed `config.file.forceAudio`
(audio NFTs now render a native `<audio>` element directly), `styled(ReactPlayer)` pinned to
`wrapper: 'div'` so the CSS nested-selector still applies. Caught and fixed during review: v3's
`onReady` fires on native `loadstart` (near-instant) instead of v2's `canplay` — both the video
and audio branches now use `onCanPlay` to avoid a blank-frame flash before the preview is
actually ready.

### react-virtualized removal (`validator-list.tsx`)
**Decision: Option C (plain `.map()`) over Option B (`@tanstack/react-virtual`).** The list was
already capped at `min(3*80, n*80)` = max 3 visible rows before internal scrolling — virtualizing
a 3-row-visible list has no practical benefit. Removed `react-virtualized` +
`@types/react-virtualized` entirely (flagged React-19-incompatible, blocking for a future DEP-13).
Manual QA (delegate/redelegate validator list scroll) found no scroll/perf issue, so the
`@tanstack/react-virtual` fallback was not needed. The height-cap logic was extracted into a
small pure `get-validator-list-height.ts` helper with its own unit test (jest's
`moduleNameMapper` doesn't resolve `@libs/*`, so the helper needed to be dependency-free to be
testable at all — pre-existing repo constraint, not new).

### Safari `.pbxproj` chunk-file robustness fix
Separate from the package bumps, but done deliberately *before* the react-player bump: the Xcode
project previously hardcoded 28 individual `*.bundle.js` file references (entry bundles + numeric
split chunks + react-player's per-provider chunks), which broke every time webpack's chunk graph
shifted (already bit DEP-3, and react-player 2→3 changes the chunk set again). Replaced all 28
hardcoded references with one `PBXShellScriptBuildPhase` that clears and re-copies
`build/firefox/*.bundle.js` at build time — immune to future chunk-graph changes. Verified via
`plutil -lint` and a full `build:safari` + `xcodebuild`, confirming the built `.app`'s bundle set
exactly matches the fresh webpack output (0 diff).

## Test plan

- [x] `npm run ci-check` (format/lint/tsc/test) — green
- [x] `npm run build:chrome` / `build:firefox` / `build:safari` — all clean
- [x] Manual QA: Receive/Wallet-QR/account-settings QR codes
- [x] Manual QA: NFT video AND audio previews (react-player v3)
- [x] Manual QA: validator list scroll in delegate/undelegate flows
- [x] Manual QA: SVG icons app-wide incl. `currentColor` recolor
- [x] `npm run dependency:circular` — same warning count as before this branch (175), no new cycles


src/apps/popup/pages/stakes/components/validator-list.tsx
```diff
@@ -38,41 +46,36 @@ export const ValidatorList = ({
...
-              height={Math.min(3 * 80, filteredValidatorsList.length * 80)}
-              width={width}
-              rowCount={filteredValidatorsList.length}
-              rowRenderer={({ index, key, style }) => {
-                const validator = filteredValidatorsList[index];
-                const logo = validator?.svgLogo || validator?.imgLogo;
+        <ScrollContainer
+          maxHeight={getValidatorListHeight(filteredValidatorsList.length)}
+        >
+          {filteredValidatorsList.map(validator => {
```
Comp0te: This drops virtualization entirely (option C from the DEP-11 doc), but the rationale there — "the list is capped at 240px, so virtualization is nearly pointless" — conflates viewport height with item count. `useFetchValidators` requests up to 250 validators (`limit: 250`, with a pagination TODO, so it may grow), and this list renders inside a dropdown that filters live while the user types. With `react-virtualized` only ~8 rows were mounted (3 visible + 5 overscan); a plain `.map()` mounts all ~250 `ValidatorPlate`s in the popup:

- each plate without a logo draws a canvas identicon (md5 + canvas draw per row) on mount;
- plates with a logo render a plain `` without `loading="lazy"`, so all logo requests to the image proxy fire at once when the dropdown opens;
- every keystroke reconciles the full filtered list (~3–4k DOM nodes).

Please use **option B** instead: `@tanstack/react-virtual` — `useVirtualizer({ count, getScrollElement, estimateSize: () => 80 })`. It's a small hook-based lib, React-19-compatible, and we're already in the TanStack ecosystem (`@tanstack/react-query`). The new `ScrollContainer` (`max-height` + `overflow-y: auto`) is exactly the scroll element the virtualizer needs, so the migration from this version is small.

ost-ptk: Good catch, thanks for the detailed breakdown. Switched to Option B: replaced the plain `.map()` with `@tanstack/react-virtual`'s `useVirtualizer` (`estimateSize: () => ROW_HEIGHT_PX`, `overscan: 5` to match the previous `react-virtualized` config), rendering absolutely-positioned rows over the existing `ScrollContainer`, which is now the virtualizer's scroll element. Only mounted/near-viewport `ValidatorPlate`s render now, so the identicon-draw and eager logo-request cost no longer scales with the full ~250-validator list. Pushed in fc4a32a6.

Comp0te (approved):

PR #1389, opened 2026-07-03
build(deps): DEP-12 — ESLint 9 flat-config migration (WALLET-1340)

ost-ptk:
## DEP-12 — ESLint 9 flat-config migration

Part of the dependency-modernization epic. Jira: https://make-software.atlassian.net/browse/WALLET-1340

> Stacked on **DEP-11 (#1388)** — this PR targets `WALLET-1336-dep-11-ui-component-libraries`. Once #1388 merges into `release/2.6.0`, GitHub will retarget this PR automatically.

### What & why

ESLint 8 is EOL-track; ESLint 9 makes flat config the default. The active `.eslintrc` extended `eslint-config-react-app` (eslintrc-only, CRA-sunset, unmaintained), so dropping it and re-composing the rule base from the individual flat-config plugin presets is the bulk of the work. `eslint-config-airbnb`, `babel-eslint`, and `eslint-plugin-flowtype` were dead devDeps (extended nowhere) → clean removals. `src` is 100% `.ts/.tsx`, so the typescript-eslint parser covers everything.

### Package changes

- `eslint` 8.57 → **9.39.4** (not 10 — the react/import/jsx-a11y plugins don't yet declare ESLint 10 peers)
- replace `@typescript-eslint/parser` 5 with the **`typescript-eslint` 8** meta-package (parser + plugin)
- `eslint-plugin-react-hooks` 4 → **7.1.1**
- `eslint-config-prettier` → **10.1.8** (avoids the compromised 9.1.1 / 10.1.6 / 10.1.7 releases)
- `eslint-import-resolver-typescript` → 4.4.5, `eslint-plugin-import` → 2.32, `eslint-plugin-jsx-a11y` → 6.10.2, `prettier` → 3.9.4
- **add** `@eslint/js`, `globals`, `confusing-browser-globals` (MIT, zero-dep — powers `no-restricted-globals`, same package react-app used)
- **remove** dead devDeps: `eslint-config-react-app`, `eslint-config-airbnb`, `babel-eslint`, `eslint-plugin-flowtype`
- `@trivago/prettier-plugin-sort-imports` held at v4 (v6 needs Node 20 — separate stretch)

### Config changes

- `.eslintrc` → **`eslint.config.mjs`** (flat), rule base re-composed from `@eslint/js`, `typescript-eslint`, `eslint-plugin-react`, `eslint-plugin-jsx-a11y`, `eslint-plugin-import`, `eslint-plugin-react-hooks`, and `eslint-plugin-prettier` recommended presets
- custom rules ported 1:1; stylistic `semi` dropped (owned by prettier + eslint-config-prettier)
- `@typescript-eslint/no-unused-vars` kept at `error` via `caughtErrors: 'none'` to match the legacy `@typescript-eslint` v5 behaviour (parity, no source churn)
- lint kept **type-unaware** (no `parserOptions.project`) to stay fast across ~588 files
- lint script `eslint ./src --ext .js,.jsx,.ts,.tsx` → `eslint ./src`

### Restored react-app baseline rules

`eslint-config-react-app` enforced a set of error/warn rules **outside of any recommended preset**, so dropping the config silently un-enforced them — they moved from `on` to `absent` (not `off`), which is why they were missed by the net-new-disabled audit below. All six pass with **zero violations** against current `src`, so they are re-enabled at `error` (parity restored, no code churn) rather than deferred:

- `array-callback-return` — missing `return` in `map`/`reduce`/`filter` callbacks
- `no-restricted-globals` — accidental use of global `event`/`name`/`length`/`top` (full `confusing-browser-globals` list)
- `no-script-url` — `javascript:` URLs
- `no-eval` / `no-new-func` — dynamic code execution
- `no-throw-literal` — throwing non-`Error` values

Everything else react-app added on top of the recommended presets is stylistic and now owned by prettier. (Thanks @Comp0te for catching this in review.)

### Scope decision — pure config migration

The stricter ESLint 9 / typescript-eslint 8 / react-hooks 7 / eslint-plugin-react recommended presets surfaced ~198 violations from rules the old `react-app` baseline never enforced. To keep this a config-only migration (and avoid churning ~50 files, incl. `crypto/`, in a stacked deps PR), those net-new rules are **disabled with `TODO(WALLET-1340 follow-up)` comments and hit counts** in `eslint.config.mjs`. Adopting them (e.g. `react-hooks/set-state-in-effect`, `no-async-promise-executor`, `@typescript-eslint/no-explicit-any`) is a dedicated follow-up "eslint-strict" ticket.

The only source changes here are **mechanical autofix** (prettier 3.9 reformat, `prefer-const`, `no-extra-boolean-cast`) across 28 files.

### Verification

- `npm install` — clean (dead deps removed, no unmet peers)
- `npm run ci-check` — **green** (format:check + lint `0 errors` / 169 pre-existing benign warnings + tsc + 36 unit tests)
- `npx eslint --print-config` — resolves react / react-hooks / jsx-a11y / import / prettier rule sets
- e2e — onboarding **6/6**, popup **45 passed** (1 pre-existing service-worker flake, passed on retry)
- husky pre-commit (lint-staged) verified on commit
- CI runs `ci-check` on Node 22.x — ESLint 9 supported


eslint.config.mjs
```diff
@@ -0,0 +1,78 @@
...
+      '@typescript-eslint/no-unused-vars': [
+        'error',
+        { args: 'after-used', caughtErrors: 'none' }
+      ],
+      'jsx-a11y/no-autofocus': 'off',
+      'import/no-cycle': [1, { maxDepth: 10, ignoreExternal: true }],
+      'import/no-named-as-default': 0,
+      'import/export': 1,
+
+      // --- Deferred (DEP-12): rules newly introduced by the ESLint 9 /
```
Comp0te, line 65: **Non-blocking — for the eslint-strict follow-up (WALLET-1340):**

The migration re-composes the rule base from the individual recommended presets, which is the right call. One side effect worth tracking: `eslint-config-react-app` shipped a set of **error/warn rules that aren't part of any recommended preset**, so they're now silently unenforced — they don't show up in the "net-new rules disabled" list because they moved from _on_ to _absent_ rather than _off_.

The ones actually worth re-adding (real bug/security catchers, not style):

- `array-callback-return` — catches a missing `return` in `map`/`reduce`/`filter` callbacks
- `no-restricted-globals` — catches accidental use of the global `event`/`name`/`length`/`top` instead of a local
- `no-script-url` — blocks `javascript:` URLs
- `no-eval` / `no-new-func` — block dynamic code execution
- `no-throw-literal` — enforces throwing `Error` instances

Everything else react-app added is stylistic and owned by prettier now. Suggest folding these five into the eslint-strict ticket alongside the already-deferred rules, rather than losing them quietly.

ost-ptk: Good catch — thanks. Rather than deferring these to the eslint-strict ticket, I re-enabled them now: I probed all five (plus `no-new-func`) against current `src` and every one reports **zero violations**, so re-adding them restores react-app parity with no code churn.

Landed in bdf27290 as a dedicated `error`-severity block in `eslint.config.mjs` (kept at `error` so a future regression fails CI rather than only warns):

- `array-callback-return`
- `no-restricted-globals` — with the full `confusing-browser-globals` list (re-added the package, MIT / zero-dep, same source react-app used), so it actually catches `event`/`name`/`length`/`top` rather than being an empty rule
- `no-script-url`
- `no-eval` / `no-new-func`
- `no-throw-literal`

Everything else react-app added on top of the recommended presets is stylistic and now owned by prettier, so nothing else is lost quietly. The eslint-strict follow-up is now scoped to just the genuinely-violating deferred rules.

Comp0te (approved):

PR #1394, opened 2026-07-07
build(deps): DEP-16 — Redux Toolkit migration + P0.1/P1.1 (WALLET-1342)

ost-ptk:
Jira: https://make-software.atlassian.net/browse/WALLET-1342 (DEP-16); P0.1/P1.1 portion of https://make-software.atlassian.net/browse/WALLET-1343.

### What

Retires `typesafe-actions` (unmaintained since 2019, 75 files) in favor of **Redux Toolkit 2**, behavior-preserving, and folds in the two security/architecture items that rewrite the same files:

- **RTK migration**: all 18 slices → `createSlice` (each `actions.ts` kept as a re-export shim — zero churn for the 75 importers), store factory → `configureStore` (saga middleware kept; serializable/immutable checks off — required by the replica flow; remote devtools via `devToolsEnhancer`), 8 saga trigger actions → standalone `createAction` with legacy strings, message envelopes → `createAction` with prepare callbacks (**all 34 sdkMethod + 7 sdkEvent + 5 bringweb3 + popupStateUpdated wire strings byte-identical**; exactly the same 4 error envelopes carry `{error: true}`), `RootState` cutover to first-party `src/background/redux/store-types.ts`, `typesafe-actions` uninstalled.
- **P0.1 — sanitize popup replica**: the state broadcast to UI replicas no longer carries `vaultCipher` (removed from the `PopupState` type), password/salt hashes (nulled; public `keys.keysDoesExist` flag instead), or `session.encryptionKeyHash` (nulled; public `encryptionKeyDoesExist` flag). Unlock flows fetch at-rest material on demand via a new `privateStateRequest` message answered **only to extension-origin senders** (fail-closed: no response — not even an error shape — for untrusted senders). `storage.local` persistence unchanged.
- **P1.1 — dispatcher split**: `background/index.ts` 992 → ~260 lines; message handling moved to typed modules under `src/background/handlers/` (sdk-methods, redux-actions, bringweb3, legacy-import, private-state) behind a thin parse→route→delegate router with verified 73/73 forwarding-case parity.
- **Tests**: 18 characterization test suites written against the OLD reducers and kept passing unchanged after conversion (0 → ~150 redux tests); jest alias mapper completed; coverage-gate decision (shared with P1.6/DEP-5): scoped `collectCoverageFrom` = redux reducers @ 100% (green); new `test:coverage` script.

### Scope addition (security)

The legacy import-account handlers (`check-secret-key-exist` / `check-account-name-is-taken` / `get-window-id`) are now gated by the same trusted-sender check — previously `check-secret-key-exist` answered a **secret-key membership oracle to any sender** (pre-existing). The only legitimate caller (import-account-with-file window) passes the gate.

### Behavior notes (intentional)

1. Unknown/unroutable messages now actually reject with the same error text — previously the `throw` inside an async Promise executor was swallowed and senders saw a forever-pending promise. Improvement, not a regression.
2. Store-internal action type strings are namespaced (`vault/accountAdded` vs `ACCOUNT_ADDED`) — verified nothing matches type strings; wire/envelope strings unchanged.
3. Immer now deep-freezes store state (incl. vault). Zero in-place mutations exist today (grep-verified by crypto review); any FUTURE in-place secret-zeroing utility must operate on copies or via a reducer, or it will throw.
4. `keysDoesExist` is persisted inside the keys blob as a harmless derived byte — always recomputed at preload; removing it would need a storage migration.
5. `fetchPrivateState()` rejection (e.g. SW restart mid-open) renders the unlock page blank without retry — fail-closed, ms-window; retry UX is a follow-up.

### Accepted residual (documented in the design spec)

The **decrypted `vault` slice** (secret phrase + account keys while unlocked) remains in the replica broadcast — the UI signing architecture depends on it; moving signing to the background is a separate epic item, out of DEP-16 scope.

### Verification

`ci-check` (62 suites / 306 tests, lint 0 errors), `test:coverage` 100% (18/18 reducers), chrome + firefox + safari builds, Playwright e2e onboarding 6/6 + popup 46/47 (+1 skipped), wire-shape spot-check (JSON key sets identical), manual QA (unlock flows, playground dapp round-trip incl. error envelopes, `popupStateUpdated` payload inspection), crypto-security review (vault slice — clean), extension-manifest review (dispatcher split — clean; no manifest/CSP changes, all MV3 top-level listeners preserved), per-task reviews + final whole-branch review.

### Follow-ups (for the WALLET-1343 follow-up PR)

- Dead private-material selectors cleanup (`selectPasswordHash` etc. — now unused, fail-safe).
- `fetchPrivateState` retry/error UX on unlock pages.
- Optional unit test freezing the 34 sdkMethod wire strings.
- P0.4 (Firefox/Safari DNR) + P1.6 (background test foundation) + P1.5 b/c/d (manifest hardening) per the agreed split.

### DIFF
--- a/src/apps/popup/pages/password-protection-page/index.tsx
+++ b/src/apps/popup/pages/password-protection-page/index.tsx
@@ -10,16 +10,15 @@ import {
 import { PasswordDoesNotExistError } from '@src/errors';
 import { getErrorMessageForIncorrectPassword } from '@src/utils';
 
-import {
-  selectPasswordHash,
-  selectPasswordSaltHash
-} from '@background/redux/keys/selectors';
+import { selectKeysDoesExist } from '@background/redux/keys/selectors';
 import {
   loginRetryCountIncremented,
   loginRetryCountReseted
 } from '@background/redux/login-retry-count/actions';
 import { selectLoginRetryCount } from '@background/redux/login-retry-count/selectors';
 import { dispatchToMainStore } from '@background/redux/utils';
+
+import { usePrivateState } from '@hooks/use-private-state';
 
 import {
   FooterButtonsContainer,
@@ -51,8 +50,8 @@ export const PasswordProtectionPage = ({
 
   const { t } = useTranslation();
 
-  const passwordHash = useSelector(selectPasswordHash);
-  const passwordSaltHash = useSelector(selectPasswordSaltHash);
+  const privateState = usePrivateState();
+  const keysDoesExist = useSelector(selectKeysDoesExist);
   const loginRetryCount = useSelector(selectLoginRetryCount);
 
   const attemptsLeft =
@@ -60,7 +59,7 @@ export const PasswordProtectionPage = ({
     loginRetryCount -
     ERROR_DISPLAYED_BEFORE_ATTEMPT_IS_DECREMENTED;
 
-  if (passwordHash == null || passwordSaltHash == null) {
+  if (!keysDoesExist) {
     throw new PasswordDoesNotExistError();
   }
 
@@ -77,6 +76,10 @@ export const PasswordProtectionPage = ({
   });
 
   const onSubmit = () => {
+    if (privateState == null) return;
+
+    const { passwordHash, passwordSaltHash } = privateState;
+
     setIsSubmitting(true);
 
     const { password } = getValues();
@@ -124,6 +127,11 @@ export const PasswordProtectionPage = ({
       setIsSubmitting(false);
     };
   };
+
+  // private state (hashes) arrives in ms; matches existing async-boot behavior
+  if (privateState == null) {
+    return null;
+  }
 
   return (
     <PopupLayout
--- a/src/apps/popup/pages/password-protection-page/index.tsx
+++ b/src/apps/popup/pages/password-protection-page/index.tsx
@@ -25,6 +25,7 @@ import {
   HeaderPopup,
   HeaderSubmenuBarNavLink,
   PopupLayout,
+  PrivateStateErrorPage,
   UnlockProtectedPageContent
 } from '@libs/layout';
 import { Button } from '@libs/ui/components';
@@ -50,7 +51,11 @@ export const PasswordProtectionPage = ({
 
   const { t } = useTranslation();
 
-  const privateState = usePrivateState();
+  const {
+    privateState,
+    error: privateStateError,
+    retry: retryPrivateState
+  } = usePrivateState();
   const keysDoesExist = useSelector(selectKeysDoesExist);
   const loginRetryCount = useSelector(selectLoginRetryCount);
 
@@ -127,6 +132,10 @@ export const PasswordProtectionPage = ({
       setIsSubmitting(false);
     };
   };
+
+  if (privateStateError) {
+    return <PrivateStateErrorPage layout="popup" onRetry={retryPrivateState} />;
+  }
 
   // private state (hashes) arrives in ms; matches existing async-boot behavior
   if (privateState == null) {
--- a/src/background/handlers/redux-actions.ts
+++ b/src/background/handlers/redux-actions.ts
@@ -1,0 +1,208 @@
+import { backgroundEvent } from '@background/background-events';
+import {
+  accountInfoReset,
+  accountPendingDeployHashesChanged,
+  accountPendingDeployHashesRemove,
+  accountTrackingIdOfSentNftTokensChanged,
+  accountTrackingIdOfSentNftTokensRemoved
+} from '@background/redux/account-info/actions';
+import {
+  dismissAppEvent,
+  resetAppEventsDismission
+} from '@background/redux/app-events/actions';
+import {
+  contactRemoved,
+  contactUpdated,
+  contactsReseted,
+  newContactAdded
+} from '@background/redux/contacts/actions';
+import { MainStore } from '@background/redux/get-main-store';
+import {
+  ledgerDeployChanged,
+  ledgerNewWindowIdChanged,
+  ledgerRecipientToSaveOnSuccessChanged,
+  ledgerStateCleared,
+  ledgerTransactionChanged
+} from '@background/redux/ledger/actions';
+import {
+  askForReviewAfterChanged,
+  ratedInStoreChanged,
+  resetRateApp
+} from '@background/redux/rate-app/actions';
+import { RootAction } from '@background/redux/store-types';
+import {
+  addWasmToTrusted,
+  removeAllWasmFromTrustedOrigin,
+  removeWasmFromTrusted,
+  resetTrustedWasmState
+} from '@background/redux/trusted-wasm/actions';
+import {
+  accountAdded,
+  accountDisconnected,
+  accountImported,
+  accountRemoved,
+  accountRenamed,
+  accountsAdded,
+  accountsImported,
+  activeAccountChanged,
+  activeAccountSupportsChanged,
+  addWatchingAccount,
+  anotherAccountConnected,
+  deploysReseted,
+  hideAccountFromListChanged,
+  secretPhraseCreated,
+  siteConnected,
+  siteDisconnected,
+  vaultLoaded,
+  vaultReseted
+} from '@background/redux/vault/actions';
+import {
+  connectWindowInit,
+  importWindowInit,
+  onboardingAppInit,
+  popupWindowInit,
+  signWindowInit,
+  windowIdChanged,
+  windowIdCleared
+} from '@background/redux/windowManagement/actions';
+
+import { enableOnboardingFlow } from '../open-onboarding-flow';
+import { keysReseted, keysUpdated } from '../redux/keys/actions';
+import { lastActivityTimeRefreshed } from '../redux/last-activity-time/actions';
+import {
+  loginRetryCountIncremented,
+  loginRetryCountReseted
+} from '../redux/login-retry-count/actions';
+import { loginRetryLockoutTimeSet } from '../redux/login-retry-lockout-time/actions';
+import {
+  recipientPublicKeyAdded,
+  recipientPublicKeyReseted
+} from '../redux/recent-recipient-public-keys/actions';
+import {
+  createAccount,
+  initKeys,
+  initVault,
+  lockVault,
+  recoverVault,
+  resetVault,
+  unlockVault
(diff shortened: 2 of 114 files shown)


src/hooks/use-private-state.ts
```diff
@@ -0,0 +1,26 @@
...
+  PrivateState,
+  fetchPrivateState
+} from '@background/handlers/private-state';
+
+export function usePrivateState(): PrivateState | null {
+  const [privateState, setPrivateState] = useState<PrivateState | null>(null);
+
+  useEffect(() => {
+    let mounted = true;
+    fetchPrivateState()
```
Comp0te: **[High] One-shot `fetchPrivateState()` with no retry/timeout — the unlock page can wedge blank.**

This fetch runs once (`[]` deps) and only `console.error`s on rejection. Every unlock/re-auth screen renders `return null` (blank) while `privateState == null` (`libs/layout/unlock-vault`, `password-protection-page`, onboarding `unlock-wallet`).

On the known MV3 service-worker-restart race, `runtime.sendMessage({ type: 'PRIVATE_STATE_REQUEST' })` can reject ("Could not establish connection"). Then `privateState` stays `null` forever and the user sees a blank screen with no spinner/error/retry until they close and reopen the popup. Previously the hashes/cipher rode the always-populated state replica, so the unlock form always rendered.

**Suggested fix:** add a timeout + bounded retry and an explicit error/blocked UI state instead of returning `null`. (Also bounds the silent-hang case in #3.)

src/apps/popup/pages/password-protection-page/index.tsx
```diff
@@ -51,16 +50,16 @@ export const PasswordProtectionPage = ({
 
   const { t } = useTranslation();
 
-  const passwordHash = useSelector(selectPasswordHash);
-  const passwordSaltHash = useSelector(selectPasswordSaltHash);
+  const privateState = usePrivateState();
```
Comp0te: **[Medium] Private state is snapshotted once and never resubscribes (stale-hash regression).**

`usePrivateState` captures the material on mount and — because it is no longer broadcast — never updates. This re-auth gate verifies the entered password against the **snapshotted** `passwordHash`/`passwordSaltHash`.

If the password is changed in another window while this page is mounted, the **new** password is rejected and the **old** one still validates. The same staleness affects unlock-vault's snapshotted `vaultCipher`/`keyDerivationSaltHash`. The old `useSelector` read stayed live via the replica.

**Suggested fix:** re-fetch when the relevant public flags change, or expose a scoped push channel for these updates.

src/background/index.ts
```diff
@@ -300,666 +168,79 @@ tabs.onUpdated.addListener(async (tabId, changeInfo, tab) => {
...
-          case getType(removeAllWasmFromTrustedOrigin):
-          case getType(resetTrustedWasmState):
-          case getType(systemColorSchemeChanged):
-            store.dispatch(action);
-            return sendResponse(undefined);
-
-          case getType(backgroundEvent.popupStateUpdated):
-            // do nothing
+        if (isPrivateStateRequest(action)) {
+          if (!isTrustedUiSender(sender)) {
```
Comp0te, line 210: **[Low-Med] Untrusted/unmatched sender path resolves neither side of the promise → silent hang.**

For a sender failing `isTrustedUiSender`, this returns without `sendResponse`/`sendError`, leaving the caller's `sendMessage` promise pending. That is intended for genuinely untrusted senders — but combined with `usePrivateState`'s swallowed `.catch`, any *legitimate-but-non-standard* UI origin whose `sender.url` doesn't string-prefix-match `runtime.getURL('')` (e.g. a Safari web-extension packaging variant, or a blob/sandboxed iframe host) hangs forever on a blank page with not even a console error.

Fixing the `usePrivateState` timeout (#1) bounds this. Additionally consider logging a diagnostic when a same-`runtime.id` sender is rejected by the URL-prefix check, so a rejected trusted page is distinguishable from a bug.

src/background/redux/keys/selectors.ts
```diff
@@ -1,9 +1,7 @@
-import { RootState } from 'typesafe-actions';
+import { RootState } from '@background/redux/store-types';
 
 export const selectKeysDoesExist = (state: RootState): boolean =>
-  state.keys.passwordHash != null &&
-  state.keys.passwordSaltHash != null &&
-  state.keys.keyDerivationSaltHash != null;
+  state.keys.keysDoesExist;
 
 export const selectPasswordHash = (state: RootState): string | null =>
```
Comp0te, line 6: **[Minor / cleanup] Dead private-material selectors.**

`selectPasswordHash`, `selectPasswordSaltHash`, `selectKeyDerivationSaltHash` (here) and `selectVaultCipher` (`vault-cipher/selectors.ts`) now have zero UI callers — the only consumer was migrated to `usePrivateState`. Left in place, they read fields that are `null` in the sanitized replica, so a future `useSelector(selectPasswordHash)` in a popup would silently get `null` at runtime instead of a compile error.

Consider removing them (or scoping to background-only). Already listed as a follow-up in the PR description.

src/background/handlers/redux-actions.ts
```diff
@@ -0,0 +1,208 @@
...
+  themeModeSettingChanged,
+  vaultSettingsReseted
+} from '../redux/settings/actions';
+import {
+  vaultCipherCreated,
+  vaultCipherReseted
+} from '../redux/vault-cipher/actions';
+import { HandlerResult } from './types';
+
+const FORWARDED_ACTION_TYPES: ReadonlySet<string> = new Set(
```
Comp0te, line 109: **[Minor / cleanup] `FORWARDED_ACTION_TYPES` is a hand-maintained ~74-entry list with no parity check.**

It is correctly fail-closed (an unknown type throws `Unknown redux action`), and I verified it currently matches the old dispatcher's cases 1:1. But it duplicates information already in the slice/action registry with nothing tying them together — add a new UI-dispatched slice action and forget to append it here, and the dispatch throws at runtime with no compile-time signal.

Consider deriving the set from the action registry, or adding a parity test that fails when a forwardable action is missing.

src/background/redux/get-main-store.ts
```diff
@@ -105,11 +109,19 @@ export async function getExistingMainStoreSingletonOrInit() {
...
+        // The MOCK store IS the background store: it keeps the full RootState
+        // (incl. vaultCipher + real hashes); only the broadcast is sanitized.
+        storeSingleton = createStore(initialStateForPopupTests);
       } else {
         storeSingleton = createStore({
           vaultCipher,
-          keys,
+          keys: keys && {
+            ...keys,
+            keysDoesExist:
```
Comp0te, line 120: **[Minor / cleanup] `keysDoesExist` derivation duplicated and persisted.**

The three-way non-null check lives here (preload) and again in `keys/reducer.ts` `withDerivedFlag`; the two can drift. It is also materialized into state and persisted under `KEYS_KEY` despite being fully derivable from the three hashes it sits beside.

Consider reusing `withDerivedFlag` here, or keeping the flag a pure selector and not persisting it.

Comp0te (commented):
Focused follow-up review of the P0.1 private-state fetch-on-demand flow, plus a few optional cleanups. The RTK migration and dispatcher split themselves look clean and behavior-preserving — I verified the dispatcher forwarding parity is exact (every UI-dispatched creator is in `FORWARDED_ACTION_TYPES`) and the SDK/event/bringweb3 wire strings are byte-identical. The inline comments below are: one higher-priority availability item on the new private-state fetch (#1) and two related to it (#2, #3), then three minor cleanups.

src/hooks/use-private-state.ts
ost-ptk: Fixed in 2c0eaf9d: `usePrivateState` now wraps the fetch in a 5s per-attempt timeout with 2 automatic retries (250/500ms backoff, targeting the SW-restart race) and returns `{ privateState, error, retry }`. All three unlock/re-auth pages render an error page with a "Try again" button instead of the blank `null` branch when the fetch ultimately fails.

src/apps/popup/pages/password-protection-page/index.tsx
ost-ptk: Confirmed the staleness — with one note: the "re-fetch when the relevant public flags change" variant would not cover this case, since a password change does not flip `keysDoesExist`/`encryptionKeyDoesExist`. Landing the scoped push channel as a follow-up in WALLET-1343: the store subscriber broadcasts a payload-free `privateStateUpdated` signal when the private-state selection changes, and `usePrivateState` re-fetches through the existing gated request. Scheduled there rather than here to keep this QA-passed branch minimal.

src/background/index.ts
ost-ptk: Fixed in 2c0eaf9d: same-`runtime.id` senders that fail the URL-prefix check now log a diagnostic `console.warn` (origin only — a content-script sender's full page URL could carry tokens in query strings), and the no-response contract for untrusted senders is unchanged. The client-side hang is bounded by the 5s timeout added for #1.

src/background/redux/keys/selectors.ts
ost-ptk: Scheduled in WALLET-1343 (dead-code sweep task in the tests/DX PR) — it also drops the unused `popupStateUpdated` type export, `RootStateKey`, and the `RootAction` dual alias. Kept out of this branch since PRs stacked on it may still grow consumers; importers get re-verified at execution time.

src/background/handlers/redux-actions.ts
ost-ptk: Scheduled in WALLET-1343: a parity test that collects `.type` from every UI-dispatched action-creator module and asserts set equality with `FORWARDED_ACTION_TYPES` (plus an explicit exclusion list for background-only actions), so drift in either direction fails CI.

src/background/redux/get-main-store.ts
ost-ptk: Scheduled in WALLET-1343: `withDerivedFlag` gets exported from `keys/reducer.ts` and reused in the `get-main-store` preload, removing the duplicated derivation. The flag stays materialized + persisted, though — dropping it from `KEYS_KEY` changes the stored shape and would need a strip/migration for existing users, which is not worth it for a derivable boolean.

Comp0te (approved):
🥇

PR #1395, opened 2026-07-07
fix: DEP-99 — private SDK ↔ content-script channel (WALLET-1343)

ost-ptk:
## DEP-99 · PR 1 of 8 — P0.2: private SDK ↔ content-script channel

Hardens the dapp SDK ↔ background request path. Previously any script on the page could forge the
`window` CustomEvent that the content script forwarded to the background, and request ids were a
guessable per-instance counter. This PR moves requests onto a private `MessageChannel` and makes ids
unpredictable.

### Commits

| Commit     | Change                                                                                                                                                                                                                                                                                                                                                    |
| ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `2cf450bc` | SDK `requestId` uses `crypto.randomUUID()` instead of a counter from 0 (guessable + collides across two provider instances).                                                                                                                                                                                                                              |
| `464bc6d0` | `requestId` degrades to a CSPRNG (`crypto.getRandomValues`) RFC4122-v4 UUID on non-secure (plain-http) origins, where `crypto.randomUUID` is unavailable — prevents every SDK request throwing on http dapps.                                                                                                                                             |
| `43340102` | Private `MessageChannel` transport: the content script hands a `MessagePort` to the page-world SDK during a `scriptTag.onload` handshake (validated by `isTrustedWindowMessage` = same-window + same-origin); the forgeable `CasperWalletMethod:Request` window-listener path is removed. Wallet-event broadcast (`emitSdkEvent`) stays on window events. |
| `8d909195` | Freeze test pinning all 34 `sdkMethod` + 7 `sdkEvent` + 5 `bringWeb3Events` wire strings, group counts, and the 4 error-envelope creators — protects this PR's plumbing and every future refactor.                                                                                                                                                        |

### Threat-model note

The page main world is shared with any same-origin script, so a `MessagePort` transferred via
`window.postMessage` cannot be cryptographically isolated from a co-resident same-origin script that
races the handshake — an unavoidable MV3 page-world property. This PR does NOT attempt that. What it
buys: rejects cross-origin/cross-window injected messages, removes the trivially-forgeable DOM-event
request surface, and keeps request/response payloads off the public window-event bus.

### Manifest / CSP / permissions

Reviewed by the extension-manifest-reviewer: **no delta** across Chrome MV3 / Firefox MV2 / Safari
MV2 — no `permissions`, `host_permissions`, `web_accessible_resources`, CSP, or DNR change. The new
transport uses page-world DOM APIs (`MessageChannel`, `postMessage` port transfer) that need no
capability grant; `sdk.bundle.js` remains a declared web-accessible resource in all three manifests.

### Verification

- ✅ `npm run ci-check` green (tsc + lint clean); `jest src/content` 15/15 against `release/2.6.0`.
- ✅ `npm run build:chrome` (webpack prod) exit 0.
- ✅ `npm run build:firefox` (webpack prod) exit 0.
- ✅ extension-manifest-reviewer: clean (no manifest/CSP/DNR delta; MV2/MV3/Safari parity intact).
- ✅ Manual playground QA: **connect** verified on the prod chrome build, console clean. Remaining SDK flows (sign / signMessage / signTypedData / disconnect / reload-tab re-handshake) ride the same port transport as connect.
- ⬜ **Safari build** (`build:safari`) — needs Xcode, not run in this environment.

### Base branch

DEP-16 (#1394) has merged into `release/2.6.0` as a squash. This branch was replanted (`rebase --onto`)
onto `release/2.6.0` so it carries **only the 4 P0.2 commits** — the un-squashed DEP-16 history and the
merge-forward were dropped (their content is already in `release/2.6.0`). Diff is now exactly the
`src/content/` P0.2 change, no DEP-16 leakage.


### DIFF
--- a/src/content/index.ts
+++ b/src/content/index.ts
@@ -2,9 +2,15 @@ import { runtime } from 'webextension-polyfill';
 
 import { initBringScript } from '@content/bring';
 
+import { SDK_HANDSHAKE_TYPE } from './sdk-channel';
 import { SdkEvent, sdkEvent } from './sdk-event';
 import { CasperWalletEventType } from './sdk-event-type';
-import { SdkMethodEventType, isSDKMethod, sdkMethod } from './sdk-method';
+import { isSDKMethod, sdkMethod } from './sdk-method';
+
+// The private port handed to the page-world SDK during the handshake. Both the
+// direct `runtime.sendMessage` response and the delayed `runtime.onMessage`
+// response ride this port back to the SDK — never the public window bus.
+let activePort: MessagePort | null = null;
 
 async function handleSdkMessage(message: unknown) {
   // delayed sdk request response
@@ -25,11 +31,8 @@ async function handleSdkMessage(message: unknown) {
       case sdkMethod.encryptMessageResponse.type:
       case sdkMethod.encryptMessageError.type:
       case sdkMethod.getActivePublicKeySupportsResponse.type:
-        window.dispatchEvent(
-          new CustomEvent(SdkMethodEventType.Response, {
-            detail: JSON.stringify(message)
-          })
-        );
+        // route the delayed response over the private port (not a window event)
+        activePort?.postMessage(message);
         return;
 
       default:
@@ -87,31 +90,44 @@ function emitSdkEvent(message: SdkEvent) {
   window.dispatchEvent(event);
 }
 
-// SDK Message proxy to the backend
-function handleSdkRequestEvent(e: Event) {
-  const requestAction = (e as CustomEvent).detail;
-  // validation
-  if (!isSDKMethod(requestAction)) {
-    throw Error(
-      'Content: invalid sdk requestAction: ' + JSON.stringify(requestAction)
-    );
-  }
+// SDK Message proxy to the backend, over the private MessageChannel.
+//
+// The content script keeps `port1`; requests that arrive on it are shaped-checked
+// with `isSDKMethod` and forwarded to the background. Direct responses ride back
+// on the same port. `port2` is transferred to the page-world SDK during the
+// handshake, so only the holder of that port (the injected SDK) can drive this
+// path — a script that forges the old `Request` window CustomEvent gets nothing,
+// because that window listener no longer exists.
+function establishSdkPort() {
+  const channel = new MessageChannel();
 
-  runtime
-    .sendMessage(requestAction)
-    .then(message => {
-      // if valid message send back response
-      if (isSDKMethod(message)) {
-        window.dispatchEvent(
-          new CustomEvent(SdkMethodEventType.Response, {
-            detail: JSON.stringify(message)
-          })
-        );
-      }
-    })
-    .catch(err => {
-      throw Error('Content: sdk request received error: ' + err);
-    });
+  channel.port1.onmessage = event => {
+    const requestAction = event.data;
+    // reject anything not shaped like an SDK method
+    if (!isSDKMethod(requestAction)) {
+      return;
+    }
+
+    runtime
+      .sendMessage(requestAction)
+      .then(message => {
+        // if valid message send back response over the private port
+        if (isSDKMethod(message)) {
+          channel.port1.postMessage(message);
+        }
+      })
+      .catch(err => {
+        console.error('Content: sdk request received error: ', err);
+      });
+  };
+
+  activePort = channel.port1;
+
+  // hand port2 to the page-world SDK; origin-scoped so it is never delivered
+  // cross-origin.
+  window.postMessage({ type: SDK_HANDSHAKE_TYPE }, window.location.origin, [
+    channel.port2
+  ]);
 }
 
 // inject sdk script - idempotent, doesn't need cleanup
@@ -125,6 +141,10 @@ function injectSdkScript() {
     scriptTag.src = runtime.getURL(inpageScriptPath);
     scriptTag.onload = function () {
       documentHeadOrRoot.removeChild(scriptTag);
+      // The SDK bundle has now evaluated and registered its handshake `message`
+      // listener, so handing over the port here closes the race that posting
+      // synchronously from `init()` (before the SDK loads) would open.
+      establishSdkPort();
     };
     documentHeadOrRoot.insertBefore(scriptTag, documentHeadOrRoot.children[0]);
   } catch (e) {
@@ -137,7 +157,6 @@ function init() {
   injectSdkScript();
 
   runtime.onMessage.addListener(handleSdkMessage);
-  window.addEventListener(SdkMethodEventType.Request, handleSdkRequestEvent);
 }
 
 // cleanup logic
@@ -147,7 +166,9 @@ function cleanup() {
   document.removeEventListener(cleanupEventType, cleanup);
 
   runtime.onMessage.removeListener(handleSdkMessage);
-  window.removeEventListener(SdkMethodEventType.Request, handleSdkRequestEvent);
+  // tear down the private port so a superseded instance can't keep proxying
+  activePort?.close();
+  activePort = null;
 }
 window.addEventListener(cleanupEventType, cleanup);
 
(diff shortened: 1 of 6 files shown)


src/content/sdk.ts
```diff
@@ -38,61 +50,84 @@ function fetchFromBackground<T extends SdkMethod['payload']>(
   options?: CasperWalletProviderOptions
 ): Promise<T> {
   return new Promise((resolve, reject) => {
+    const port = sdkPort;
+    if (!port) {
+      // handshake hasn't completed yet — the content script hands the port over
+      // on SDK load, so this should only happen if a request is fired before the
+      // page fully initialised.
+      reject(Error(`SDK PORT NOT ESTABLISHED: ${requestAction.type}`));
```
Comp0te: **Pre-handshake requests hard-fail instead of waiting (behavioral regression).**

There is a real window where `window.CasperWalletProvider` already exists (assigned at bundle evaluation) but `sdkPort` is still `null`: `establishSdkPort()` runs in the `scriptTag.onload` task, plus one more task for the `postMessage` delivery. A dapp using the common "poll for `window.CasperWalletProvider`, call immediately" pattern can land in that gap and get a hard reject here.

The old transport had no such gap — the content-script `Request` listener existed before the SDK was even injected, so a request fired at any time was picked up.

Suggestion: instead of rejecting, queue the request while `sdkPort` is `null` and flush the queue when the handshake completes, keeping the existing per-request timeout as the bound. That turns the failure mode from "flaky at page load" into "waits a few ms".

src/content/index.ts
```diff
@@ -87,31 +90,44 @@ function emitSdkEvent(message: SdkEvent) {
...
+// handshake, so only the holder of that port (the injected SDK) can drive this
+// path — a script that forges the old `Request` window CustomEvent gets nothing,
+// because that window listener no longer exists.
+function establishSdkPort() {
+  const channel = new MessageChannel();
+
+  channel.port1.onmessage = event => {
+    const requestAction = event.data;
+    // reject anything not shaped like an SDK method
+    if (!isSDKMethod(requestAction)) {
```
Comp0te: **Relay filter could be an allowlist of known request types (defense-in-depth).**

`isSDKMethod` only checks the shape — any `{type: string, meta: {requestId: string}}` passes, so this relay forwards arbitrary-typed messages to the background, including `*:Response`-shaped ones and redux-action types like `resetVault`.

Today this is safe only because of ordering in `src/background/index.ts`: the `isSDKMethod` branch runs before `handleReduxAction`, and `handleSdkMethod` rejects unknown types — the comment there pins exactly this. But the appendix doc (P0.2 note) calls this filter out as a single point of failure: if that router ordering ever changes, the relay becomes a store-write primitive.

Since this PR already adds a freeze test pinning all 34 `sdkMethod` wire strings, a `Set` of the request-direction type strings here is a few lines and removes the dependency on router ordering entirely — only genuine SDK requests should ever ride page → background.

src/content/index.ts
```diff
@@ -87,31 +90,44 @@ function emitSdkEvent(message: SdkEvent) {
...
+    runtime
+      .sendMessage(requestAction)
+      .then(message => {
+        // if valid message send back response over the private port
+        if (isSDKMethod(message)) {
+          channel.port1.postMessage(message);
+        }
+      })
+      .catch(err => {
+        console.error('Content: sdk request received error: ', err);
```
Comp0te, line 143: **`runtime.sendMessage` failure leaves the dapp hanging up to 30 minutes.**

When the send fails (e.g. the MV3 service-worker restart race), this only logs to the console — the SDK never learns the request died and waits out the per-request timeout, which defaults to 30 minutes.

`requestAction.meta.requestId` is in scope here, so the port can carry a synthetic error envelope back and the dapp's promise rejects immediately instead. (This matches the old behavior, so not a regression — but it's a cheap improvement while this code is being touched.)

src/content/index.ts
```diff
@@ -87,31 +90,44 @@ function emitSdkEvent(message: SdkEvent) {
...
+      });
+  };
+
+  activePort = channel.port1;
+
+  // hand port2 to the page-world SDK; origin-scoped so it is never delivered
+  // cross-origin.
+  window.postMessage({ type: SDK_HANDSHAKE_TYPE }, window.location.origin, [
```
Comp0te: **Nit: throws on opaque-origin documents.**

For a document with an opaque origin (e.g. a page served with a CSP `sandbox` header), `window.location.origin` is the string `"null"`, which is not a valid `targetOrigin` — `postMessage` throws a `SyntaxError`. Since this runs inside the `onload` handler, the exception is uncaught and the SDK silently never receives a port; every subsequent request rejects.

Edge case (matches are http/https, no `match_about_blank`), but wrapping this call in try/catch with a `console.error` would match the injection error handling in `injectSdkScript` above.

src/content/index.ts
```diff
@@ -147,7 +166,9 @@ function cleanup() {
   document.removeEventListener(cleanupEventType, cleanup);
```
Comp0te: **Nit (pre-existing): listener removed from the wrong target.**

The cleanup listener is registered on `window` (line 173) but removed from `document` here, so this removal is a no-op and the listener stays registered forever. Harmless in practice since `cleanup()` is idempotent, but since this PR edits `cleanup()` anyway, it's a one-word fix: `window.removeEventListener(...)`.

Comp0te (commented):
Reviewed the transport change against the P0.2 item in `docs/refactoring/99-appendix-security-architecture.md`. The core design is sound: the forgeable `CasperWalletMethod:Request` listener is gone, the handshake is source+origin gated, request ids are CSPRNG-based, and I verified that a page-crafted message with a redux action `type` + `meta.requestId` is trapped by the `isSDKMethod` branch in the background router before `handleReduxAction` can see it. Findings below — one behavioral regression worth fixing before merge (pre-handshake requests hard-fail), the rest are cheap hardening/nits.

src/content/sdk.ts
ost-ptk: Fixed. Pre-handshake requests are now parked in a `portWaiters` queue and flushed in order once the handshake delivers the port, instead of hard-rejecting. Each parked request keeps its own per-request timeout as the bound, so a handshake that never completes surfaces as a normal `SDK RESPONSE TIMEOUT` rather than a hang. Added tests for the park/flush path and for the timeout-before-handshake case (a late handshake must not resurrect an already-timed-out request).

src/content/index.ts
ost-ptk: Fixed. The relay now checks membership in `SDK_REQUEST_TYPES` (a `ReadonlySet` of the 12 request-direction `.type` strings) in addition to `isSDKMethod`, so only genuine SDK requests can ride page → background. A `*:Response`/`*:Error` envelope or a redux action `type` is now dropped at the relay regardless of the background router branch ordering.

src/content/index.ts
ost-ptk: Fixed. On `runtime.sendMessage` failure the catch now posts a synthetic error envelope (`{ type: `${type}:Error`, payload, meta, error: true }`) back over the private port, so the SDK matches it by `requestId`, sees `error: true`, and rejects the dapp promise immediately instead of waiting out the per-request timeout.

src/content/index.ts
ost-ptk: Fixed. The handshake `postMessage` is now wrapped in try/catch with a `console.error`, matching the injection error handling in `injectSdkScript`, so an opaque-origin document (`window.location.origin === "null"`) no longer throws an uncaught `SyntaxError` out of the onload handler.

src/content/index.ts
ost-ptk: Fixed. `cleanup()` now calls `window.removeEventListener(...)`, matching where the listener is registered (line 173), so the removal is no longer a no-op.

Comp0te (approved):

PR #1396, opened 2026-07-08
refactor: DEP-99 — window state into Redux + duplicate-response root cause (WALLET-1343)

ost-ptk:
## DEP-99 · PR 2 of 8 — P0.5: window state into Redux + duplicate-response root cause

Moves approval-window state into the Redux `windowManagement` slice (retiring a parallel module-level
closure) and fixes the duplicate SDK-response race at its root by deduplicating server-side, keyed by
`requestId`, instead of relying on fragile per-page UI guards.

### Commits

| Commit     | Change                                                                                                                                                                                                                                                                                                                                         |
| ---------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `13f896c0` | `windowManagement` slice → request state machine: `requests: Record<requestId, 'open'\|'responded'\|'closed'>`, actions `windowRequestOpened`/`windowRequestResponded`/`windowClosed`, selector `selectRequestStatus`.                                                                                                                         |
| `f0c1ed2c` | Background `open-window.ts` sources/writes `windowId` through the slice (deletes the module-level `let windowId` closure — the second source of truth); the leaky per-creation `windows.onRemoved` listener (fired on _any_ window close) replaced by ONE init-time listener that dispatches `windowClosed()` only on id-match.                |
| `73060a62` | SDK responses rerouted UI→background→tab (`SDK_RESPONSE_TO_TAB`). The background dedupes by `requestId` — marks `responded` **optimistically before** the `await tabs.sendMessage`, so two interleaved responses (e.g. a sign followed by the `beforeunload` cancel) can't both deliver. `responseSentRef` removed from all 4 signature pages. |
| `9fda015e` | `emitSdkEventToActiveTabs`/`WithOrigin` now `await Promise.all` with per-tab error isolation (one tab lacking a content-script listener no longer aborts the broadcast).                                                                                                                                                                       |
| `9df9da11` | `sendSdkResponseToSpecificTab` restored to an always-resolves contract (a mid-flight SW teardown must not leave an approval window open on the `await` sites).                                                                                                                                                                                 |

### Dedupe design note

The guard drops a response **only when status is `'responded'`**, never on `'closed'`. A `beforeunload`
cancel and the `windows.onRemoved`→`windowClosed()` transition race; dropping on `'closed'` could
suppress a legitimate cancel for a window closed without a prior response. First-response-wins (drop on
`'responded'`, marked synchronously before the send) is the atomic, race-safe rule.

### Manifest / CSP / permissions

No change — this PR touches only `src/background/*` and the signature/connect UI pages; no manifest,
CSP, DNR, or content-script-entry edit. (extension-manifest-reviewer not required for this PR.)

### Verification

- ✅ `npm run ci-check` green (36 suites / 183 tests; tsc + lint clean).
- ✅ `npm run build:chrome` and `npm run build:firefox` (webpack prod) exit 0.
- ✅ Each task individually reviewed + a whole-branch review (Ready to merge).
- ⬜ **Manual QA (owner: reviewer)** — window lifecycle + dedupe are integration-level:
  1. **duplicate-cancel repro** — sign a deploy, let the approval window close → dapp gets EXACTLY ONE response (the signature), no trailing cancel.
  2. Cancel (close without signing) → exactly one cancel.
  3. All flows deliver: connect / sign / signMessage / signTypedData / decrypt / switchAccount.
  4. Approval window **reuse** (second sign while one is open reuses the window) and **fresh-after-close**; unrelated browser-window close does NOT clear tracking.
  5. Reload the approval window mid-flow → exactly one response reaches the dapp.
- ⬜ **Safari build** (`build:safari`) — needs Xcode, not run in CI here.

### Follow-up (non-blocking, from whole-branch review)

- `requests` map entries for connect/switch requests linger until SW restart (harmless — the slice is not persisted, so it resets each SW session). A future prune could drop terminal entries.
- **Dedupe is not restart-durable (acknowledged, low-likelihood residual).** Because `windowManagement` is not persisted, an MV3 service-worker restart in the millisecond gap between a sign response and the `beforeunload` cancel resets `requests` to `{}` — both responses then read status `undefined` and can be delivered, reintroducing the exact duplicate this PR removes. Same root cause as the linger note above (non-persisted slice). The old page-scoped `responseSentRef` survived SW restarts, so this is a slight robustness trade-off in that narrow window; keep-alive + the tiny sign→cancel gap make it unlikely in practice, and #1395's synthetic-error-on-send covers the request path, not this response path. Persisting terminal request statuses would close it if it ever surfaces.

### Review fixups (this branch, post-review)

- **Sender gate on the response reroute** (review #2, was optional): `handleSdkResponseToTab` now runs outside the uniform handler loop and is gated on `isTrustedUiSender(sender)` — mirroring `handleLegacyImport`. Belt-and-suspenders on top of the P0.2 private-channel + `SDK_REQUEST_TYPES` allowlist; an untrusted sender is silently dropped.
- **Nits** (review #3): `open-window.ts` fire-and-forget `createOpenWindow(...)(props)` now has a `.catch` (no unhandled rejection if `windows.create` fails); removed the dead trailing `return;` in both `emitSdkEventToActiveTabs*` `.map` callbacks in `utils.ts`.
- `npm run ci-check` green (36 suites / 184 tests; +1 untrusted-sender test).

### Base branch

Branches off `release/2.6.0` (DEP-16 already merged).

### DIFF
--- a/src/apps/signature-request/pages/decrypt-message/index.tsx
+++ b/src/apps/signature-request/pages/decrypt-message/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { shallowEqual, useSelector } from 'react-redux';
 
@@ -62,8 +62,6 @@ export function DecryptMessagePage() {
     );
   }
 
-  const responseSentRef = useRef(false);
-
   const accounts = useSelector(selectVaultAccounts, shallowEqual);
 
   const connectedAccountNames = useSelector(
@@ -117,8 +115,6 @@ export function DecryptMessagePage() {
   }, [requestOrigin, connectAnotherAccount, signingAccount.name]);
 
   const handleCancel = useCallback(() => {
-    if (responseSentRef.current) return;
-    responseSentRef.current = true;
     sendSdkResponseToSpecificTab(
       sdkMethod.decryptMessageResponse({ cancelled: true }, { requestId }),
       requestTabId
@@ -157,7 +153,6 @@ export function DecryptMessagePage() {
       return;
     }
 
-    responseSentRef.current = true;
     sendSdkResponseToSpecificTab(
       sdkMethod.decryptMessageResponse(
         { decryptedMessage, cancelled: false },
--- a/src/apps/signature-request/pages/sign-eip712/index.tsx
+++ b/src/apps/signature-request/pages/sign-eip712/index.tsx
@@ -3,13 +3,7 @@ import {
   IEIP712SignTypedDataOptions,
   IEIP712TypedData
 } from 'casper-wallet-core';
-import React, {
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState
-} from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { shallowEqual, useSelector } from 'react-redux';
 
@@ -73,7 +67,6 @@ export function SignEip712Page() {
   }
 
   const [showRawJson, setShowRawJson] = useState(false);
-  const responseSentRef = useRef(false);
 
   const accounts = useSelector(selectVaultAccounts, shallowEqual);
   const eip712JsonById = useSelector(selectEip712JsonById, shallowEqual);
@@ -185,8 +178,6 @@ export function SignEip712Page() {
   }, [requestOrigin, connectAnotherAccount, signingAccount.name]);
 
   const handleCancel = useCallback(() => {
-    if (responseSentRef.current) return;
-    responseSentRef.current = true;
     sendSdkResponseToSpecificTab(
       sdkMethod.signTypedDataResponse(
         {
@@ -204,7 +195,7 @@ export function SignEip712Page() {
   }, [requestId, requestTabId]);
 
   const handleSign = useCallback(() => {
-    if (!typedData || responseSentRef.current) {
+    if (!typedData) {
       return;
     }
 
@@ -224,7 +215,6 @@ export function SignEip712Page() {
         options
       });
 
-      responseSentRef.current = true;
       sendSdkResponseToSpecificTab(
         sdkMethod.signTypedDataResponse(
           {
@@ -241,7 +231,6 @@ export function SignEip712Page() {
       );
       closeCurrentWindow();
     } catch {
-      responseSentRef.current = true;
       sendSdkResponseToSpecificTab(
         sdkMethod.signTypedDataError(
           Error(ErrorMessages.signTypedData.SIGNING_FAILED.description),
--- a/src/apps/signature-request/pages/sign-message/index.tsx
+++ b/src/apps/signature-request/pages/sign-message/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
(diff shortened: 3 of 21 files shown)


Comp0te:
## Code review — P0.5 (window state → Redux + duplicate-response root cause)

Reviewed against the current base (`release/2.6.0` after #1395). Overall this is a clean, well-tested implementation that matches the P0.5 prescription: window state moves out of the module-level closure into the `windowManagement` slice, and the duplicate-cancel race is fixed at the root via server-side dedupe keyed by `requestId` (first response wins, marked `responded` synchronously before the `await tabs.sendMessage` — the atomicity reasoning is correct). Test coverage is genuinely good.

A few items, none blocking:

### 1. Dedupe state is not restart-durable (worth a note, low-likelihood residual)
`windowManagement` is not persisted, so an MV3 service-worker restart between the sign response and the `beforeunload` cancel resets `requests` to `{}` — both responses then read status `undefined` and can be delivered, reintroducing the exact duplicate this PR removes. The old per-page `responseSentRef` was page-scoped and survived SW restarts, so this is a slight robustness trade-off in that narrow window (keep-alive + the millisecond gap between sign and cancel make it unlikely in practice). #1395's synthetic-error-on-send-failure covers the page→background *request* path, not the UI→background→tab *response* path rerouted here, so it doesn't mitigate this case. Suggest just acknowledging it alongside the existing "requests linger until SW restart" follow-up.

### 2. `handleSdkResponseToTab` has no `sender` check (defense-in-depth, optional)
The handler reads `tabId`/`action` from the message and does `tabs.sendMessage(tabId, action)` with no verification the message came from the extension's own UI. With #1395 in the base this is **not reachable from a page** — the private `MessageChannel` transport plus the `SDK_REQUEST_TYPES` allowlist mean `SDK_RESPONSE_TO_TAB` can neither reach `port1` nor pass the request-type filter, independent of router branch ordering. So this is now belt-and-suspenders only: an optional `sender.id === runtime.id` assertion would guard against a future allowlist bug or content-script-world compromise. Fine to defer.

### 3. Nits
- `open-window.ts`: `openWindow` invokes `createOpenWindow(...)(props)` fire-and-forget with no `.catch`; if `windows.create` rejects, the request stays `'open'` forever and the promise rejects unhandled. Harmless to dedupe (only `'responded'` drops), but a `.catch` would tidy it.
- `utils.ts`: the trailing `return;` after `console.error('Tab without id…')` inside the `.map` callback is dead code.

Everything else looks correct, and the reroute composes correctly with #1395's private-port response path (all signature response types are in the content-script response switch). Good to merge once #1 is acknowledged.

ost-ptk:
Thanks for the thorough review. Addressed in `3b55703c`:

**#1 — dedupe not restart-durable:** acknowledged in the PR description (Follow-up section). Same root cause as the existing linger note (non-persisted `windowManagement`); leaving it as a documented, low-likelihood residual for now — persisting terminal request statuses would close it if it ever surfaces in practice. Agreed #1395 covers the request path, not this response path.

**#2 — `sender` check on `handleSdkResponseToTab`:** added. The handler is now pulled out of the uniform handler loop and gated on `isTrustedUiSender(sender)`, mirroring `handleLegacyImport`; an untrusted sender is silently dropped (`{ handled: true }`, no forward). Belt-and-suspenders on top of the P0.2 private channel + `SDK_REQUEST_TYPES` allowlist, exactly as you framed it.

**#3 — nits:**
- `open-window.ts`: added a `.catch` to the fire-and-forget `createOpenWindow(...)(props)` — a rejecting `windows.create` now logs instead of an unhandled rejection.
- `utils.ts`: removed the dead trailing `return;` (both `emitSdkEventToActiveTabs*` `.map` callbacks had it).

`npm run ci-check` green — 36 suites / 184 tests (+1 untrusted-sender case).

Comp0te (approved):
👍

PR #1398, opened 2026-07-08
fix: DEP-99 — SW keep-alive on chrome.alarms + persisted security deadlines (WALLET-1343)

ost-ptk:
## Summary

Part of the DEP-99 security & architecture backlog — **P0.3: MV3 service-worker lifetime hardening**.
Jira: https://make-software.atlassian.net/browse/WALLET-1343

Chrome kills the MV3 background service worker after ~30s of idle, which used to (a) silently kill the `setInterval`-based keep-alive, (b) drop the login-retry lockout and auto-lock countdowns, and (c) risk killing the SW mid crypto operation. Three fixes, one commit each (+ one review-driven hardening commit):

1. **Keep-alive on `chrome.alarms`** (`d3736480`, hardened in `e67c1dda`) — `setInterval` replaced with a `chrome.alarms` alarm (`casper-keep-alive`, 0.5 min, the MV3 minimum), which survives SW restarts. The `onAlarm` listener registers synchronously at module eval behind `isChromeBuild` — the gate is load-bearing: Firefox/Safari manifests don't declare the `alarms` permission, so the polyfill exposes no `alarms` namespace there. Start/stop are edge-triggered via a tri-state guard (`null` = fresh SW start), so per-dispatch store subscriptions no longer reset the alarm schedule, while the first call after a cold start always reaches the alarms API (a locked cold start must still clear a stale alarm surviving from the previous SW life).
2. **Lockout / auto-lock deadlines persisted as absolute timestamps** (`fb5f8343`) — on arming, `Date.now() + remaining` is written to `storage.local` under two new obfuscated keys (`get-main-store.ts`; key strings documented immutable). `startBackground` now also triggers the lockout saga, so after an SW restart the brute-force lockout timer resumes with the residual delay (or resets immediately if elapsed) instead of waiting for the next popup open. Deadlines are **derived, not authoritative**: if the deadline key is missing/cleared, the saga recomputes from the already-persisted lockout start time — clearing the new key cannot shorten a lockout. Keys are cleared on firing/reset (lockout) and on lock (auto-lock). Redux state shape unchanged.
3. **SW anchor during long crypto sagas** (`9ec18620`) — a refcounted `runtime.getPlatformInfo()` heartbeat (20s) held for the duration of `unlockVaultSaga` / `updateVaultCipher` / `createAccountSaga`, released in `finally` (verified against `takeLatest` cancellation — no wakelock leak). Includes a one-time `[keepalive] SW resumed mid-<flow>` cold-start breadcrumb (`console.debug`, flow name only — no secrets).

## Approved deviations from the plan

- **Port anchor → extension-API heartbeat (Task 3.3).** The plan's `runtime.connect`/Port anchor is a proven no-op on current Chrome: since Chrome 114 opening a port no longer resets the SW idle timer, and `runtime.onConnect` never fires in the caller's own context, so a self-connect from the SW immediately disconnects. Replaced (with approval) by the documented pattern: any extension-API call resets the idle timer (Chrome 110+), so a 20s `getPlatformInfo` heartbeat during long crypto work keeps the SW alive. Evidence and rationale are in the module doc comment of `sw-keep-alive-anchor.ts`.
- **Auto-lock deadline persistence is write-only today (Task 3.2).** The session slice (which holds `isLocked` and the decrypted vault) is not rehydrated after an SW restart, so the wallet always cold-starts locked and the auto-lock resume branch never runs in production. Kept per plan (decision: keep) — it is harmless (the key is cleared on every lock), and becomes functional automatically if session persistence ever lands.
- **`redux-saga-test-plan` added now** (devDependency; the plan formally introduces it in PR 8, but Task 3.2's saga tests need it). Adds 3 tiny transitive deps.

## Review round (Comp0te, 2026-07-08) — addressed in `f04cb620`

- **`minimum_chrome_version` 92 → 120** (`manifest.v3.json`): Chrome < 120 silently clamps `periodInMinutes: 0.5` to 1 minute, and a 60s pulse loses the MV3 SW at its ~30s idle deadline (silent session loss). ⚠️ Release management: this bump and the alarms keep-alive must ship in the **same store release** — do not cherry-pick one without the other. Permission/host delta: none in any manifest; Firefox `strict_min_version: 113.0` and the Safari manifest intentionally unchanged (persistent MV2 background pages, no `alarms` permission).
- Persisted lockout/auto-lock deadline reads now validated (`typeof` + `Number.isFinite`) — corrupt storage values fail closed into the full recompute instead of clearing the brute-force lockout.
- `clearAutoLockDeadline` moved after the `lockedEvent` emit in `lockVaultSaga` — a storage rejection can no longer swallow the dapp lock notification (order pinned by test).
- `timeoutCounterSaga` now clears a stale `AUTO_LOCK_DEADLINE_KEY` on a locked cold start (guard-fail branch), matching the lockout saga's hygiene.

## Known property (not a regression)

The 0.5 min alarm period equals Chrome's 30s idle deadline, so an occasional mid-idle SW death before the next pulse remains possible by design (0.5 min is the MV3 minimum). The failure mode degrades to pre-PR behavior: the SW restarts locked and the persisted deadlines now recover the security timers.

## Verification

- `npm run ci-check` — green (35 suites / 183 tests; tsc clean; lint 0 errors)
- `npm run build:chrome` / `build:firefox` / `build:safari` — all build (Safari: pre-existing CFBundleVersion warning only)
- `npm run e2e:chrome:headless:onboarding` — 6/6 passed
- `npm run e2e:chrome:headless:popup` — 46 passed, 1 skipped (pre-existing skip)
- New unit coverage: 9 keep-alive tests (incl. non-Chrome module-load regression test and edge-trigger guard), 9 vault-saga tests (redux-saga-test-plan; residual vs full delay, elapsed-deadline reset, storage effect args), 7 anchor tests (refcount, idempotent disposers, once-only breadcrumb)

### Manual QA checklist (reviewer/QA)

- [x] Unlock the wallet, trigger the login-retry lockout (5 wrong passwords), kill the SW via `chrome://serviceworker-internals` mid-countdown → the lockout timer resumes with the residual time (does not restart from full, cannot be bypassed) — _verified by owner 2026-07-08: unlock form reappears at the original T0+5:00 deadline after an SW kill mid-countdown_
- [x] With the wallet unlocked, confirm the SW stays alive (alarm pulses every 30s; no session loss during normal use) — _verified 2026-07-08: single edge-triggered `KeepAlive alarm started./stopped.` per transition, no per-dispatch log spam; the `Receiving end does not exist` ping error with no UI open is pre-existing behavior (same catch existed with setInterval)_
- [x] Unlock and watch the SW survive the unlock flow (breadcrumb `[keepalive] SW resumed mid-<flow>` appears in the SW console after an SW restart) — _verified 2026-07-08: `[keepalive] SW resumed mid-unlock` observed_
- [ ] Firefox and Safari builds load and background starts normally (no `alarms` usage there)

### DIFF
--- a/src/background/keep-alive.ts
+++ b/src/background/keep-alive.ts
@@ -1,4 +1,4 @@
-import { runtime } from 'webextension-polyfill';
+import { alarms, runtime } from 'webextension-polyfill';
 
 import { isChromeBuild } from '@src/utils';
 
@@ -7,23 +7,32 @@ import { selectKeysDoesExist } from '@background/redux/keys/selectors';
 import { selectVaultIsLocked } from '@background/redux/session/selectors';
 import { selectVaultCipherDoesExist } from '@background/redux/vault-cipher/selectors';
 
-let keepAliveInterval: ReturnType<typeof setInterval> | null = null;
+const KEEP_ALIVE_ALARM_NAME = 'casper-keep-alive';
 
-// Function to start the keep-alive interval
-export function startKeepAlive() {
-  if (!keepAliveInterval) {
-    keepAliveInterval = setInterval(keepAlive, 15000); // 15 seconds
-    console.log('KeepAlive interval started.');
-  }
+// chrome.alarms.onAlarm must be registered synchronously at module load so it
+// re-registers on every service worker cold start (MV3 requirement). The
+// isChromeBuild gate is load-bearing: only the Chrome (MV3) manifest declares
+// the `alarms` permission, so on Firefox/Safari the polyfill exposes no
+// `alarms` namespace and an ungated registration would crash the background
+// entry at load.
+if (isChromeBuild) {
+  alarms.onAlarm.addListener(alarm => {
+    if (alarm.name === KEEP_ALIVE_ALARM_NAME) {
+      keepAlive();
+    }
+  });
 }
 
-// Function to stop the keep-alive interval
+// Function to start the keep-alive alarm
+export function startKeepAlive() {
+  alarms.create(KEEP_ALIVE_ALARM_NAME, { periodInMinutes: 0.5 });
+  console.log('KeepAlive alarm started.');
+}
+
+// Function to stop the keep-alive alarm
 export function stopKeepAlive() {
-  if (keepAliveInterval) {
-    clearInterval(keepAliveInterval);
-    keepAliveInterval = null;
-    console.log('KeepAlive interval stopped.');
-  }
+  alarms.clear(KEEP_ALIVE_ALARM_NAME);
+  console.log('KeepAlive alarm stopped.');
 }
 
 // Function to check and manage the keep-alive mechanism based on vault state
--- a/src/background/keep-alive.ts
+++ b/src/background/keep-alive.ts
@@ -23,15 +23,31 @@ if (isChromeBuild) {
   });
 }
 
+// Edge-triggered guard: manageKeepAlive runs on every store update, and an
+// unguarded alarms.create would reset the alarm's schedule on each dispatch,
+// pushing the next fire to ~30s after the last update — exactly Chrome's idle
+// deadline. `null` means "unknown" (fresh SW start): alarms persist across SW
+// restarts, so the first start/stop call must always reach the alarms API —
+// create refreshes the surviving alarm, clear kills a stale one.
+let alarmActive: boolean | null = null;
+
 // Function to start the keep-alive alarm
 export function startKeepAlive() {
+  if (alarmActive === true) {
+    return;
+  }
   alarms.create(KEEP_ALIVE_ALARM_NAME, { periodInMinutes: 0.5 });
+  alarmActive = true;
   console.log('KeepAlive alarm started.');
 }
 
 // Function to stop the keep-alive alarm
 export function stopKeepAlive() {
+  if (alarmActive === false) {
+    return;
+  }
   alarms.clear(KEEP_ALIVE_ALARM_NAME);
+  alarmActive = false;
   console.log('KeepAlive alarm stopped.');
 }
 
--- a/src/background/keep-alive.ts
+++ b/src/background/keep-alive.ts
@@ -36,6 +36,10 @@ export function startKeepAlive() {
   if (alarmActive === true) {
     return;
   }
+  // 0.5 min is the alarms-API minimum only on Chrome >= 120; older Chromes
+  // silently clamp sub-minute periods to 1 minute, and a 60s pulse loses the
+  // service worker at its ~30s idle deadline (silent session loss). The
+  // manifest pins `minimum_chrome_version: 120` for exactly this reason.
   alarms.create(KEEP_ALIVE_ALARM_NAME, { periodInMinutes: 0.5 });
   alarmActive = true;
   console.log('KeepAlive alarm started.');
--- a/src/background/redux/sagas/vault-sagas.ts
+++ b/src/background/redux/sagas/vault-sagas.ts
@@ -1,4 +1,5 @@
 import { put, select, takeLatest } from 'redux-saga/effects';
+import { storage } from 'webextension-polyfill';
 
 import { getActiveAccountSupports, getUrlOrigin } from '@src/utils';
 
@@ -7,6 +8,10 @@ import {
   MapTimeoutDurationSettingToValue
 } from '@popup/constants';
 
+import {
+  AUTO_LOCK_DEADLINE_KEY,
+  LOGIN_RETRY_LOCKOUT_DEADLINE_KEY
+} from '@background/redux/get-main-store';
 import {
   loginRetryLockoutTimeReseted,
   loginRetryLockoutTimeSet
@@ -76,7 +81,7 @@ import {
 export function* vaultSagas() {
   yield takeLatest(lockVault.type, lockVaultSaga);
   yield takeLatest(
-    [loginRetryLockoutTimeSet.type, popupWindowInit.type],
+    [loginRetryLockoutTimeSet.type, popupWindowInit.type, startBackground.type],
     setDelayForLockoutVaultSaga
   );
   yield takeLatest(unlockVault.type, unlockVaultSaga);
@@ -114,12 +119,18 @@ export function* vaultSagas() {
 /**
  * on lock destroy session, vault and deploys
  */
-function* lockVaultSaga() {
+export function* lockVaultSaga() {
   try {
     yield put(sessionReseted());
     yield put(vaultReseted());
     yield put(deploysReseted());
     yield put(accountInfoReset());
+
+    // The vault is locked, so the persisted auto-lock deadline is spent —
+    // remove it so a stale value can never fire against a future session.
+    // `timeoutCounterSaga` locks by dispatching this same `lockVault` action,
+    // so this single site covers both the timeout and manual-lock paths.
+    yield* sagaCall(clearAutoLockDeadline);
 
     emitSdkEventToActiveTabs(() => {
       return sdkEvent.lockedEvent({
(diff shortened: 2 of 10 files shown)


src/background/keep-alive.ts
```diff
@@ -7,23 +7,48 @@ import { selectKeysDoesExist } from '@background/redux/keys/selectors';
...
+
+// Function to start the keep-alive alarm
 export function startKeepAlive() {
-  if (!keepAliveInterval) {
-    keepAliveInterval = setInterval(keepAlive, 15000); // 15 seconds
-    console.log('KeepAlive interval started.');
+  if (alarmActive === true) {
+    return;
   }
+  alarms.create(KEEP_ALIVE_ALARM_NAME, { periodInMinutes: 0.5 });
```
Comp0te, line 43: **Chrome < 120 clamps `periodInMinutes: 0.5` to 1 minute — silent session-loss regression on older Chrome.**

The 30-second alarm minimum only exists since Chrome 120 (Dec 2023); on earlier versions a packed extension's sub-minimum period is clamped to 1 minute. `manifest.v3.json` currently declares no `minimum_chrome_version`, so on Chrome 88–119 the pulse becomes 60s > the 30s idle deadline: an idle unlocked wallet's SW dies at ~30s and the in-memory session is lost — i.e. the wallet self-locks ~30s after the last activity. That's a regression vs `setInterval(15s)`, which kept an alive SW alive indefinitely.

The affected population is small (~1–3% of desktop Chrome, dominated by Chrome 109 permanently stranded on Windows 7/8.1) and arguably shouldn't be running a self-custody wallet on an unpatched EOL browser anyway — so the clean fix is to add `"minimum_chrome_version": "120"` to `manifest.v3.json` in this PR rather than ship the silent regression. Edge/Brave inherit the alarms behavior from their underlying Chromium version and honor `minimum_chrome_version`, so the floor behaves consistently there.

Also worth tweaking the comment here: 0.5 min is the minimum only on Chrome ≥ 120.

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -134,36 +146,75 @@ function* lockVaultSaga() {
...
+
+  if (action.type === loginRetryLockoutTimeSet.type) {
+    // Arming: compute and persist the absolute deadline.
+    deadline = loginRetryLockoutTime + LOCK_VAULT_TIMEOUT;
+    yield* sagaCall(writeLockoutDeadline, deadline);
+  } else {
+    // Resume: read the persisted deadline, falling back to a computed one.
+    const stored = yield* sagaCall(readLockoutDeadline);
+    deadline =
+      (stored[LOGIN_RETRY_LOCKOUT_DEADLINE_KEY] as number | undefined) ??
```
Comp0te: **Persisted lockout deadline is trusted without type validation — fails open on garbage.**

`??` only catches nullish: any other corrupt value under this key (string, `0`, negative, object) passes through, `timeLeft` becomes `NaN` or negative, `timeLeft > 0` is false, and the saga immediately dispatches `loginRetryCountReseted`/`loginRetryLockoutTimeReseted` — clearing the brute-force lockout.

Only extension code can write `storage.local`, so exploiting this needs a second bug — but that's the same single-point-of-failure pattern the P0.2 note in the security appendix warns about. Since this value guards the brute-force lockout, validate defensively:

```ts
const raw = stored[LOGIN_RETRY_LOCKOUT_DEADLINE_KEY];
deadline =
  typeof raw === 'number' && Number.isFinite(raw)
    ? raw
    : loginRetryLockoutTime + LOCK_VAULT_TIMEOUT;
```

(The auto-lock counterpart fails closed — garbage → immediate `lockVault()` — so it's fine as is, though the same check would keep the two paths symmetric.)

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -114,13 +120,19 @@ export function* vaultSagas() {
...
     yield put(sessionReseted());
     yield put(vaultReseted());
     yield put(deploysReseted());
     yield put(accountInfoReset());
 
+    // The vault is locked, so the persisted auto-lock deadline is spent —
+    // remove it so a stale value can never fire against a future session.
+    // `timeoutCounterSaga` locks by dispatching this same `lockVault` action,
+    // so this single site covers both the timeout and manual-lock paths.
+    yield* sagaCall(clearAutoLockDeadline);
```
Comp0te: **Deadline clear is now the only fallible step between the state resets and the SDK `lockedEvent` emit.**

If `storage.local.remove` rejects, the `catch` swallows it and `emitSdkEventToActiveTabs(lockedEvent)` never runs — connected dapps aren't told the wallet locked. Before this PR nothing fallible sat between the resets and the emit. Suggest moving the clear after the emit (or giving it its own try/catch).

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -265,28 +335,41 @@ function* timeoutCounterSaga() {
...
-
-      if (timeoutExpired) {
-        yield put(lockVault());
+      let deadline: number;
+
+      if (action.type === startBackground.type) {
+        // Resume: read the persisted deadline, falling back to a computed one.
+        const stored = yield* sagaCall(readAutoLockDeadline);
+        deadline =
+          (stored[AUTO_LOCK_DEADLINE_KEY] as number | undefined) ??
```
Comp0te: Nit: unlike `setDelayForLockoutVaultSaga` (which clears its stale key on `startBackground` when no lockout is active), a stale `AUTO_LOCK_DEADLINE_KEY` survives a locked cold start — the `!vaultIsLocked` guard returns before storage is touched. Harmless today (the resume path can't run while locked, and arming overwrites), but a one-line clear in the locked branch would remove a stale value that only stays safe by accident of the current guard order.

Comp0te (commented):
Reviewed against the P0.3 item in `docs/refactoring/99-appendix-security-architecture.md`.

**Compliance:** alarms-based pulse ✅; persisted absolute deadlines with resume-on-`startBackground` ✅; breadcrumb ✅; Port anchor replaced by the extension-API heartbeat — the deviation is justified (since Chrome 114 an open port no longer resets the SW idle timer) and well documented in `sw-keep-alive-anchor.ts`.

Verified the load-bearing security claims: `loginRetryLockoutTime` is persisted independently (`LOGIN_RETRY_LOCKOUT_KEY`), so deleting the new deadline key recomputes from the lockout start and cannot shorten a lockout; both MV2 manifests are `persistent: true`, so the anchor's no-op claim holds; `startKeepAlive`/`stopKeepAlive` are only reachable through the `isChromeBuild`-gated `initKeepAlive`, so the unguarded `alarms.*` calls can't fire on Firefox/Safari; the `finally`-based anchor release is safe under `takeLatest` cancellation.

Solid implementation with strong test coverage. One change requested before merge (the `minimum_chrome_version` floor — see inline), one cheap hardening of the lockout timer worth doing in the same PR, and two minor notes.

src/background/keep-alive.ts
ost-ptk: Fixed in f04cb620 — `minimum_chrome_version` raised 92 → 120 in `manifest.v3.json` (the 92 floor had landed via PR #1397's merge-forward just after your review), and the comment above `alarms.create` now spells out the pre-120 clamp and cross-references the manifest pin. Firefox `strict_min_version: 113.0` / Safari intentionally unchanged (persistent MV2 background pages, no `alarms` permission). Noted for release management: this bump must ship in the same store release as the 0.5-min alarm — no cherry-picking the alarm change without it.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in f04cb620 — the resume read is now `typeof raw === 'number' && Number.isFinite(raw) ? raw : loginRetryLockoutTime + LOCK_VAULT_TIMEOUT`, so corrupt values fall back to the full recompute (fail closed) instead of clearing the lockout. The auto-lock read path got the identical validation for symmetry. Covered by parametrized tests (string / NaN) asserting the full `LOCK_VAULT_TIMEOUT` residual is re-armed, not just "no reset".

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in f04cb620 — `clearAutoLockDeadline` moved after the `lockedEvent` emit, so a storage rejection can no longer swallow the dapp notification; `tabs.query` is invoked synchronously before the storage call yields. The ordering is pinned by a test comparing `invocationCallOrder` (tabs.query < storage.remove).

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in f04cb620 — `timeoutCounterSaga`'s guard-fail branch (vault absent / locked / no activity time) now clears `AUTO_LOCK_DEADLINE_KEY`, matching the lockout saga's stale-key hygiene, with a covering test. The clear can't race the arming write: arming requires exactly the unlocked+active state this branch negates.

Comp0te (approved):
👍

PR #1399, opened 2026-07-09
feat: DEP-99 — saga error channel + tab-id fallback (WALLET-1343)

ost-ptk:
## DEP-99 PR4 — saga error channel (P1.2 + P1.4)

Part of the DEP-99 security & architecture backlog ([WALLET-1343](https://make-software.atlassian.net/browse/WALLET-1343)). Fourth of the thematic PRs; targets `release/2.6.0`.

Before this change, background sagas that failed only `console.error`'d — the failure was invisible to the user and to any dapp waiting on a response. This PR adds a non-fatal error surface and closes two silent-failure gaps.

### What's in it

**P1.2 — saga error channel + non-fatal banner**

- New `sagaError` / `dismissSagaError` actions and `selectSagaErrors` selector on the `appEvents` slice. `errors` is capped to the last 10, uses a monotonic id counter, and is **runtime-only** — it is deliberately excluded from `storage.local` persistence (only `dismissedEventIds` round-trips), so a stale error never reappears after a service-worker restart.
- All 10 background saga catch blocks (`vault-sagas`, `onboarding-sagas`, `check-casper2-network-saga`) now `yield put(sagaError(...))` alongside the existing `console.error`.
- A dismissable `SagaErrorBanner` renders `selectSagaErrors` from the read-only replica; dismiss dispatches `dismissSagaError(id)` back to the background store. It is mounted in every UI app (popup, onboarding, connect-to-app, signature-request, import-account-with-file), so an error raised while any window is open surfaces there.

**P1.4 — stale/dead tab-id delivery fallback**

- When the background delivers an SDK **response** to the originating dapp tab and that tab is gone (closed / no listener) or the captured `tabId` is invalid, it now recovers the dapp origin from the request window's URL and re-delivers to any active same-origin tab via `emitSdkEventToActiveTabsWithOrigin`, surfacing a non-fatal `sagaError`. The existing atomic response-dedup (mark-responded-before-`await`) and the `isTrustedUiSender` gate are preserved.

**Private-state resubscribe (PR #1394 review #2)**

- The UI used to snapshot private state once on mount and never refresh, so a password change in one window left another window validating the old hash. The background now emits a **payload-free** `privateStateUpdated` signal when private material changes, and `usePrivateState` re-fetches on receipt (reusing the existing timeout + bounded-retry path). No secret material is ever placed in the broadcast (P0.1 preserved).

### Scope / deviations from the appendix

- **Sentry is a deliberate NON-GOAL.** P1.2 in the appendix reads "…+ Sentry", but there is no Sentry SDK in the repo and adding telemetry to a self-custody wallet is a product decision — so P1.2 here is scoped to the in-app error channel + banner only. This is an intentional omission, not an oversight.
- **P1.4 `tabs.get` pre-check deviation.** The appendix suggests pre-checking `tabs.get(tabId)`; instead the handler catches the `tabs.sendMessage` rejection and falls back. This is atomic (no TOCTOU gap between the check and the send) and saves a round-trip, while detecting the same dead/stale-tab condition.

### Security notes

- No private material (`passwordHash` / salts / `vaultCipher`) appears in any broadcast to UI replicas; the `privateStateUpdated` signal carries nothing but its type string.
- The `sagaError` surfaced on a failed SDK delivery references only the `tabId` and a static reason — never the response action or its payload (which may contain a signature / encrypted message).
- **Intentional trade-off:** the same-origin delivery fallback broadcasts the response to _all_ active tabs of the requesting dapp's origin, not only the original tab. This is bounded by an exact origin match, and only the tab with the matching pending `requestId` resolves it — a best-effort recovery, not a cross-origin leak.

### Review response (Comp0te)

All findings from the first review are addressed:

- **Critical — banner un-dismissable** (`dismissSagaError` missing from `FORWARDED_ACTION_TYPES`): fixed + routing-level test added.
- **Medium — misleading fallback message**: message now reflects whether the fallback actually delivered.
- **Medium — mark-responded on zero delivery / unguarded emit**: `emitSdkEventToActiveTabsWithOrigin` returns a delivered count (mark responded only when `> 0`); the fallback emit is wrapped so a `tabs.query` rejection can't reject the handler.
- **Medium — banner popup-only**: now mounted in all UI apps (onboarding-saga errors surface).
- **Low — id-reuse race**: replaced derived id with a monotonic counter + regression test.
- **Minors**: safe `errorToMessage` for non-`Error` throws (all 10 catch blocks); translated dismiss `aria-label`/`title`; banner now full-width (removed the contradictory fixed `360px`).

### Verification

- `npm run ci-check` green (256 tests, tsc clean, 0 lint errors).
- Chrome + Firefox production builds succeed.
- Whole-branch review: ready to merge (no Critical/Important findings); first external review fixes applied.

### Owner-gated (not done here)

- Safari build (requires macOS + Xcode).
- Manual QA: (a) force a saga error → banner shows + **dismisses** (validates the Critical fix); (b) force an onboarding-saga error → banner renders in the onboarding window; (c) close the dapp tab before approving a request → response reaches a same-origin tab; (d) change password in one window with `password-protection-page` mounted in another → the mounted page accepts the NEW password.
- Full Playwright e2e (onboarding + popup) via CI.

### Notes

- New user-facing strings ride the English fallback (`locale:extract_pot` intentionally not run — the stale catalog would add unrelated diff).

### DIFF
--- a/src/apps/popup/index.tsx
+++ b/src/apps/popup/index.tsx
@@ -27,6 +27,7 @@ import {
   darkTheme,
   lightTheme
 } from '@libs/ui';
+import { SagaErrorBanner } from '@libs/ui/components/saga-error-banner/saga-error-banner';
 
 import { AppRouter } from './app-router';
 
@@ -68,6 +69,7 @@ const Tree = () => {
         <ThemeProvider theme={isDarkMode ? darkTheme : lightTheme}>
           <GlobalStyle />
           <ReduxProvider store={store}>
+            <SagaErrorBanner />
             <QueryClientProvider client={newQueryClient}>
               <ErrorBoundary>
                 <AppRouter />
--- a/src/background/handlers/sdk-response-to-tab.ts
+++ b/src/background/handlers/sdk-response-to-tab.ts
@@ -30,14 +30,38 @@ function recoverDappOrigin(url: string | undefined): string | null {
   }
 }
 
+// Same-origin delivery fallback. Returns the number of tabs the response was
+// SUCCESSFULLY delivered to (0 when there is no recoverable origin, or when the
+// broadcast matched no active same-origin tab). Wrapping the emit isolates a
+// `tabs.query`-level rejection so the caller's error-surface dispatch still runs
+// and the handler never rejects back to the signature page.
+async function deliverViaOrigin(
+  origin: string | null,
+  action: SdkResponseToTabMessage['action']
+): Promise<number> {
+  if (!origin) {
+    return 0;
+  }
+  try {
+    return await emitSdkEventToActiveTabsWithOrigin(origin, action);
+  } catch {
+    // tabs.query / query-level failure → nothing delivered.
+    return 0;
+  }
+}
+
 // SECURITY: the dapp `action` is the SDK response and may carry secret material
 // (a `signatureHex` / `encryptedMessage` / signed payload). This surfaced error
 // must reference ONLY the tabId + a static reason — never the action or any
-// part of its payload.
-function deliveryFailedError(tabId: unknown) {
+// part of its payload. The message is parameterized on whether the same-origin
+// fallback actually delivered, so a support reader is not told the response was
+// recovered when it was in fact lost.
+function deliveryFailedError(tabId: unknown, fallbackDelivered: boolean) {
   return sagaError({
     source: 'sdk-response-to-tab',
-    message: `SDK response delivery to tab ${tabId} failed; used same-origin fallback`
+    message: fallbackDelivered
+      ? `SDK response delivery to tab ${tabId} failed; delivered via same-origin fallback`
+      : `SDK response delivery to tab ${tabId} failed; no same-origin fallback available — response not delivered`
   });
 }
 
@@ -96,18 +120,17 @@ export async function handleSdkResponseToTab(
   const validTab = Number.isInteger(tabId) && tabId >= 0;
 
   if (!validTab) {
-    // No usable tab. If we can recover the dapp origin, deliver via the
-    // same-origin fallback (broadcast to any active tab of that origin) and
-    // mark responded — we ARE delivering, so a later duplicate must dedupe.
-    // With no origin, nothing is delivered: do NOT mark responded (a valid
-    // retry must still deliver). Either way surface the non-fatal error.
-    if (origin) {
-      if (requestId != null) {
-        store.dispatch(windowRequestResponded({ requestId }));
-      }
-      await emitSdkEventToActiveTabsWithOrigin(origin, action);
+    // No usable tab. Attempt the same-origin fallback (broadcast to any active
+    // tab of the recovered dapp origin). Only mark responded when the fallback
+    // ACTUALLY delivered to at least one tab — we ARE delivering, so a later
+    // duplicate must dedupe. When nothing was delivered (no origin, or the
+    // broadcast matched zero tabs), do NOT mark responded: a valid retry must
+    // still be able to deliver. Either way surface the non-fatal error.
+    const delivered = await deliverViaOrigin(origin, action);
+    if (delivered > 0 && requestId != null) {
+      store.dispatch(windowRequestResponded({ requestId }));
     }
-    store.dispatch(deliveryFailedError(tabId));
+    store.dispatch(deliveryFailedError(tabId, delivered > 0));
     return { handled: true, response: undefined };
   }
 
@@ -129,12 +152,11 @@ export async function handleSdkResponseToTab(
     await tabs.sendMessage(tabId, action);
   } catch {
     // Tab gone / no listener → try the same-origin fallback (if we can recover
-    // the origin) and surface the non-fatal error. The optimistic mark above
-    // already deduped any retry.
-    if (origin) {
-      await emitSdkEventToActiveTabsWithOrigin(origin, action);
-    }
-    store.dispatch(deliveryFailedError(tabId));
+    // the origin) and surface the non-fatal error, choosing the message based
+    // on whether the fallback actually delivered. The optimistic mark above
+    // already deduped any retry (delivery failure is terminal by design).
+    const delivered = await deliverViaOrigin(origin, action);
+    store.dispatch(deliveryFailedError(tabId, delivered > 0));
   }
 
   return { handled: true, response: undefined };
--- a/src/background/handlers/sdk-response-to-tab.ts
+++ b/src/background/handlers/sdk-response-to-tab.ts
@@ -1,5 +1,6 @@
 import { Runtime, tabs } from 'webextension-polyfill';
 
+import { sagaError } from '@background/redux/app-events/actions';
 import { MainStore } from '@background/redux/get-main-store';
 import { windowRequestResponded } from '@background/redux/windowManagement/actions';
 import { selectRequestStatus } from '@background/redux/windowManagement/selectors';
@@ -7,9 +8,38 @@ import {
   SDK_RESPONSE_TO_TAB,
   SdkResponseToTabMessage
 } from '@background/send-sdk-response-to-specific-tab';
+import { emitSdkEventToActiveTabsWithOrigin } from '@background/utils';
 
 import { isTrustedUiSender } from './private-state';
 import { HandlerResult } from './types';
+
+// Recover the originating dapp origin from the sender page URL. The response
+// windows are opened with `?origin=<dappOrigin>` in their query string (e.g.
+// `signature-request.html?requestId=..&origin=<dappOrigin>&tabId=..#/..`), and
+// `sender.url` IS that page URL. `isTrustedUiSender` has already guaranteed
+// `sender.url` is a defined extension URL by the time we parse it. Returns null
+// if the param is absent or the URL is unparseable.
+function recoverDappOrigin(url: string | undefined): string | null {
+  if (!url) {
+    return null;
+  }
+  try {
+    return new URL(url).searchParams.get('origin');
+  } catch {
+    return null;
+  }
+}
+
+// SECURITY: the dapp `action` is the SDK response and may carry secret material
+// (a `signatureHex` / `encryptedMessage` / signed payload). This surfaced error
+// must reference ONLY the tabId + a static reason — never the action or any
+// part of its payload.
+function deliveryFailedError(tabId: unknown) {
+  return sagaError({
+    source: 'sdk-response-to-tab',
+    message: `SDK response delivery to tab ${tabId} failed; used same-origin fallback`
+  });
+}
 
 // Server-side dedupe of SDK responses (P0.5 root cause). The signature UI pages
 // used to `tabs.sendMessage` the response to the dapp tab directly and guard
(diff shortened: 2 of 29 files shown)


src/libs/ui/components/saga-error-banner/saga-error-banner.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const { t } = useTranslation();
+
+  const errors = useSelector(selectSagaErrors);
+
+  if (errors.length === 0) {
+    return null;
+  }
+
+  const handleDismiss = (id: SagaError['id']) => {
+    dispatchToMainStore(dismissSagaError(id));
```
Comp0te, line 70: 🔴 **Critical — this dispatch can never reach the background store, so errors are un-dismissable.**

`dispatchToMainStore` sends the action via `runtime.sendMessage`, and the background only forwards UI-originated actions listed in `FORWARDED_ACTION_TYPES` (`src/background/handlers/redux-actions.ts`) — this PR never adds `dismissSagaError` there. The message falls through every handler in the router and hits `throw Error('Background: Unknown redux action: ...')` (`src/background/index.ts`), which `dispatchToMainStore`'s `.catch` reduces to a `console.error`.

**Net effect:** the moment any saga error fires, this fixed `z-index: tooltip` banner permanently overlays the top of the popup and cannot be dismissed until the service worker restarts.

It slipped through because the tests exercise the reducer directly, never the message routing.

**Fix:** add `dismissSagaError` to the import and to `FORWARDED_ACTION_TYPES` in `redux-actions.ts`. Ideally also add a routing-level test (`handleReduxAction(dismissSagaError(1), store)` → `handled: true`, dispatch called) so this allowlist-gap class is covered for future actions.

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -1,16 +1,46 @@
...
+}
+
+// SECURITY: the dapp `action` is the SDK response and may carry secret material
+// (a `signatureHex` / `encryptedMessage` / signed payload). This surfaced error
+// must reference ONLY the tabId + a static reason — never the action or any
+// part of its payload.
+function deliveryFailedError(tabId: unknown) {
+  return sagaError({
+    source: 'sdk-response-to-tab',
+    message: `SDK response delivery to tab ${tabId} failed; used same-origin fallback`
```
Comp0te: 🟡 **Misleading user-facing message.** This static text claims "used same-origin fallback", but `deliveryFailedError` is also dispatched in the branches where **no fallback ran** (`origin` is null — both the invalid-tab path and the catch path). A user or support engineer reading the banner will believe delivery was recovered when the response was actually lost.

Parameterize it, e.g. `fallback used` vs `no fallback available — response not delivered`.

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -62,9 +92,22 @@ export async function handleSdkResponseToTab(
...
+    // No usable tab. If we can recover the dapp origin, deliver via the
+    // same-origin fallback (broadcast to any active tab of that origin) and
+    // mark responded — we ARE delivering, so a later duplicate must dedupe.
+    // With no origin, nothing is delivered: do NOT mark responded (a valid
+    // retry must still deliver). Either way surface the non-fatal error.
+    if (origin) {
+      if (requestId != null) {
+        store.dispatch(windowRequestResponded({ requestId }));
+      }
+      await emitSdkEventToActiveTabsWithOrigin(origin, action);
```
Comp0te: 🟡 Two issues in this branch:

1. **"we ARE delivering" overstates.** The request is marked `responded` before this call, but `emitSdkEventToActiveTabsWithOrigin` only targets `active: true` tabs and may match **zero** tabs — nothing delivered, yet a legitimate retry is now permanently deduped. The old invalid-tab path deliberately did *not* mark responded. Consider having the emit util return a delivered count and only marking when `> 0` (it already knows how many tabs matched).

2. **Unguarded `await`.** If `tabs.query` inside the emit rejects, the `store.dispatch(deliveryFailedError(...))` below is skipped and the handler rejects back to the signature page — the one case where the error surface itself fails. Wrap the emit in try/catch (same applies to the catch-path call at L135).

src/apps/popup/index.tsx
```diff
@@ -68,6 +69,7 @@ const Tree = () => {
         <ThemeProvider theme={isDarkMode ? darkTheme : lightTheme}>
           <GlobalStyle />
           <ReduxProvider store={store}>
+            <SagaErrorBanner />
```
Comp0te, line 72: 🟡 **Banner is popup-only, but several of the wired catch blocks fire when the popup isn't open.** Onboarding saga errors (`initKeysSage`, `initVaultSaga`, `recoverVaultSaga`, `resetVaultSaga`) happen while only the onboarding app is open — nothing renders them there, and being runtime-only they're lost on SW restart before the user ever opens the popup. If popup-only is the intended P1.2 scope, say so in the PR description; otherwise mount the banner in the onboarding / signature-request apps too.

src/background/redux/app-events/reducer.ts
```diff
@@ -16,9 +17,42 @@ const slice = createSlice({
...
+    sagaError: (
+      state,
+      action: PayloadAction<{
+        source: string;
+        message: string;
+        code?: string;
+      }>
+    ) => {
+      const nextId = state.errors.length
+        ? Math.max(...state.errors.map(error => error.id)) + 1
```
Comp0te: 🔵 **Id reuse race.** Deriving `nextId` from the max of *remaining* entries means dismissing the highest-id error lets the next error reuse that id — and a dismiss that's still in flight (popup → background round-trip) can then delete the *new* error. A monotonically increasing counter stored in the slice avoids this.

Comp0te (changes requested):
Reviewed against `docs/refactoring/99-appendix-security-architecture.md` (P1.2 / P1.4 + the PR #1394 review #2 follow-up), with a focus on bugs and security.

The architecture is right: errors live in the background store (single-writer invariant respected), the banner renders from the read-only replica, the private-state signal is payload-free, and the same-origin fallback is origin-exact and gated on `isTrustedUiSender`. But there is **one blocking functional bug** — the banner's dismiss button cannot work (see inline comment in `saga-error-banner.tsx`) — plus a few medium items inline.

### Security assessment (verified, not just taken from the description)

- **Origin recovery chain is sound.** The `?origin=` query param is written by the background itself from the validated dapp sender URL when opening request windows (`handlers/sdk-methods.ts`); `handleSdkResponseToTab` is gated on `isTrustedUiSender` before parsing; extension UI pages were already trusted to name arbitrary `tabId`s, so parsing their URL grants no new capability.
- **Fallback broadcast is bounded**: `getUrlOrigin(tab.url) === origin`, http(s) only, active tabs only. A response reaching a *different same-origin tab* is the documented trade-off — acceptable.
- **No private material in broadcasts**: `privateStateUpdated` is payload-free (tested); `appEvents.errors` is genuinely runtime-only — persistence writes only `dismissedEventIds` and rehydration forces `errors: []` (verified in `get-main-store.ts`).
- **No new attack surface**: web pages can't send internal runtime messages, so neither `sagaError` injection nor a spoofed `privateStateUpdated` (which only triggers a re-fetch that is itself `isTrustedUiSender`-gated) is reachable from a dapp.
- The surfaced delivery error references only `tabId` + static text; tests assert signature material never appears. 👍

### Minor (no inline anchor)

- `String(err)` renders `[object Object]` for non-Error throws; `err instanceof Error ? err.message : String(err)` is safer for a user-facing string (applies to all 10 saga catch blocks).
- `aria-label={'dismiss'}` / `title={'dismiss'}` are untranslated literals (the heading correctly uses `Trans`).
- `BannerContainer` sets both `width: 360px` and `left: 0; right: 0` — contradictory; the width wins. Cosmetic.
- P1.2 spec says "…+ Sentry"; there's no Sentry SDK in the repo, so that half is silently dropped — worth an explicit backlog note rather than implicit omission.
- P1.4 spec says "pre-check `tabs.get(tabId)`" — the catch-on-send + fallback implementation is arguably better (atomic), but the deviation deserves a line in the description.

### Spec compliance

P1.2 ✅ (minus Sentry, n/a) · P1.4 ✅ (deliberate deviation) · #1394 review #2 ✅ (clean payload-free signal + re-fetch).

src/libs/ui/components/saga-error-banner/saga-error-banner.tsx
ost-ptk: Good catch — confirmed and fixed. `dismissSagaError` was never added to `FORWARDED_ACTION_TYPES`, so the dismiss message fell through the router to the `Unknown redux action` throw and the banner was un-dismissable until an SW restart. Fixed in `69eb6265`: added `dismissSagaError` to the allowlist (and its import) in `handlers/redux-actions.ts`, plus a routing-level test (`redux-actions.test.ts`) asserting `handleReduxAction(dismissSagaError(1), store)` → `{ handled: true }` + `dispatch` called, and that an unknown type still returns `{ handled: false }` — so this allowlist-gap class is now covered. `sagaError` itself is intentionally not forwarded (it's dispatched by background sagas via `put`).

src/background/handlers/sdk-response-to-tab.ts
ost-ptk: Fixed in `4b565c33`. `deliveryFailedError` now takes a `fallbackDelivered` flag and emits one of two messages: `…failed; delivered via same-origin fallback` vs `…failed; no same-origin fallback available — response not delivered`. Both call sites pass `delivered > 0`, so the banner no longer claims recovery when the response was actually lost. (Message still references only `tabId` + static text — no action/payload.)

src/background/handlers/sdk-response-to-tab.ts
ost-ptk: Both fixed in `4b565c33`.

1. `emitSdkEventToActiveTabsWithOrigin` now returns the count of successful deliveries (incremented only after a resolved `tabs.sendMessage`). The invalid-tab branch marks `windowRequestResponded` only when `delivered > 0`, so a legitimate retry is no longer deduped when zero active tabs matched. The valid-tab path keeps its optimistic mark-before-`await` (the atomic dedup) unchanged.
2. Both fallback emits go through a `deliverViaOrigin` helper that wraps the emit in try/catch and returns `0` on throw — so a `tabs.query` rejection no longer skips the `sagaError` dispatch or rejects the handler; it still resolves `{ handled: true }` and surfaces the not-delivered error. New tests cover delivered>0 / =0 / emit-throws on both branches.

src/apps/popup/index.tsx
ost-ptk: Fixed in `3eef4934`: `SagaErrorBanner` is now mounted in all UI app entries — `onboarding`, `connect-to-app`, `signature-request`, and `import-account-with-file` — inside each app's `ReduxProvider` (same pattern as popup). So onboarding-saga errors (`initKeysSage`/`initVaultSaga`/`recoverVaultSaga`/`resetVaultSaga`) now render in the onboarding window.

src/background/redux/app-events/reducer.ts
ost-ptk: Fixed in `d592135b`: replaced the `Math.max(existing)+1` derivation with a monotonic `nextErrorId` counter stored in the slice (runtime-only, like `errors` — not persisted; rehydrate normalizes it to `0`). Added a regression test: add ids 0,1 → dismiss the highest (1) → the next error gets id `2`, not the freed-up `1`, so an in-flight dismiss can no longer delete a reused id.

ost-ptk:
Thanks for the thorough pass — all findings are addressed (commits `69eb6265` / `4b565c33` / `d592135b` / `3eef4934` / `fe536df4`, replies inline on each thread). The Minors:

- **`String(err)` → `[object Object]`**: fixed across all 10 saga catch blocks in `fe536df4` via a shared `errorToMessage(err)` helper (`err instanceof Error ? err.message : String(err)`), with unit tests.
- **`aria-label`/`title` untranslated**: fixed in `3eef4934` — now `t('Dismiss')`.
- **`BannerContainer` width vs `left/right`**: fixed in `3eef4934` — removed `width: 360px`, kept `left: 0; right: 0` so the banner spans the full window width (also resolves the "not full-tab-spanning" note).
- **Sentry half of P1.2**: deliberate NON-GOAL, not a silent omission — no Sentry SDK in the repo, and adding telemetry to a self-custody wallet is a product decision, so P1.2 was scoped to the error channel + banner. Documented in the PR description.
- **P1.4 `tabs.get` pre-check deviation**: the catch-on-send + fallback is atomic (no TOCTOU) and one fewer round-trip; documented in the PR description.

`ci-check` green (256 tests, tsc clean, 0 lint errors).

Comp0te (approved):

PR #1400, opened 2026-07-09
fix: DEP-99 — lottie-web swap + nonce-based style CSP (WALLET-1343)

ost-ptk:
## What & why

Part of **DEP-99 / P1.5(a)** of the dependency-modernization & security-hardening ticket
[WALLET-1343](https://make-software.atlassian.net/browse/WALLET-1343).

This PR swaps the `@lottiefiles/react-lottie-player` React wrapper for the underlying
`lottie-web` library behind a thin local `<LottiePlayer>` component, which removes the last
nonce-incompatible `<style>`-injector and unblocks tightening the Chrome-production style CSP:
`style-src 'unsafe-inline'` → a build-time **nonce**.

## Changes

- **`refactor: swap react-lottie-player for lottie-web`** — new `src/libs/ui/components/lottie-player`
  wrapper (`lottie.loadAnimation({ renderer: 'svg', … })`, destroyed on unmount, reloaded when
  `src`/`loop`/`autoplay` change); all 8 call-sites migrated. `lottie-web` pinned exact `5.13.0`.
- **`fix: nonce-based style-src CSP`** — one nonce per build (`crypto.randomBytes`), wired identically
  to the manifest CSP, `__webpack_nonce__` (style-loader, via a first array-entry `set-webpack-nonce.ts`)
  and `process.env.CSP_NONCE` (styled-components’ `CspStyleSheetManager`). `popover.tsx` switched from
  `setAttribute('style', …)` to a CSSOM `.style.overflow` write (also drops brittle string handling).
- **`fix: clone lottie animationData`** — `structuredClone` per instance; `lottie-web` mutates the
  animationData it is given, and several components share one imported JSON, so without a clone the
  animation played only once (caught in manual QA).

## Final Chrome-production CSP

```
script-src 'self' 'wasm-unsafe-eval';
style-src 'self' 'nonce-<per-build value>';
style-src-attr 'unsafe-inline'
```

`<style>`/`<link>` **elements** are nonce-locked (via `style-src` — `style-src-elem` is unset and falls
back to it), so an injected stylesheet without the build nonce is blocked. Inline style **attributes**
and `element.style` writes — which React (`style={{…}}`), `lottie-web`, `react-loading-skeleton` and
`react-tiny-popover` all perform at runtime and which a nonce cannot cover — are permitted via
`style-src-attr 'unsafe-inline'`. This is a deliberate, partial hardening: a net improvement over the
old blanket `style-src 'unsafe-inline'` (which also allowed injected `<style>` elements).

**Tradeoff:** the nonce is generated per build and shipped in the packaged manifest, so it is effectively
public/constant for a release and does not add per-response unpredictability. Its value is narrow but
legitimate — it lets `style-src` drop `'unsafe-inline'` for stylesheet *elements* while the extension’s
own bundled/styled-components sheets pass.

## Cross-browser parity

- **Chrome MV3 prod:** the nonce + `style-src-attr` shape above.
- **Chrome MV3 dev / Firefox MV2:** `style-src 'unsafe-inline'` (unchanged).
- **Safari MV2:** CSP untouched (`getCSP()` returns `undefined`, as before).
- No `permissions` / `host_permissions` / `web_accessible_resources` / DNR / manifest-file changes.

## Verification

- `ci-check` green (39 suites / 202 tests, 0 lint errors); Chrome + Firefox builds succeed.
- **Manual QA (Chrome production build):** zero CSP style violations across all 5 apps (popup,
  onboarding, connect-to-app, signature-request, import-account-with-file); lottie animations loop
  correctly across multiple instances.

## Reviews

Whole-branch review + `extension-manifest-reviewer`: clean / ready. Non-blocking minors noted
(plan-mandated defensive null-ref guard in the wrapper; a dormant string-`path` branch kept for
drop-in parity; the static-nonce framing above).

## Owner gates before un-drafting

- [ ] Safari build + in-browser load check (macOS/Xcode).
- [ ] Optional: full sign/QA pass.

> Note: a pre-existing post-onboarding "Password doesn't exist" race (cleared by refresh) was observed
> during QA; it is unrelated to this PR (no vault/keys/messaging changes here) and tracked separately.


Comp0te:
**Suggestion (non-blocking):** consider importing `lottie-web/build/player/lottie_light` instead of the default `lottie-web` entry in `src/libs/ui/components/lottie-player/lottie-player.tsx`.

The wrapper hardcodes `renderer: 'svg'`, which is exactly what the light build provides. The full build additionally ships the canvas/html renderers and the expressions engine — `eval`/`Function`-based code that could never execute under our `script-src 'self'` CSP anyway (which also proves the bundled spinner/dots animations don't use expressions). Switching to the light build:

- removes eval-adjacent dead code from a wallet bundle (nice for store review too),
- noticeably cuts bundle size (the expressions engine + extra renderers are a large chunk of lottie-web),
- requires no other changes: `import lottie from 'lottie-web/build/player/lottie_light';` is a drop-in for the `loadAnimation` usage here.

Only caveat: light build supports SVG rendering only and no expressions — both already true for our usage.

ost-ptk:
Applied in 87a500c3 — switched `src/libs/ui/components/lottie-player/lottie-player.tsx` to `import lottie from 'lottie-web/build/player/lottie_light'`.

Confirmed:
- **Drop-in, fully typed** — `lottie-web/build/player/lottie_light.d.ts` re-exports the main type declarations, so strict `tsc` stays clean with no extra shims.
- **Bundle shrink** (`build:chrome`, uncompressed): `popup.bundle.js` −139 KB (2,915,148 → 2,775,657), `onboarding.bundle.js` −138 KB (2,517,485 → 2,379,308), with the same reduction in `connectToApp`/`signatureRequest`; `build/chrome` 26M → 25M.
- **CSP unchanged** and `ci-check` green.

Good catch on removing the eval/`Function`-based expressions engine from the wallet bundle — nicely aligned with this PR's `script-src 'self'` hardening and store review. Thanks!

Comp0te (approved):

PR #1401, opened 2026-07-09
perf: DEP-99 — debounce vault re-encryption (WALLET-1343)

ost-ptk:
## DEP-99 P1.3 — debounce vault re-encryption

Part of the DEP-99 security & architecture series. Stacked on `release/2.6.0` (PR3, `01343e0e`).

### Problem

`updateVaultCipher` re-encrypted the **entire** vault into `vaultCipher` on every
vault-mutating action via `takeLatest` over 15 triggers. A burst of edits (multi-field
account changes, rapid connect/disconnect) fired one full AES-256-GCM re-encryption each,
redundantly re-encrypting-and-persisting the same growing plaintext several times in a row.

### Change

- **Debounce** the re-encryption: `takeLatest([15 triggers], updateVaultCipher)` →
  `debounce(500, [15 triggers], updateVaultCipher)`. A burst now coalesces into a single
  trailing-edge re-encryption. The 15-trigger list is unchanged.
- **Flush on lock:** `lockVaultSaga` now runs a synchronous `updateVaultCipher()` as its
  first step, before `sessionReseted()` / `vaultReseted()`. Without this, an edit made in
  the last 500 ms before a lock would never be persisted (its debounced encrypt would never
  fire before teardown). Reusing `updateVaultCipher` keeps the ciphertext path identical.
- Minor: the encrypt call is now `sagaCall(encryptVault, key, vault)` instead of
  `sagaCall(() => encryptVault(key, vault))` — exactly equivalent, purely so the call is
  matchable in tests.
- **Straggler guard (review follow-up):** `updateVaultCipher` early-returns when the session
  key is null. After the P1.2 saga-error channel (#1399, merged into this branch), a debounced
  straggler firing after a lock would otherwise throw and surface a spurious
  "Encryption key doesn't exist" banner on routine edit-then-lock. It is now a silent no-op and
  still cannot overwrite the flushed cipher.

### Correctness / safety

- **Ciphertext format is unchanged** — `base64(iv[12] || ciphertext || tag[16])` AES-256-GCM.
  The crypto path (`libs/crypto/vault.ts`, `libs/crypto/aes.ts`) is untouched; existing
  vaults still decrypt. Confirmed by `crypto-security-reviewer` (CLEAN, byte-identity verified).
- **Post-lock straggler is harmless:** a debounced run that fires ~500 ms after a lock reads a
  null session key and `encryptVault` throws before touching the vault (caught internally), so
  it can never overwrite the freshly flushed cipher with a wiped/stale vault.
- Tests pin both behaviors: a burst of 3 edits produces exactly one re-encryption, and the
  lock flush dispatches `vaultCipherCreated` **before** `sessionReseted` (ordering asserted via
  the run's ordered `effects.put`, and verified to fail if the flush is moved after teardown).

### Benchmark (Task 6.2)

`encryptVault` wall-time (200-iteration average, warmed, `@noble/ciphers` AES-256-GCM):

| Vault                             | plaintext | per call    |
| --------------------------------- | --------- | ----------- |
| 50 accounts                       | 15.9 KB   | **0.21 ms** |
| 100 accounts                      | 29 KB     | 0.33 ms     |
| 50 accounts + 50×4 KB signed-JSON | 221 KB    | 2.14 ms     |

Per-call main-thread block is far below the ~50 ms threshold, so **no `chrome.offscreen`
worker is needed** and `minimum_chrome_version` is unaffected. (Benchmark was a throwaway; not
committed.)

### Known tradeoff

Debounce opens an up-to-500 ms window where the persisted `vaultCipher` lags the in-memory
vault. The SW keep-alive anchor spans only the encryption itself, not the debounce wait, so an
MV3 service-worker kill inside that window loses the last change until it is re-encrypted. The
trigger list includes `deployPayloadReceived` / `eip712PayloadReceived`, so this window covers
pending dapp signing-request payloads as well as account edits — an SW kill there drops the
persisted request and the dapp retries. No partial or corrupt cipher can result (the persisted
blob is always a complete encryption of a prior consistent vault). This is the durability/UX
tradeoff P1.3 intentionally accepts; the lock flush closes the same window for the lock path.

### Verification

Branch is merged forward with `release/2.6.0` (brings in #1399 P1.2 + #1400), so the diff below
shows only this PR's two files.

- `npm run ci-check` — green (45 suites / 259 tests, tsc clean).
- `e2e:chrome:headless:onboarding` — 6/6 passed.
- `e2e:chrome:headless:popup` — 45 passed, 1 skipped (baseline), 1 flaky retry on an unrelated
  signature-cancel dialog test.
- `crypto-security-reviewer` — CLEAN (byte-identity of the cipher path confirmed).

**Pending (per-PR gate):** chrome/firefox/safari builds + the manual browser QA checklist
(lock/unlock + rapid account edits round-trip).


Comp0te (commented):
Reviewed against the DEP-99 appendix (P1.3). The implementation is correct and I found no security issues — details below, all findings are minor / follow-up material.

**Verified:**
- Straggler-after-**lock** is harmless as claimed: `encryptVault(null, …)` throws at `libs/crypto/vault.ts:9` and `updateVaultCipher` catches internally, so it can't overwrite the flushed cipher.
- Straggler-after-**reset** (path not covered in the PR description) is also safe: `resetVaultSaga` resets the session key in the same synchronous put-block as the vault wipe, so a pending debounced encrypt fires against a null key and can't resurrect a cipher of the emptied vault.
- Crypto path untouched, fresh CSPRNG IV per call, blob format identical; flush ordering (cipher put before `sessionReseted`) is pinned by the test.
- Flush-on-lock is a genuine improvement over the raw P1.3 spec — plain `debounce(500)` alone would have introduced a lost-edit-on-lock bug.

**Follow-ups worth tracking:**
- The P1.3 spec also asks to move encryption into a Web Worker; the benchmark (0.21–2.14 ms) justifies dropping that, but the decision should be recorded as a status note in the appendix doc (same pattern as P0.4 / P1.5), so the item doesn't look forgotten.

**Nits (no action needed):** `flushProvides` is typed as select-selector tuples but includes a `call.fn` provider; `{} as never` for the vault stub would read better as `as VaultState`; the debounce test always runs the full 700 ms wall-clock since the root saga never completes (~0.7 s added to the suite).

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -124,6 +129,16 @@ export function* vaultSagas() {
...
   try {
+    // Flush any debounced-but-not-yet-persisted vault change before teardown.
+    // updateVaultCipher now runs on debounce(500ms); a vault edit in the last
+    // 500ms before lock would otherwise never be encrypted. Reusing
+    // updateVaultCipher keeps the ciphertext path (and blob format) identical.
+    // It reads the still-live encryptionKeyHash + vault; the resets below wipe
+    // both. A stale debounced straggler that fires ~500ms later runs against the
+    // reset session (encryptionKeyHash === null) — encryptVault throws, is caught
+    // inside updateVaultCipher, and can never overwrite this flushed cipher.
+    yield* updateVaultCipher();
```
Comp0te, line 145: The throw-and-catch design for the post-lock straggler is safe, but it logs `console.error("Encryption key doesn't exist")` on every lock that follows a vault edit within 500 ms — an error log for a perfectly legitimate state. Once P1.2 lands (saga `catch` → UI toast/Sentry), this becomes a spurious user-facing error on routine locks.

Suggestion: early-`return` in `updateVaultCipher` when `encryptionKeyHash` is null (locked isn't an error), or cancel the pending debounce task here instead of relying on the throw.

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -96,7 +100,8 @@ export function* vaultSagas() {
     ],
     timeoutCounterSaga
   );
-  yield takeLatest(
+  yield debounce(
```
Comp0te, line 108: Two notes on the `takeLatest` → `debounce` swap, neither blocking:

1. `takeLatest` cancelled an in-flight run when a new trigger arrived; `debounce` doesn't — a task already past its selects keeps running, so two `updateVaultCipher` instances can briefly overlap and the *earlier* (staler) cipher could in principle be persisted last. This requires a single encryption to outlast the 500 ms debounce window (~2 ms measured), so it's practically unreachable — but the invariant (encrypt ≪ debounce) is worth a one-line comment for whoever grows the vault later.

2. The trigger list includes `deployPayloadReceived` / `eip712PayloadReceived`, so the 500 ms durability window also covers pending dapp signing-request payloads, not just account edits — an SW kill in that window drops the persisted request and the dapp has to retry. Same accepted tradeoff, just a wider scope than the PR description's "vault edits" suggests.

src/background/redux/sagas/vault-sagas.test.ts
```diff
@@ -270,4 +293,47 @@ describe('lockVaultSaga', () => {
...
+      (e: { payload: { action: { type: string } } }) => e.payload.action.type
+    );
+    expect(putTypes.indexOf(vaultCipherCreated.type)).toBeGreaterThanOrEqual(0);
+    expect(putTypes.indexOf(vaultCipherCreated.type)).toBeLessThan(
+      putTypes.indexOf(sessionReseted.type)
+    );
+  });
+});
+
+describe('updateVaultCipher debounce', () => {
```
Comp0te, line 342: Three test suggestions:

1. The `encryptVault` spy is never restored and jest config has no `restoreMocks`, so the mock leaks into any test added after this one in the file. Add `afterEach(() => jest.restoreAllMocks())` (or restore in-test).

2. Nothing asserts that *spaced* edits (> 500 ms apart) each produce their own encryption — this test alone can't distinguish correct trailing-edge debounce from accidental over-coalescing (e.g. a leading-edge or infinite-coalesce misconfiguration).

3. The straggler-after-lock harmlessness is argued in the PR description but not pinned by a test; a `withReducer`-based run (edit → `lockVault` before the debounce elapses → run past 500 ms → assert the flushed cipher wasn't overwritten and `encryptVault` ran exactly once with a live key) would lock it in.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Done — implemented the early-`return` in `updateVaultCipher` when `encryptionKeyHash == null` (commit 4480b0f2). This became a real bug once #1399 (P1.2) merged in: the straggler's throw hit the new `sagaError` dispatch and would have surfaced a spurious "Encryption key doesn't exist" banner on any routine edit-then-lock. The straggler is now a silent no-op and still can't overwrite the lock flush. Added a `silent no-op when a debounced run lands after the session is locked` test (verified it fails if the guard is removed).

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Both addressed (commit 4480b0f2):

1. Added a comment at the `debounce` call noting the invariant — an encryption (~2 ms even on a 50-account vault) stays far below the 500 ms window, so the no-cancel overlap the reviewer describes is unreachable in practice; kept as a caution for whoever grows the vault/crypto.
2. Widened the PR description's durability-window note to call out that the trigger list includes `deployPayloadReceived`/`eip712PayloadReceived`, so the ≤500 ms window covers pending dapp signing-request payloads too, not just account edits.

src/background/redux/sagas/vault-sagas.test.ts
ost-ptk: Thanks — addressed 2 of 3:

1. The `encryptVault` spy is already restored: there's a file-level `afterEach(() => jest.restoreAllMocks())` (top of the file), and `jest.spyOn` mocks are covered by `restoreAllMocks`. No leak — this describe is also the last in the file. No change made.
2. Added the straggler-after-lock test (`silent no-op when a debounced run lands after the session is locked`) — asserts `encryptVault` isn't called and neither `vaultCipherCreated` nor `sagaError` is put when the debounced run lands on a null key. Verified it fails without the guard.
3. Skipped the spaced-edits (>500 ms → two encryptions) test intentionally: `expectSaga` can't express timed/spaced dispatches (all `.dispatch()`es are delivered up front), and `debounce(500, …)` is a fixed redux-saga primitive with no leading-edge/infinite-coalesce knob that our code sets — so such a test would exercise the library, not our configuration. The burst test (3 rapid → 1) plus the straggler test cover the behavior we own. Happy to add it if you'd prefer a wall-clock integration test.

Comp0te (approved):

PR #1402, opened 2026-07-10
DEP-99 — P1.6: test foundation, coverage gate & DX (WALLET-1343)

ost-ptk:
## DEP-99 / P1.6 + long-horizon — test foundation, coverage gate & DX

Final PR of the DEP-99 security-architecture backlog (**WALLET-1343**). Scope is testing +
developer experience; no user-facing runtime behavior changes.

Jira: https://make-software.atlassian.net/browse/WALLET-1343

### What's in it

- **Saga test foundation** — `expectSaga` tests for `unlockVaultSaga` (the vault-unlock flow was the
  one untested core saga; lockout-resume and the window/sign state machine were already covered by
  PR3's saga tests and PR2's `windowManagement` reducer tests respectively).
- **Per-handler tests + coverage universe** — real behavioral tests for every background message
  handler (`private-state`, `legacy-import`, `redux-actions`, `bringweb3`, `sdk-methods`), focused on
  the security gates (`isTrustedUiSender`, the legacy-import membership-oracle close, the fail-closed
  action-forwarding). `collectCoverageFrom` is widened to `background/handlers/**` and
  `redux/sagas/**` with realistic per-directory thresholds; reducers stay at 100%.
- **⚠️ `ci-check` now runs `test:coverage`** — previously `ci-check` ran `test` (no `--coverage`), so
  **no coverage threshold — including the existing reducer-100% gate — was actually enforced in CI**.
  It is now. Local `ci-check` is slightly slower and the coverage gate is real.
- **Storage-keys architecture doc** — `docs/architecture/storage-keys.md` documents why the
  obfuscated `storage.local` key names are immutable/append-only, that opacity ≠ encryption, and the
  future direction; plus a pointer comment in `get-main-store.ts`.
- **Bundle analyzer** — `webpack-bundle-analyzer` behind `ANALYZE=true` (`npm run build:analyze`),
  static report, zero footprint in normal/prod builds. Dev extension auto-reload was evaluated and
  deferred (it would re-enable the background/content-script reload the existing
  `hot:false`/`liveReload:false` setting deliberately excludes) — decision recorded in
  `utils/webserver.js`.
- **Dead-code sweep** — removed the dead private-material selectors left by DEP-16 P0.1
  (`selectPasswordHash`/`selectPasswordSaltHash`/`selectKeyDerivationSaltHash`/`selectVaultCipher`),
  the unused `popupStateUpdated` type and `RootStateKey`, and collapsed the redundant `RootAction`
  alias.
- **`FORWARDED_ACTION_TYPES` parity test** — ties the hand-maintained forwarding set to the actual
  RTK action creators, so adding a UI action and forgetting to forward it now fails at test time.
- **`keysDoesExist` derivation dedupe** — single source (`withDerivedFlag`) shared by the reducer and
  the storage preload; the flag is still recomputed from the hashes on load (a poisoned persisted
  flag cannot survive a service-worker restart).

### Verification

- `npm run ci-check` green (51 suites / 340 tests, coverage gate satisfied).
- Chrome + Firefox production builds green; analyzer has zero footprint in normal builds.
- Onboarding e2e 6/6.

### Stacking

Stacked on PR6 (#1401, P1.3 debounce) — opened with base `WALLET-1343-dep-99-p13-debounce`.
Retarget to `release/2.6.0` once #1401 merges. The branch is merge-forwarded to the current PR6
tip (which itself carries release/2.6.0 → PR4/PR7 + PR6's post-lock straggler fix); the parity test
picked up PR4's `sagaError` as a background-only creator during that merge and its `EXCLUSIONS` list
was updated accordingly.

### Follow-ups (non-blocking)

- The parity test derives its creator universe from the 17 action modules `redux-actions.ts` imports.
  Two more `redux/*/actions.ts` modules exist (`active-origin`, `active-origin-favicon`) — both
  currently background-only (sole dispatch at `background/index.ts`), so there is no forwarding gap
  today, but they are invisible to the parity check. Deriving the universe from an `fs` glob of
  `src/background/redux/*/actions.ts` so a future module can't silently opt out is tracked as
  **WALLET-1349**.

### Owner gates (not done here)

- Safari build/load (Xcode).
- Full e2e via CI (popup + Firefox smoke).
- Quick manual smoke — no runtime UI changes expected.


docs/architecture/storage-keys.md
```diff
@@ -0,0 +1,98 @@
...
+subscription tick (see `get-main-store.ts`).
+
+| Constant                       | Obfuscated key   | Stores                                                                                           | Secret?                                      | Plaintext at rest?                                                     |
+| ------------------------------ | ---------------- | ------------------------------------------------------------------------------------------------ | -------------------------------------------- | ---------------------------------------------------------------------- |
+| `VAULT_CIPHER_KEY`             | `zazXu8w9GyCtxZ` | The encrypted vault blob (`vaultCipher` slice) — accounts, keys, and other secret vault contents | **Yes** — the only secret material persisted | No — AES-256-GCM ciphertext, `base64(iv[12]\|\|ciphertext\|\|tag[16])` |
+| `KEYS_KEY`                     | `2yNVAEQJB5rxMg` | Password/salt hashes and key-derivation salt hash (`keys` slice)                                 | No (hashes, not the password or key itself)  | Yes                                                                    |
```
Comp0te: `KEYS_KEY` is marked **Secret? No**, but the codebase treats exactly these fields as secrets: `private-state.ts` calls `passwordHash`/`passwordSaltHash`/`keyDerivationSaltHash` "at-rest secrets" and serves them on demand behind the trusted-sender gate precisely so they don't reach untrusted contexts (P0.1). Under this doc's own threat model (anyone who can read `storage.local`), a plaintext password hash + salt is offline password-guessing material for the wallet password.

Suggest rewording to something like: *"Sensitive — offline password-guessing material; gated as private state at runtime"* so the doc doesn't contradict the P0.1 posture.

src/background/handlers/bringweb3.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  typeof emitSdkEventToActiveTabs
+>;
+
+const store = { getState: () => ({}) } as unknown as MainStore;
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('handleBringWeb3 — getActivePublicKey', () => {
```
Comp0te, line 68: This pins the open **P0.6** behavior (`getActivePublicKey` discloses the account address with no lock check and no connection check) as the expected result. Characterization tests of current behavior are fine, but please add a comment referencing P0.6 so the assertion isn't read as an endorsement — and so whoever closes P0.6 knows this test must flip rather than be "fixed" back.

Same applies to the ungated `promptLoginRequest` tests below (the P0.6 forced-popup primitive).

Comp0te (commented):
Reviewed against the DEP-99 security-architecture backlog (P1.6 + long-horizon items). Solid overall — the `ci-check` → `test:coverage` change makes the previously-unenforced coverage gate real, the handler tests exercise the actual `isTrustedUiSender` gate (not a mock of it), and the dead private-material selectors are confirmed consumer-free. Two asks inline, plus non-blocking notes:

- **Parity-test universe gap** — `active-origin` / `active-origin-favicon` are invisible to the check. Verified both are background-only today (sole dispatch at `background/index.ts`), so no live gap, but worth a Jira ticket for the fs-glob derivation rather than leaving it only in the PR body.
- **`bringweb3.ts` locked path is fire-and-forget** — `windows.getCurrent().then(...)` with no await/catch (pre-existing). It's what forces the test to pump microtasks manually (`await Promise.resolve()` ×2); awaiting it in the handler would fix both the unhandled-rejection smell and the test brittleness.
- **jest.config comment nit** — Jest checks a file against every matching threshold group and removes it from `global`, rather than picking the "LONGEST path key". The effect here is identical (reducers don't live under `handlers/` or `sagas/`), just the comment.
- **Coverage floors** sit at/just below achieved (e.g. sagas functions 35 vs ~37) — reasonable to keep the gate honest, but worth ratcheting periodically so headroom doesn't become silent regression room.

docs/architecture/storage-keys.md
ost-ptk: Good catch — fixed in cf0afa91. `KEYS_KEY` is now marked **Sensitive** ("the password hash + salts are offline password-guessing material; treated as private state (P0.1), served only behind the trusted-sender gate at runtime"), and I narrowed `VAULT_CIPHER_KEY` to "the only slice encrypted at rest" so the two rows no longer contradict the P0.1 posture.

src/background/handlers/bringweb3.test.ts
ost-ptk: Done in cf0afa91 — added a NOTE above both the `getActivePublicKey` and `promptLoginRequest` suites making clear they characterize the **current, known-vulnerable P0.6 behavior** (address disclosure / forced-popup with no lock or connection check) and that the assertions must **flip** when P0.6 is closed, not be restored.

ost-ptk:
Thanks for the review — addressed in cf0afa91:

- **storage-keys `KEYS_KEY` framing** → reworded to *Sensitive* (offline password-guessing material, private state behind the trusted-sender gate per P0.1); `VAULT_CIPHER_KEY` narrowed to "only slice encrypted at rest" so the rows don't contradict.
- **bringweb3 P0.6 characterization** → added NOTEs above the `getActivePublicKey` + `promptLoginRequest` suites (must flip when P0.6 is closed).
- **bringweb3.ts fire-and-forget** → now `await`s `windows.getCurrent()`/`create()` in the locked branch; rejection propagates to the router and the test no longer pumps microtasks.
- **jest.config comment nit** → corrected (per-group check + drop-from-global, not a "longest path key").
- **Parity-universe fs-glob** → tracked as **WALLET-1349** (non-blocking; verified no live gap today).
- **Coverage-floor ratcheting** → noted; will bump periodically to keep headroom honest.

ci-check green; e2e (chrome onboarding 6/6, firefox smoke 1/1, popup unaffected) unchanged.

Comp0te (approved):

Issue #730, opened 2023-07-18
[BUG] Sign promise does not fullfill when the user closes the window


davidatwhiletrue:
**Describe the bug**
`sign()` method returns a promise that fulfills when the user signs or cancels the transaction signature.
But if the user closes the casper wallet window, the promise that not fulfill.

**Expected behavior**
The expect result would be that the promise fulfills with `canceled` property set to `true`. Same as when the user clicks on `Cancel` button.

**Required information:**
 - Operating System: Windows 11 & Mac OS
 - Browser Name and Version: Chrome and Edge
 - Casper Wallet Version

Comp0te:
Fixed in [v2.7.0](https://github.com/make-software/casper-wallet/releases/tag/v2.7.0) by #1427.

Approval requests are now tied to the windows displaying them, and a single `windows.onRemoved` listener in the background cancels any request whose last displaying window went away. Closing the signing window with 'X' now delivers `signResponse({ cancelled: true })` to the dapp — the same payload as pressing Cancel — instead of leaving the promise pending until its own timeout.

The same applies to `signMessage`, `signTypedData` and `decryptMessage`. A short grace period before the cancel lets a genuine in-flight response (including a Ledger confirmation) land first, so a real signature is never destroyed.

Issue #560, opened 2023-03-02
`requestConnection` and `requestSwitchAccount` promises do not fulfill it the window is closed with 'X' button


davidatwhiletrue:
`requestConnection` and `requestSwitchAccount` promises fulfill when the user completes the requested action or the user presses cancel in the window.
But the promises do not fulfill if the user closes the window with the 'X' button (at least on Windows).

Comp0te:
Fixed in [v2.7.0](https://github.com/make-software/casper-wallet/releases/tag/v2.7.0) by #1427.

Approval requests are now tied to the windows displaying them, and a single `windows.onRemoved` listener in the background cancels any request whose last displaying window went away. Closing the window with 'X' now delivers `connectResponse(false)` / `switchAccountResponse(false)` to the dapp — the same payload as pressing Cancel — so both promises fulfill instead of hanging.

All six approval methods are covered (`connect`, `switchAccount`, `sign`, `signMessage`, `signTypedData`, `decryptMessage`). See also #730.

PR #1405, opened 2026-07-14
fix: WALLET-1344 — cancel pending dapp SDK requests when the approval window is closed via "X"

ost-ptk:
## Summary

Dapp SDK promises (`requestConnection`, `requestSwitchAccount`, `sign`, `signMessage`, `signMessageEIP712`, `decryptMessage`) previously **hung forever** when the user dismissed the wallet approval popup with the OS **"X"** button instead of Confirm/Cancel. This PR makes the background `windows.onRemoved` listener the authoritative close-canceller, so closing via "X" now resolves the dapp promise with the method-correct Cancel shape.

- Fixes make-software/casper-wallet#560 (connect / switch account)
- Fixes make-software/casper-wallet#730 (sign)
- Jira: https://make-software.atlassian.net/browse/WALLET-1344

## Root cause (re-verified after the DEP-99 refactor)

DEP-99 (WALLET-1343) built the request registry and server-side response dedup, but it never sent a cancellation on window close — `windowClosed` was a pure reducer. Actual cancellation still depended on the per-page `beforeunload → handleCancel` handlers, which Chrome frequently does **not** fire when an extension popup is torn down via "X" (#730), while the connect-to-app pages had **no** close handler at all (#560).

## What changed

1. **Registry carries a descriptor.** A new parallel `pendingRequests: Record<requestId, { tabId, origin, method }>` map on the `windowManagement` slice. The existing `requests` status map and `selectRequestStatus` are kept byte-unchanged, so the server-side dedup path and its tests are untouched. Populated at the 6 SDK-method dispatch sites.
2. **Background cancel-on-close handler** (`cancelOpenRequestsForClosedWindow`). On the tracked window's removal it waits a bounded grace (`CANCEL_GRACE_MS` = 250 ms), then cancels only the requests still open (`initiallyOpen ∩ stillOpen`), sending each its method-correct Cancel to the dapp tab via `tabs.sendMessage` (with same-origin delivery fallback). It dispatches `windowIdCleared` (not `windowClosed`) so it can never clobber a request opened during the grace.
3. **The grace protects genuine in-flight responses.** A Confirm/Cancel button response — or a **Ledger success cleanup that programmatically closes the approval window** — lands and marks the request `'responded'` within the grace, so the post-grace re-read skips it and a real signature is never replaced by a close-cancel. This is why no `use-ledger.ts` change was needed.
4. **Removed the unreliable `beforeunload` cancel handlers** from the 4 signature-request pages (`onRemoved` + grace is now the single close-cancel path for all approval pages).
5. **Side-fixes.** Two `handleCancel`s emitted a payload-correct but wrong-event-type response (`connectResponse` on a switch-account request; `signResponse` on a sign-message request) — corrected to the method's own builder. Harmless today (the SDK correlates by `requestId`), fixed for contract consistency.

6. **`pendingRequests` is kept out of the broadcast popup state.** `selectPopupState` pushed the whole `windowManagement` slice to every UI replica, which would have put each in-flight request's dapp origin and tabId into every open extension window (and the approval window is a single reused slot, so two origins can be open at once). The broadcast is narrowed to `{ windowId, requests }`; `createMainStoreReplica` restores the reducer's expected shape with an empty map. Found by the pre-merge security review.

## Testing

- `npm run ci-check` green — 358 tests. Coverage: `windowManagement/reducer.ts` 100%; `handlers/` directory 95.78 / 85.99 / 100 / 95.75 (floor 85 / 100 / 95 / 95); new `cancel-open-requests-on-close.ts` and `deliver-via-origin.ts` at 100%.
- `npm run build:chrome` and `npm run build:firefox` both pass.
- Unit tests exercise the grace re-check, the _answered-during-grace_ and _new-request-during-grace_ cases, and both delivery-fallback branches, using fake timers.

## Manual verification (cannot be covered by CI)

- [x] **Ledger on-device regression** — verified on hardware: a real Ledger `sign` / `signMessage` whose completion programmatically closes the approval window still returns the signature, not a cancel. The 250 ms grace holds in practice.
- [x] **Interactive e2e on Chrome**, against the playground dapp. Instrumented the background to log every SDK request and every cancel-on-close, and confirmed against the logs:
  - one `sign` → "X" → exactly one request cancelled, one Cancel delivered;
  - `sign` → Confirm/Cancel button → the request is already `'responded'` when `onRemoved` fires, so the close path correctly cancels nothing (no double-response);
  - two overlapping requests → "X" → both are open and both are cancelled — two requests, two promises, two resolutions. Before this PR the first of the two hung forever.
- [x] **Firefox (MV2)** — verified, close-via-"X" cancels correctly there too.

## Pre-merge review

Two independent reviews were run over the whole branch — a general engineering review and a security review. Both findings they raised are addressed or triaged:

- **Security (Medium): `pendingRequests` in the broadcast popup state** — fixed, see item 6 above.
- **General (Important): window reuse abandons an unanswered request** — real, but pre-existing and not a regression of this branch. See the discussion comment on this PR; needs a decision.

Verified clean by the security review: the cancel path cannot be induced to target an attacker-chosen tab (its `tabId`/`origin` come only from browser-supplied `sender` fields, and the path is triggered solely by the browser's own `windows.onRemoved`); `requestId` is 128-bit and SDK-internal, so cross-origin request confusion is not reachable; both dedupe paths mark `'responded'` synchronously before their own `await`, so one approval cannot yield two deliveries; cancel payloads carry no secret material for any of the 6 method shapes.

## Out of scope (documented follow-up)

- **MV3 durability.** The registry is in-memory (not persisted). A request pending while the vault is **locked** and the service worker is suspended mid-grace would find an empty registry on a cold `onRemoved` wake and cancel nothing. The warm/unlocked case (all signing, unlocked connect/switch) is covered by the existing keep-alive alarm. Persisting the registry is a separate ticket.

### DIFF
--- a/src/apps/connect-to-app/pages/switch-account/index.tsx
+++ b/src/apps/connect-to-app/pages/switch-account/index.tsx
@@ -31,7 +31,7 @@ export function SwitchAccountPage() {
 
   const handleCancel = async () => {
     await sendSdkResponseToSpecificTab(
-      sdkMethod.connectResponse(false, { requestId }),
+      sdkMethod.switchAccountResponse(false, { requestId }),
       requestTabId
     );
     closeCurrentWindow();
--- a/src/apps/signature-request/pages/decrypt-message/index.tsx
+++ b/src/apps/signature-request/pages/decrypt-message/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { shallowEqual, useSelector } from 'react-redux';
 
@@ -121,12 +121,6 @@ export function DecryptMessagePage() {
     );
     closeCurrentWindow();
   }, [requestId, requestTabId]);
-
-  useEffect(() => {
-    window.addEventListener('beforeunload', handleCancel);
-
-    return () => window.removeEventListener('beforeunload', handleCancel);
-  }, [handleCancel]);
 
   const handleDecrypt = useCallback(async () => {
     try {
--- a/src/apps/signature-request/pages/sign-eip712/index.tsx
+++ b/src/apps/signature-request/pages/sign-eip712/index.tsx
@@ -3,7 +3,7 @@ import {
   IEIP712SignTypedDataOptions,
   IEIP712TypedData
 } from 'casper-wallet-core';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { shallowEqual, useSelector } from 'react-redux';
 
@@ -248,12 +248,6 @@ export function SignEip712Page() {
     requestId,
     requestTabId
   ]);
-
-  useEffect(() => {
-    window.addEventListener('beforeunload', handleCancel);
-
-    return () => window.removeEventListener('beforeunload', handleCancel);
-  }, [handleCancel]);
 
   const renderFooter = () => {
     if (shouldTryToConnectAccount) {
--- a/src/apps/signature-request/pages/sign-message/index.tsx
+++ b/src/apps/signature-request/pages/sign-message/index.tsx
@@ -253,7 +253,7 @@ export function SignMessagePage() {
 
   const handleCancel = useCallback(() => {
     sendSdkResponseToSpecificTab(
-      sdkMethod.signResponse({ cancelled: true }, { requestId }),
+      sdkMethod.signMessageResponse({ cancelled: true }, { requestId }),
       requestTabId
     );
     closeCurrentWindow();
--- a/src/apps/signature-request/pages/sign-message/index.tsx
+++ b/src/apps/signature-request/pages/sign-message/index.tsx
@@ -259,12 +259,6 @@ export function SignMessagePage() {
     closeCurrentWindow();
   }, [requestId, requestTabId]);
 
-  useEffect(() => {
-    window.addEventListener('beforeunload', handleCancel);
-
-    return () => window.removeEventListener('beforeunload', handleCancel);
-  }, [handleCancel]);
-
   return (
     <LayoutWindow
       renderHeader={() => (
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -279,12 +279,6 @@ export function SignTransactionPage() {
     closeCurrentWindow();
   }, [requestId, requestTabId]);
 
-  useEffect(() => {
-    window.addEventListener('beforeunload', handleCancel);
-
-    return () => window.removeEventListener('beforeunload', handleCancel);
-  }, [handleCancel]);
-
   const {
     ledgerEventStatusToRender,
     makeSubmitLedgerAction,
--- a/src/background/handlers/cancel-open-requests-on-close.test.ts
+++ b/src/background/handlers/cancel-open-requests-on-close.test.ts
@@ -1,0 +1,213 @@
+import { tabs } from 'webextension-polyfill';
+
+import { emitSdkEventToActiveTabsWithOrigin } from '@background/utils';
+
+import { sdkMethod } from '@content/sdk-method';
+
+import {
+  CANCEL_GRACE_MS,
+  buildCancelResponse,
+  cancelOpenRequestsForClosedWindow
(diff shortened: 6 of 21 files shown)


ost-ptk:
## Window reuse abandons an unanswered request — found while reviewing this PR, needs a decision

@Comp0te — this came out of the pre-merge review of this branch. It is **not** a regression introduced here, but it is adjacent enough that it should be a conscious call rather than something we discover later.

### What happens

The approval window is a single reused slot. All six SDK handlers call `openWindow(...)` without `isNewWindow`, so `createOpenWindow` takes the `reuseExistingWindow` path (`src/background/create-open-window.ts:83-109`), which does:

```ts
await tabs.update({ url: getUrlByWindowApp(windowApp, searchParams) });
```

That is a full navigation of the same tab to the _new_ request's page. The previous request's UI is unmounted, but **nothing dispatches anything into `windowManagement`** — no `closeCurrentWindow()`, no response to the dapp. The first request just stays `'open'`, unanswered and now invisible to the user.

### Verified on a dev build

Instrumented the background to log every SDK request and every cancel-on-close:

```
[SDK-REQ] Sign         8ab262c1-…
[SDK-REQ] SignMessage  2b84a21b-…      ← reuses the window, Sign's page is gone
[CANCEL]  window … removed; open requests =
          ['sign:8ab262c1-…', 'signMessage:2b84a21b-…']   ← both still open
```

Same with two `sign` calls in a row. The dapp gets two Cancel responses when the window is finally closed — which is arithmetically correct (two requests, two promises), but the first one was never actually shown to the user for approval.

### Why it is not a regression of this PR

Before this branch the abandoned request hung **forever**. The four signature pages had a `beforeunload` handler that fired on this navigation and cancelled — but the connect-to-app pages never had one at all, so `connect` / `switchAccount` already leaked this way (that is #560). Now the abandoned request always resolves, just later than ideal: it is cancelled when the window is eventually closed, because `selectOpenRequests` is global rather than scoped to a `windowId`.

So the change is strictly an improvement. What is left is **latency and UX**, not a lost promise.

### The trap in the obvious fix

"Cancel the open requests before `tabs.update`" does not work as written: in `sdk-methods.ts` the new request's `windowRequestOpened` is dispatched **before** `openWindow` is called, so at reuse time the new request is already `'open'` and a blanket cancel would kill it too. Any fix needs an explicit `exceptRequestId` (or a reordering of the dispatch), across all six call sites.

### Options

|       | Approach                                                                                                  | Cost / risk                                                                                                                                                            |
| ----- | --------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **A** | Cancel abandoned requests immediately at reuse time, with `exceptRequestId` threaded through `openWindow` | Deterministic and in line with this branch's "background is authoritative" model. Touches `create-open-window.ts` + all 6 `sdk-methods.ts` sites, needs its own tests. |
| **B** | Leave as is, new ticket                                                                                   | Behaviour is already correct-but-delayed. Zero risk to this release.                                                                                                   |
| **C** | Queue the second request instead of silently replacing the first                                          | Best UX (the user actually gets to approve both), biggest change — needs design.                                                                                       |

My recommendation is **B for this PR + a follow-up ticket**, since A changes a signature used by all six SDK entry points and deserves its own review rather than riding along here. But if you would rather have A in this branch, it is a contained change and I am happy to do it.

**@Comp0te — leave as is, fix here, or new ticket?**

---

Two smaller follow-ups surfaced by the same review, listed for completeness (no action needed in this PR):

- `requests` and `pendingRequests` are never pruned — terminal `'responded'` entries live for the service worker's lifetime. Pre-existing for `requests`; this branch adds a second map with the same lifecycle.
- `windowClosed` is now dispatch-dead (the close path uses `windowIdCleared`). Kept deliberately for the reducer/parity guard; worth removing once this design has proven itself in production.

Comp0te:
@ost-ptk B — leave as is, new ticket. This PR is release-bound and A changes the `openWindow` contract for all six entry points; that deserves its own review.

One correction to the framing though: for the four signature pages the old `beforeunload` _did_ fire reliably on the reuse navigation (it's an in-page navigation, not window teardown), so those requests used to be cancelled immediately at reuse time — after this PR they're cancelled only at window close. Nothing is lost, but it's a latency regression for the sign→sign case, so the follow-up is genuinely needed, not nice-to-have.

Please make the ticket cover: A (with `exceptRequestId`) as the near-term fix, C as the long-term direction, plus the two housekeeping items from your comment (map pruning, removing `windowClosed` after a release cycle of soak).

Comp0te (approved):

Issue #56, opened 2022-04-14
How to work Safari build with Webpack dev server


vladimir-gachkovsky:
Need clarify how to works Safari build in details with Webpack Dev Server.

At the moment, `npm run start:safari` opens the application in Safari browser and we can open every file in the `build/firefox` folder, but changing those files does nothing. It looks like the Safari build has "copies" of these files. Purpose of this task is need to figure out exactly how the build works

PR #1406, opened 2026-07-15
WALLET-1348: CSPR.name expiration banners

ost-ptk:
## What

Warns users when an account's **cspr.name** on the active network expires within **14 days**, and routes them to https://cspr.name to renew.

- **Home banner** (`CsprNameExpirationBanner`) — appears when ≥1 name on the active network expires within 14 days and is not dismissed. "View names" opens the screen; "Dismiss" hides the banner (persisted per network).
- **Expiring-names screen** (`/cspr-name-expirations`) — lists all names expiring within 14 days, ascending, styled like the deploy-details rows (name left, date without time right). Footer opens https://cspr.name.
- **Persisted background slice** `csprNameExpirations` — per-network map `{ [network]: { [publicKey]: { csprName, expiresAt, dismissed } } }`, new immutable storage key.
- **Fetch** — popup react-query hook, once per popup session per network: `getAccountsInfo` → filter names → `resolveAccountFromCsprName` batched ≤5 concurrent → dispatch. Isolated failures (one failed lookup never sinks the batch).

## Behavior notes

- The 14-day window is a selector/dismiss-time filter; storage keeps all non-expired resolved names.
- Dismiss suppresses only the **banner**; the screen always lists all names in the window (it is the place to review/renew). A renewed name (new expiration date) or a newly-expiring name shows the banner again.
- All data (expirations + dismissals) is tracked per network.

## Dependency

Requires core change **make-software/casper-wallet-core#56** (adds `IAccountInfo.csprNameExpiresAt`, mapped from the resolution `expires_at`). This branch **temporarily pins** `casper-wallet-core` to that PR's branch commit; it must be **re-pinned to master** once #56 merges before this PR lands.

## Testing

`npm run ci-check` green (format + lint + tsc + 48 unit tests). New pure logic (reducer, selectors, concurrency helper, formatter) is unit-tested; React parts verified via tsc/lint/build and manual extension QA (banner show/hide/dismiss+persist, screen, network switching). `build:chrome` succeeds.

## Jira

https://make-software.atlassian.net/browse/WALLET-1348


src/apps/popup/pages/home/index.tsx
```diff
@@ -82,6 +96,7 @@ export function HomePageContent() {
       {activeMarketingEvent && (
         <AppEventBanner activeMarketingEvent={activeMarketingEvent} />
       )}
+      {showCsprNameExpirationBanner && <CsprNameExpirationBanner />}
```
Comp0te: **Critical: banner priority and session latch are missing (acceptance criteria 10 & 11).**

Both banners render as independent siblings, so when a marketing event and an expiring name coincide, two banners stack on Home (and marketing event #2 uses the same `cspr-name-banner.svg` background, so they'd be visually near-identical).

The spec requires: at most one banner at a time, expiration wins, and after dismissing the expiration banner the marketing banner stays suppressed until the next popup open. The plan (Task 6 Step 6) has the exact design — a `useRef` session latch:

```tsx
const wasExpirationBannerShown = useRef(false);
if (showCsprNameExpirationBanner) {
  wasExpirationBannerShown.current = true;
}
const showMarketingBanner =
  Boolean(activeMarketingEvent) &&
  !showCsprNameExpirationBanner &&
  !wasExpirationBannerShown.current;
```

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+        // whole batch (which would leave the banner hidden for every name
+        // this session). Treat a failed lookup as "unresolved" (null).
+        async csprName => {
+          try {
+            return await accountInfoRepository.resolveAccountFromCsprName(
+              csprName,
+              network,
+              false
+            );
+          } catch {
```
Comp0te: **Critical: a partial resolution failure silently wipes stored records and dismissed flags.**

This `catch { return null; }` makes a *failed* lookup (timeout, 429, 5xx — the core repo throws `AccountInfoError` for all of these) indistinguishable from a *legitimately absent* one (the repo already returns `null` itself for 404/expired). The nulls are filtered out, the dispatch still fires, and the reducer does a full replace of the network's map — so the failed names are deleted, including their `dismissed: true` flags.

Concrete scenario: 5 names, 2 dismissed; 3 resolutions hit a transient 429 (5 concurrent requests to one endpoint is exactly the shape that trips rate limits). The dispatch fires with 2 records, the other 3 are wiped. Next session they resolve fine and come back with `dismissed: false` — the banner the user explicitly dismissed resurfaces. Symmetrically, a name genuinely expiring in 3 days vanishes from the warning list for the whole session after one failed request.

Note the inversion: a *total* `getAccountsInfo` failure is safe (queryFn rejects, no dispatch, prior state survives), but a *partial* failure is the destructive path.

Suggested fix: keep failures distinguishable from absence — e.g. return a `FAILED` sentinel from the catch and either skip the dispatch when any resolution failed, or pass the failed names in the payload so the reducer *retains* the previous record (and its dismissed flag) instead of deleting it. Also log the error instead of swallowing it bare.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+    queryFn: async () => {
+      const accountHashes = publicKeys.map(getAccountHashFromPublicKey);
+
+      const accountsInfo = await accountInfoRepository.getAccountsInfo({
+        accountHashes,
+        network,
+        withProxyHeader: false
+      });
+
+      const csprNames = Object.values(accountsInfo)
```
Comp0te: **Critical: one unfunded account kills the feature for the whole session — silently.**

`getAccountsInfo` in casper-wallet-core builds its result as `remoteAccountsInfo[hash] ?? this._accountsInfoMapCache.get(hash)`, so when the API omits a hash (typical for fresh/unfunded accounts with no on-chain account info) and the cache has nothing, the record value is literally `undefined` despite the `Record<string, IAccountInfo>` return type.

`info.csprName` then throws `TypeError: Cannot read properties of undefined`, the whole queryFn rejects, and (since the query error isn't consumed anywhere) the banner never shows for *any* account — every session, deterministically, until the empty account is funded.

One-line fix:

```ts
const csprNames = Object.values(accountsInfo)
  .filter(info => info != null)
  .map(info => info.csprName)
  .filter((name): name is string => name != null);
```

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+import { runWithConcurrency } from './run-with-concurrency';
+
+const MAX_CONCURRENT_RESOLUTIONS = 5;
+
+export const useFetchCsprNameExpirations = (): void => {
+  const networkSetting = useSelector(selectActiveNetworkSetting);
+  const publicKeys = useSelector(selectVaultAccountsPublicKeys);
+  const network = networkSetting.toLowerCase() as CasperNetwork;
+
+  const { data } = useQuery({
```
Comp0te: **A total fetch failure is unobservable.**

Only `data` is destructured, and the shared query client defines no global error handler (TanStack Query v5 doesn't log query errors to the console by default). If `getAccountsInfo` throws at popup open, the query sits in error state and nothing anywhere observes it — the feature is silently dead for the session, and a report of "I never got the renewal warning" is undebuggable: no console error, no way to distinguish "API failed" from "no expiring names".

At minimum consume `error` and log it, following the existing `handleError` pattern (`src/libs/services/utils.ts`).

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+import { getAccountHashFromPublicKey } from '@libs/entities/Account';
+
+import { runWithConcurrency } from './run-with-concurrency';
+
+const MAX_CONCURRENT_RESOLUTIONS = 5;
+
+export const useFetchCsprNameExpirations = (): void => {
+  const networkSetting = useSelector(selectActiveNetworkSetting);
+  const publicKeys = useSelector(selectVaultAccountsPublicKeys);
+  const network = networkSetting.toLowerCase() as CasperNetwork;
```
Comp0te: **Untyped network keys can enter persisted state.**

`NetworkSetting` has four members (`Mainnet | Testnet | Devnet | Integration`), but this cast narrows to `CasperNetwork = 'mainnet' | 'testnet'`. On Devnet/Integration the fetch still runs and the persisted `CsprNameExpirationsState` gains a `'devnet'`/`'integration'` key that its type (`Partial<Record<CasperNetwork, …>>`) says cannot exist — and it flows into `storage.local`.

The cast is an inherited codebase pattern, but this is the only place it reaches persistence. A small guard, e.g. `toCasperNetwork(setting): CasperNetwork | undefined`, and skipping the feature (`enabled: false`) on unsupported networks would close it.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const { data } = useQuery({
+    queryKey: ['CSPR_NAME_EXPIRATIONS', network, publicKeys.join(',')],
+    enabled: publicKeys.length > 0,
+    staleTime: Infinity,
+    gcTime: Infinity,
+    // Fetch once per popup session — override the query client's 3-min
+    // refetchInterval default so this does not poll while Home is mounted.
+    refetchInterval: false,
+    queryFn: async () => {
```
Comp0te, line 33: **The queryFn — the riskiest logic in the PR — has zero tests.**

The fetch → filter → resolve-with-concurrency → build-records pipeline (error isolation, null/missing-`csprNameExpiresAt` skipping, the ≤5-concurrency contract against the repository) lives inline in the hook, where jest's `moduleNameMapper` can't even resolve the `@background/*` imports.

The plan (Task 5) had this as a pure `getCsprNameExpirations(publicKeys, network, repository)` function with an injected repository and three tests, including the concurrency assertion. Extracting it back out would make the partial-failure and undefined-entry issues testable, keep the hook a thin wrapper, and pin the once-per-session behavior (`refetchInterval: false`) that was already fixed once in this branch and currently has no regression test.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
```diff
@@ -0,0 +1,89 @@
...
+      );
+
+      const records: Record<string, CsprNameExpirationInput> = {};
+      resolved.forEach(info => {
+        if (
+          info != null &&
+          info.csprName != null &&
+          info.csprNameExpiresAt != null
+        ) {
+          records[info.publicKey] = {
```
Comp0te: **Records are keyed by the resolution response's `publicKey`, not the vault account's key.**

It's internally consistent (banner/dismiss/list all read the same map), but if the API returns the key in a different case/format than the vault, or a name resolves to a non-wallet account, the stored key diverges from the vault account key. The plan kept an accountHash → wallet publicKey map and keyed records by the vault's own key — that normalization would remove the edge case.

src/apps/popup/pages/expiring-cspr-names/content.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const CsprNameExpirationsContent: React.FC = () => {
+  const networkSetting = useSelector(selectActiveNetworkSetting);
+  const network = networkSetting.toLowerCase() as CasperNetwork;
+  const expiringNames = useSelector((state: RootState) =>
+    selectExpiringCsprNames(state, network)
+  );
+
+  return (
+    <ContentContainer>
```
Comp0te, line 38: The screen has no title — the plan's content had a "CSPR.name expirations" header (`<ParagraphContainer top={SpacingSize.XL}><Typography type="header">…`) above the Tile; here only the Tile renders.

src/background/redux/cspr-name-expirations/selectors.test.ts
```diff
@@ -0,0 +1,78 @@
...
+  ({ csprNameExpirations: { mainnet: mainnetMap } }) as any;
+
+describe('csprNameExpirations selectors', () => {
+  it('returns [] when the network has no map', () => {
+    expect(
+      selectExpiringCsprNames({ csprNameExpirations: {} } as any, 'mainnet')
+    ).toEqual([]);
+  });
+
+  it('includes only records within the next 14 days, sorted ascending', () => {
```
Comp0te: The window tests only use 3/10/30-day offsets — no case near the 14-day boundary (e.g. 13.9 vs 14.1 days), none at "expiring right now" (`diff ≈ 0`), and none for an invalid `expiresAt` string (NaN is silently excluded by the filter — safe direction, but unlogged and untested). These are one-liners and would catch an off-by-one edit (`<=` → `<`) or a constant change. Consider `jest.spyOn(Date, 'now')` to make the boundary cases deterministic instead of live `Date.now()`.

src/libs/services/account-info/run-with-concurrency.test.ts
```diff
@@ -0,0 +1,41 @@
+import { runWithConcurrency } from './run-with-concurrency';
+
+describe('runWithConcurrency', () => {
```
Comp0te: Worth one test pinning the rejection semantics: a worker rejection rejects the whole batch (`Promise.all` propagates). The only current caller neutralizes this with a per-item catch, but the helper is generic and exported — a future caller could easily assume per-item isolation, and a "helpful" refactor could silently change the behavior.

Minor plan deviation, fine as-is: the 14-day window and batch-size constants live next to their consumers (`FOURTEEN_DAYS_MS` in selectors.ts, `MAX_CONCURRENT_RESOLUTIONS` here) rather than in `src/constants.ts`.

src/apps/popup/pages/home/index.tsx
ost-ptk: Fixed in fda0c2bb. Home now renders at most one banner with the expiration banner taking priority, plus a `useRef` session latch: dismissing the expiration banner no longer reveals the marketing banner within the same popup session — it appears only on the next popup open when no expiration banner is active (AC10/AC11).

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Kept as-is by decision after discussing with the maintainer: this matches the official spec's design. `getCsprNameExpirations` uses per-item `.catch(() => null)` and the reducer full-replaces the active network's map, which is also the intended stale-record cleanup (AC12). Preserving `dismissed` across transient resolution failures would deviate from the approved spec, so it is deferred. The logic now lives in the extracted, unit-tested `getCsprNameExpirations`.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Fixed in fda0c2bb. The extracted `getCsprNameExpirations` filters with `Object.entries(accountsInfo).filter(([, info]) => Boolean(info?.csprName))`, so an `undefined` entry returned by `getAccountsInfo` for an unfunded account no longer throws and no longer rejects the whole query.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Acknowledged; not changed in this pass. The rewritten hook follows the spec shape (dispatch inside `queryFn`, no `error` consumer), so a total `getAccountsInfo` failure stays silent. Happy to add a `handleError`-style log as a follow-up if you want it in this PR.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Checked the installed `casper-wallet-core`: `type CasperNetwork = 'mainnet' | 'testnet' | 'devnet' | 'integration'` (4 members). So `'devnet'`/`'integration'` are valid keys of `Partial<Record<CasperNetwork, …>>`; the cast does not produce out-of-type keys and nothing untyped reaches `storage.local`. The two-member assumption comes from the spec's Global Constraints line, which is out of date. `.toLowerCase() as CasperNetwork` is also the established pattern across the other service hooks. Leaving as-is by decision. (The only real effect is that the fetch runs pointlessly on dev/integration where cspr.name is not served — harmless, returns empty.)

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Fixed in fda0c2bb. Extracted the pipeline into a pure `getCsprNameExpirations(publicKeys, network, repository)` with an injected repository and unit tests covering: resolve-only-accounts-with-a-name, skip-on-failure/no-expiry, and the ≤5 concurrency contract. `jest.config.js` `moduleNameMapper` was extended with `@libs/*` and `@background/*`; the hook is now a thin wrapper that dispatches the result.

src/libs/services/account-info/use-fetch-cspr-name-expirations.ts
ost-ptk: Fixed in fda0c2bb. `getCsprNameExpirations` builds an `accountHash → vault publicKey` map and keys the payload by the vault account's own public key (`publicKeyByHash[accountHash]`), not the resolution response's `publicKey`.

src/apps/popup/pages/expiring-cspr-names/content.tsx
ost-ptk: Fixed in fda0c2bb. The screen (now `src/apps/popup/pages/expiring-cspr-names/content.tsx`) renders a `CSPR.name expirations` header above the Tile.

src/background/redux/cspr-name-expirations/selectors.test.ts
ost-ptk: The time-window logic moved out of the selector into the pure `getExpiringCsprNames(map, now)`, tested deterministically in `expiring-cspr-names.test.ts` with a fixed `now` and cases at past / +3d / exactly +14d / +15d. The 13.9-vs-14.1 and NaN-`expiresAt` cases are not covered yet — can add them there if you want.

src/libs/services/account-info/run-with-concurrency.test.ts
ost-ptk: `run-with-concurrency` was removed in fda0c2bb; batching now uses `chunkArray` + `Promise.all` inside `getCsprNameExpirations`, whose test asserts the ≤5-concurrency behavior. The generic exported pool (and this rejection-semantics concern) no longer exists.

ost-ptk:
@Comp0te thanks for the follow-up commits — the failed-resolution/`failedPublicKeys` split, the `getCasperNetwork` map and the stored-record gating are all clear improvements, and I've pulled them in.

One question on `29ccffa4`:

```diff
-export const CSPR_NAME_EXPIRATION_NOTICE_DAYS = 14;
+export const CSPR_NAME_EXPIRATION_NOTICE_DAYS = 14 + 365;
```

Was this an intentional product change to the notice window, or a local QA aid that slipped into the branch? Asking because:

1. The spec (`2026-07-13-cspr-name-expiration-banner.md`) defines the window as 14 days, so at `14 + 365` the banner shows for essentially every name that has more than a year left — which reads as a behaviour change rather than a tweak.
2. It largely disables the caching you just added in `15d8fe66`. Rule 2 in `shouldResolve` ("more than the notice window until `expiresAt` → skip") is what avoids the network call; with a ~379-day window almost every record now falls *inside* the window, so non-dismissed names get re-resolved every session again.
3. The commit is the only one on the branch without a Conventional Commits prefix or the ticket key, which is partly why I suspect it wasn't meant to ship.

If it is a deliberate product decision I'm happy to keep it — I'd just write it as a plain number with a short comment on the rationale, and we should update the spec + acceptance criteria to match. Otherwise I'll revert the constant back to `14` (keeping the SVG `transform` change from the same commit).

Comp0te (approved):

Issue #609, opened 2023-03-30
[Safari ] Casper Wallet | Downloading account keys | Account key are not downloaded for Safari browser


karatnyk:
**Pre-conditions:**
A vault with an imported account exists

**Steps to reproduce:**
1. Navigate to 'Download account keys' screen
2. Provide a valid password and click on 'Download' button
3. Observer that success screen ('Your secret key was downloaded') is displayed
4. Open default 'Downloads' folder

**Expected result:** .pem key of an imported account is present in default 'Downloads' folder
**Actual result:** .pem key is not present in 'Downloads' folder

![vlc_gxlP9phrlj.gif](https://images.zenhubusercontent.com/6241d5bd38a85b3df1624280/d53e7790-27bf-4ab2-8293-a5235325171c)

adammake:
We need to create a ticket for a long term solution and then link it to this bug

Comp0te:
Fixed in [v2.7.0](https://github.com/make-software/casper-wallet/releases/tag/v2.7.0) by #1413.

Root cause: the export screen ran in the ephemeral extension popup. Triggering the download stole focus, Safari tore the popup document down, and the browser revoked the `blob:` URL mid-read — so the file never landed, while the success screen claimed it had.

The flow now opens in a dedicated window that outlives the download click (on all three browsers, to avoid a second behavioural branch). The final screen no longer asserts a write it cannot verify — `browser.downloads` does not exist in Safari Web Extensions — and offers a working 'Download again' instead. The old mitigation that merely hid the menu entry on Safari while leaving the route reachable is gone.

PR #1408, opened 2026-07-20
fix: WALLET-1345 — Safari key export writes the file, and the final screen stops claiming it did

ost-ptk:
Fixes the long-standing Safari bug where "Download account keys" showed a success screen but never wrote the file.

- Jira: [WALLET-1345](https://make-software.atlassian.net/browse/WALLET-1345)
- GitHub issue: #609 (open since 2023-04)

## Root cause

The export screen ran in the **ephemeral extension popup**. Triggering a download steals focus, Safari closes the popup, its document is torn down, and the browser revokes every `blob:` URL that document owned — aborting the in-flight read. The file never landed.

This was established by probing on a real Safari build, not inferred. In a persistent tab, four variants all downloaded correctly with correct filenames: a detached anchor with an immediate `revokeObjectURL` (a literal reproduction of the shipped code), an attached anchor with immediate revoke, an attached anchor with deferred revoke, and a `data:` URL. The same probe inside the popup produced no file and the popup closed. An earlier single-statement popup probe *did* produce a file, but named `Unknown` — a partially-won race.

So the anchor+blob mechanism, the detached anchor, the revoke timing, the `download` attribute and the MIME type are all exonerated. **`downloadFile` in `utils.ts` is deliberately unchanged** — changing it would have been a fix with no evidence behind it. Only the context was at fault.

The 2023 mitigation merely hid the menu entry on Safari (`hide: isSafariBuild`) while leaving the route registered, so the feature was never disabled — only made unreachable. That gate is now removed.

## What changed

**The flow opens in a dedicated window** (`windows.create({ type: 'popup' })`), verified on Safari to survive the download. Applied to **all three browsers** rather than gating Safari, to avoid a second behavioural branch — this is the main regression surface for review, since Chrome and Firefox worked before and their flow now moves off the popup.

**The final screen no longer claims a completed write.** Nothing can verify one: `downloadFile` is a fire-and-forget anchor click, and `browser.downloads` — the only API that reports success — [does not exist in Safari Web Extensions](https://developer.apple.com/forums/thread/660046). Rather than fake a signal, the screen states what to look for and offers a working **Download again**. That retry is only possible because the surface now outlives the click; in the popup the context was already gone by the time a user noticed the file was missing.

**Navigation corrected for a window.** `Back` is kept only where it genuinely steps back inside the flow. The two screens with nowhere to go back to now close the window instead of running a dead `navigate(-1)`, which in a single-history-entry window was a no-op that trapped the user. The header drops the wallet menu, network switcher and connection status — they otherwise let a user navigate the entire wallet inside a 376px export window.

**Failures are now visible to the user.** Previously every failure path in this flow was silent. See below — this turned out to be the larger half of the change.

### Failure handling

The project has **no Sentry or equivalent** — no error-tracking dependency of any kind. A `console.error` is therefore visible only to someone who opens devtools on the extension, which in practice means nobody. Production failures in this flow were invisible to users and to us simultaneously, so a failed key export was indistinguishable from a dead button. Every failure path now has a visible outcome.

**A failed export lands on a dedicated `Failure` step, not a banner.** This is deliberate: a banner on the Success screen would put "your keys were saved" next to a failure, which is the exact defect this ticket exists to remove. The step states that no file was written and that *nothing left the wallet and the accounts are unchanged* — the fear a secret-key export error actually raises — and offers **Try again**, which returns to account selection rather than retrying blind, so the user can confirm or change the selection first. `Download again` on the Success screen routes into the same step if it fails.

**`openExportKeysSurface` no longer swallows its rejection.** If the window cannot be created there is no export surface on which anything could be reported, so the menu reports it instead, and dismisses itself only once the window actually exists — otherwise the message would have nowhere to render.

**`closeExportKeysSurface` still only logs.** Deliberate: if closing fails, the window visibly stays on screen, so the failure is self-evident and the user's remaining action — closing it from the title bar — is the same one a message would suggest.

**Only the error name is logged, never the thrown value.** On this path the thrown value is built from key material, and it must not reach the console.

**The `Error` component is imported aliased** (`Error as ErrorMessage`). Imported unaliased it shadows the global `Error` constructor, which silently breaks any `new Error(...)` later in the same file — this bit during development.

### The shared password screen — note for reviewers

`PasswordProtectionPage` gates five flows: this one plus contact-details, backup-secret-phrase, wallet-qr-code and change-password. It is touched in two ways, and only one of them is scoped to key export.

**Scoped to this flow.** A new optional `onCloseWindow` prop swaps the header for a bare one whose only action closes the window. The four other callers do not pass it and therefore take the unchanged branch, byte-for-byte identical to before. This exists because a dedicated window has a single history entry, so the default `back` link was a dead `navigate(-1)` that trapped the user on the first screen of the export window.

**Affects all five flows.** The password field is now disabled while the password is being verified, matching the submit button, which was already disabled on the same condition. Previously, pressing Enter left the field editable and a user could keep typing into a form already on its way out. This is a pre-existing bug found while testing the export window; the fix is deliberately applied to the shared component rather than worked around locally, so it changes behaviour in the other four flows too. Worth a look during review even though it is outside the ticket's scope.

Neither change touches `onSubmit`, the verify-password worker call, or `setPasswordConfirmed` — the password gate itself is untouched.

## Testing

`npm run ci-check` passes (51 suites, 341 tests). e2e passes. Manually verified.

No unit tests were added: UI is outside `collectCoverageFrom` in `jest.config.js` and the repo has no component-testing layer. An e2e test would not have covered Safari — the only platform with the bug — nor could it assert that a file was written.

## Known and deliberately not addressed

**One-time window widening on Safari.** The first export shows Safari's native download-permission dialog, which is wider than the window, so Safari widens the window to fit it. Subsequent exports have no dialog and no resize, and the next window is created at the normal width. This is platform UI we do not control; forcing the size back would race the still-open dialog.

**Stale i18n catalogs.** `lang/` has not been regenerated since 2022 — roughly 400 source strings already fall back to English in every locale. The three new strings are no worse off. Regenerating rebuilds the entire catalog (~1750 changes) and belongs in its own ticket.

**`revokeObjectURL` race in `utils.ts`.** The synchronous revoke immediately after `click()` is latent fragility, explicitly refuted as the cause here. Worth a separate cleanup ticket.

[WALLET-1345]: https://make-software.atlassian.net/browse/WALLET-1345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/popup/pages/download-account-keys/failure.tsx
+++ b/src/apps/popup/pages/download-account-keys/failure.tsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { Trans, useTranslation } from 'react-i18next';
+
+import {
+  ContentContainer,
+  IllustrationContainer,
+  ParagraphContainer,
+  SpacingSize
+} from '@libs/layout';
+import { SvgIcon, Typography } from '@libs/ui/components';
+
+export const Failure = () => {
+  const { t } = useTranslation();
+
+  return (
+    <ContentContainer>
+      <IllustrationContainer>
+        <SvgIcon
+          src="assets/illustrations/error.svg"
+          width={190}
+          height={120}
+        />
+      </IllustrationContainer>
+
+      <ParagraphContainer top={SpacingSize.XL}>
+        <Typography type="header">
+          <Trans t={t}>Couldn’t download your keys</Trans>
+        </Typography>
+      </ParagraphContainer>
+
+      <ParagraphContainer top={SpacingSize.Medium}>
+        <Typography type="body" color="contentSecondary">
+          <Trans t={t}>
+            No file was saved. Nothing left your wallet and your accounts are
+            unchanged, so it is safe to try again.
+          </Trans>
+        </Typography>
+      </ParagraphContainer>
+    </ContentContainer>
+  );
+};
--- a/src/apps/popup/pages/download-account-keys/index.tsx
+++ b/src/apps/popup/pages/download-account-keys/index.tsx
@@ -47,25 +47,35 @@ export const DownloadAccountKeysPage = () => {
   }
 
   const downloadKeys = async () => {
-    const zip = new JSZip();
+    try {
+      const zip = new JSZip();
 
-    for (const account of selectedAccounts) {
-      const asymmetricKey = createAsymmetricKeys(
-        account.publicKey,
-        account.secretKey
-      );
+      for (const account of selectedAccounts) {
+        const asymmetricKey = createAsymmetricKeys(
+          account.publicKey,
+          account.secretKey
+        );
 
-      if (asymmetricKey.secretKey) {
-        const file = asymmetricKey.secretKey.toPem();
-        zip.file(`${account.name}_secret_key.pem`, file);
+        if (asymmetricKey.secretKey) {
+          const file = asymmetricKey.secretKey.toPem();
+          zip.file(`${account.name}_secret_key.pem`, file);
+        }
       }
-    }
 
-    await zip.generateAsync({ type: 'blob' }).then(function (content) {
+      const content = await zip.generateAsync({ type: 'blob' });
       downloadFile(new Blob([content]), 'casper-wallet-secret_keys.zip');
-    });
 
-    setDownloadAccountKeysStep(DownloadAccountKeysSteps.Success);
+      setDownloadAccountKeysStep(DownloadAccountKeysSteps.Success);
+    } catch (error) {
+      // Stay on this step: advancing to Success here is exactly the false
+      // "your keys were saved" this ticket exists to remove. Only the error
+      // name is logged — the thrown value is built from key material and must
+      // not reach the console.
+      console.error(
+        'downloadKeys: failed to build or download the keys archive',
+        error instanceof Error ? error.name : 'unknown error'
+      );
+    }
   };
 
   const content = {
--- a/src/apps/popup/pages/download-account-keys/index.tsx
+++ b/src/apps/popup/pages/download-account-keys/index.tsx
@@ -3,7 +3,8 @@ import React, { useCallback, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 
 import { PasswordProtectionPage } from '@popup/pages/password-protection-page';
-import { RouterPath, useTypedNavigate } from '@popup/router';
+
+import { closeExportKeysSurface } from '@background/open-export-keys-surface';
 
 import { createAsymmetricKeys } from '@libs/crypto/create-asymmetric-key';
 import {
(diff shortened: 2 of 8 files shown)


Comp0te (commented):
Reviewed with focus on logic errors and security regressions. Root-cause analysis and the window-based approach hold up under verification: the popup app uses `HashRouter`, so the `popup.html#/download-account-keys` deep link resolves; all step transitions are local state with no leftover `navigate(-1)`; vault lock swaps to `LockedRouter`, which unmounts the flow and resets `isPasswordConfirmed`, so re-unlock requires the password again; and the password gate itself is untouched as described.

### 1. `closeExportKeysSurface` can close the user's entire browser window

The repo already has `closeCurrentWindow` (`src/background/close-current-window.ts`), which guards with `currentWindow.type === 'popup'` before calling `windows.remove`. The new helper drops that guard.

If this page ever renders outside the dedicated window — e.g. a user opens `popup.html#/download-account-keys` directly in a tab, which is possible in all three browsers since extension pages are URL-addressable — `windows.getCurrent()` returns the **normal browser window**, and Close/Done closes the whole window with every tab in it. The existing guard exists precisely for this case.

Suggest reusing `closeCurrentWindow` (the new helper is a near-duplicate of it otherwise) or adding the same `type === 'popup'` check. Worth fixing before merge.

### 2. No window reuse — repeated clicks spawn multiple export windows

`createOpenWindow` tracks and refocuses an existing window; `openExportKeysSurface` doesn't. With the export window open, reopening the toolbar popup and clicking the menu item again creates a second independent export surface. Not a leak, but inconsistent with the established pattern — and the error copy even says "Close any other wallet windows and try again."

### 3. Stuck-disabled state in the shared password page

In `PasswordProtectionPage.onSubmit`, the `onClick(password).then(...)` chain has no `.catch`. If a caller's `onClick` rejects, `isSubmitting` stays `true` forever — previously that froze only the button; with this PR the field freezes too, across all five flows that use this page. Also, disabling a focused input drops focus, so after a wrong password the user has to click back into the field (`autoFocus` won't refire). A `.catch`/`finally` resetting `isSubmitting` would fix both.

### Nits

- Success still renders immediately after the fire-and-forget `downloadFile` — acknowledged in the PR description; the "check your Downloads folder" copy is an honest mitigation.
- `zip.file(`${account.name}_secret_key.pem`)` — account names containing `/` create folders inside the zip. Pre-existing, unchanged by this PR, just noting it survives the move.

### Security regression verdict

No regression found. Key material never reaches logs (`error.name` only), the password gate is untouched, vault lock/timeout still applies inside the new window, and "Download again" without re-entering the password is scoped to the already-authenticated window session — equivalent to the old popup behavior. The removal of `hide: isSafariBuild` re-enables the feature on Safari, which is the point of the fix. The one genuinely risky spot is finding 1 — a data-loss-adjacent hazard (closing the user's browser window), not an exposure of secrets.

PR #1410, opened 2026-07-21
fix: WALLET-1347 — render QR codes dark-on-light in both themes

ost-ptk:
[WALLET-1347](https://make-software.atlassian.net/browse/WALLET-1347)

## Problem

QA reported that Android devices sometimes fail to scan the sync QR code while the extension is in dark theme, with no issues in light theme.

All three QR call sites passed theme colours to `qrcode.react`, so in dark theme the symbol was rendered **inverted** — white modules (`#FFFFFF`) on `#262730`. ISO/IEC 18004 defines a QR symbol as dark modules on a light background; decoding an inverted symbol is an optional extra that scanners cannot be relied on to attempt. A mostly-dark screen also pushes the phone camera into a longer exposure, which blooms the bright modules into their darker neighbours.

Two further problems made the symbol harder to resolve than it needed to be:

- **No quiet zone.** `qrcode.react` v4 defaults `marginSize` to `0`, so the only quiet zone was the container's fixed `16px` padding. That padding is measured in pixels while the quiet zone requirement is measured in modules, so it silently drops below the required 4 modules as soon as the modules get bigger.
- **Error-correction level `H`.** 30% recovery buys nothing on a screen — there is no print damage to recover from — but it pushed a 204-byte sync chunk to 77×77 modules, i.e. 3.84 CSS px per module inside a 360px popup.

## Change

A shared `QrCode` component (`src/libs/ui/components/qr-code/qr-code.tsx`) is now the single place these parameters live. It pins the symbol to black-on-white in both themes, carries its own 4-module quiet zone, and drops to level `M`. Each constant carries a comment explaining *why*, so that "hardcoded colours in a themed component" doesn't get helpfully reverted later.

Because the quiet zone now lives inside the canvas, the container's background colour no longer affects decoding, and its padding became redundant:

- **`wallet-qr-code`** — the card gives its full content width to the symbol (296 → 328px). It no longer needs a background of its own, since the canvas covers it edge to edge.
- **`account-settings`** — this was the smallest QR we render on the screen with the most room to spare. 120 → 240px, and it is now centred; previously the bare `<canvas>` was `display: inline` inside a plain block, so it sat left-aligned.
- **`receive`** — unchanged apart from adopting the shared component.

Module size per screen, before → after:

| Screen | Before | After | |
|---|---|---|---|
| `wallet-qr-code` | 3.84px | **5.05px** | +31% |
| `account-settings` | 2.45px | **5.33px** | +118% |
| `receive` | 6.04px | **6.58px** | +9% |

These figures were computed by running the payloads through the encoder bundled in `qrcode.react`, not estimated.

## Testing

`npm run ci-check` passes. The QR rendering itself has no unit test — `jest.config.js` runs on `testEnvironment: 'node'` with no jsdom, and `QRCodeCanvas` needs a canvas — so this needs a QA pass on the Android device where the original failure was seen, rather than a "looks fine on my machine".

Worth checking during review: the layout changes on `wallet-qr-code` (full-width card) and `account-settings` (240px, centred) were made without a designer, and are the two things most likely to want adjusting.

[WALLET-1347]: https://make-software.atlassian.net/browse/WALLET-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/libs/ui/components/qr-code/qr-code.tsx
+++ b/src/libs/ui/components/qr-code/qr-code.tsx
@@ -1,0 +1,50 @@
+import { QRCodeCanvas } from 'qrcode.react';
+import React from 'react';
+
+/**
+ * QR symbols are always rendered dark-on-light, in both themes.
+ *
+ * ISO/IEC 18004 defines a symbol as dark modules on a light background.
+ * Reading an inverted (light-on-dark) symbol is an optional extra that many
+ * decoders never attempt — ZXing, which backs most Android scanners, gives up
+ * on them — so painting the code with the dark palette made it unscannable on
+ * some phones (WALLET-1347). A mostly-dark screen also pushes the phone camera
+ * into a longer exposure, which blooms the bright modules into their darker
+ * neighbours.
+ */
+const QR_FOREGROUND_COLOR = '#000000';
+const QR_BACKGROUND_COLOR = '#FFFFFF';
+
+/**
+ * Quiet zone, in modules. `qrcode.react` leaves it out by default, which left
+ * us relying on the container's padding — and that padding is a fixed number
+ * of pixels, so it silently drops below the 4 modules the spec requires as
+ * soon as the modules get bigger.
+ */
+const QR_MARGIN_SIZE = 4;
+
+/**
+ * On a screen there is nothing to recover from: no print damage, no smudges.
+ * The extra error correction of level `H` only buys density — a 200-char sync
+ * chunk needs 77x77 modules at `H` against 57x57 at `M` — and density is what
+ * actually breaks scanning on a small popup. `boostLevel` (on by default) still
+ * upgrades the level for free whenever it fits in the same version.
+ */
+const QR_ERROR_CORRECTION_LEVEL = 'M';
+
+interface QrCodeProps {
+  value: string;
+  size: number;
+}
+
+export const QrCode = ({ value, size }: QrCodeProps) => (
+  <QRCodeCanvas
+    id="qrCode"
+    value={value}
+    size={size}
+    fgColor={QR_FOREGROUND_COLOR}
+    bgColor={QR_BACKGROUND_COLOR}
+    marginSize={QR_MARGIN_SIZE}
+    level={QR_ERROR_CORRECTION_LEVEL}
+  />
+);
--- a/src/apps/popup/pages/account-settings/content.tsx
+++ b/src/apps/popup/pages/account-settings/content.tsx
@@ -1,9 +1,8 @@
-import { QRCodeCanvas } from 'qrcode.react';
 import React, { useCallback } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
-import styled, { useTheme } from 'styled-components';
+import styled from 'styled-components';
 
 import { RouterPath, useTypedNavigate } from '@popup/router';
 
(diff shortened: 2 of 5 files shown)


ost-ptk:
@comp0te — one thing this PR deliberately does **not** touch, since it needs a decision on the mobile side.

The real driver of QR density is the payload size, and there is free headroom in it. In `src/background/workers/generate-sync-wallet-qr-data-worker.ts`:

```ts
const qrString = JSON.stringify([b64cipher, b64salt, b64iv]);
const qrBytes  = Uint8Array.from(Buffer.from(qrString));
const qrData   = convertBytesToBase64(qrBytes);   // base64 on top of base64
const qrDataArray = qrData.match(/.{1,200}/g);
```

Two observations:

1. **The outer base64 is redundant.** `qrString` is a JSON array of values that are already base64, i.e. pure ASCII. Encoding it again inflates the payload by ~33% for no benefit. Dropping it would mean ~25% fewer chunks at the same chunk size.

2. **The 200-char chunk size can be lowered.** 200 chars → 65 modules; 120 → 57 (+14% module size); 100 → 49 (+33%). The cost is more frames in the animation, so a longer scan.

On the mobile side I checked `getQrDataFromMultiQrs` (`src/utils/common.ts`) in `casper-wallet-mobile`:

```ts
multiQrContent.map(qr => qr.split('$')).sort((a, b) => Number(a[0]) - Number(b[0])).map(arr => arr[2]).join('')
```

Nothing assumes a chunk size — it splits on `$`, sorts by index and concatenates, and the base64 alphabet contains no `$` — so **(2) is safe to change unilaterally**, the app would not notice.

**(1) is not.** `QRCodePasswordConfirmation/use-screen.ts` mirrors the extension exactly (`toByteArray → Buffer → JSON.parse → [cipher, salt, iv]`), so removing the outer base64 requires a coordinated change in both repos and a compatibility story for older app versions.

Nothing in the mobile repo was modified. Flagging it in case it is worth a follow-up ticket — if the fix in this PR turns out not to be enough on its own, payload size is the next lever, and it is a much bigger one than anything available in the renderer.

src/libs/ui/components/qr-code/qr-code.tsx
```diff
@@ -0,0 +1,50 @@
...
+const QR_ERROR_CORRECTION_LEVEL = 'M';
+
+interface QrCodeProps {
+  value: string;
+  size: number;
+}
+
+export const QrCode = ({ value, size }: QrCodeProps) => (
+  <QRCodeCanvas
+    id="qrCode"
```
Comp0te, line 42: Minor: the hardcoded `id="qrCode"` came along from the old call sites, but nothing references it — there are no `getElementById('qrCode')` usages in the repo. Since this is now a shared component, a fixed id also means duplicate DOM ids if two QR codes ever end up on the same screen. Suggest dropping the `id` prop entirely.

Comp0te:
@ost-ptk Thanks for the thorough analysis — I checked both points against the code and they hold up: the outer base64 is indeed redundant (pure-ASCII JSON re-encoded, +33%), and the mobile parser is genuinely chunk-size-agnostic (`getQrDataFromMultiQrs` splits on `$`, sorts numerically, no size assumption).

That said, we're leaving the payload as is for now. The reported failure was theme-dependent — light theme scanned fine even at 3.84px modules with no quiet zone — so this PR's fix should be sufficient on its own, and the base64 drop needs a coordinated mobile change with a staged rollout, which isn't worth it while the problem is expected to be solved.

Filed [WALLET-1356](https://make-software.atlassian.net/browse/WALLET-1356) to capture the analysis and the rollout plan (mobile ships a tolerant parser first — the raw payload starts with `[`, which can't begin a base64 string, so no version field needed). We'll pick it up if QA still sees scan issues after 2.6.0 or if sync payloads grow enough for the frame count to hurt.

[WALLET-1356]: https://make-software.atlassian.net/browse/WALLET-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1414, opened 2026-07-21
fix: WALLET-1347 — render QR codes dark-on-light in both themes

ost-ptk:
[WALLET-1347](https://make-software.atlassian.net/browse/WALLET-1347)

> Replaces #1410, which was opened against `release/2.6.0` and closed when that branch went away. Same commit, retargeted at `release/2.7.0`. The discussion on the payload format is on the old PR; it is repeated as a comment below so it does not get lost.

## Problem

QA reported that Android devices sometimes fail to scan the sync QR code while the extension is in dark theme, with no issues in light theme.

All three QR call sites passed theme colours to `qrcode.react`, so in dark theme the symbol was rendered **inverted** — white modules (`#FFFFFF`) on `#262730`. ISO/IEC 18004 defines a QR symbol as dark modules on a light background; decoding an inverted symbol is an optional extra that scanners cannot be relied on to attempt. A mostly-dark screen also pushes the phone camera into a longer exposure, which blooms the bright modules into their darker neighbours.

Two further problems made the symbol harder to resolve than it needed to be:

- **No quiet zone.** `qrcode.react` v4 defaults `marginSize` to `0`, so the only quiet zone was the container's fixed `16px` padding. That padding is measured in pixels while the quiet zone requirement is measured in modules, so it silently drops below the required 4 modules as soon as the modules get bigger.
- **Error-correction level `H`.** 30% recovery buys nothing on a screen — there is no print damage to recover from — but it pushed a 204-byte sync chunk to 77×77 modules, i.e. 3.84 CSS px per module inside a 360px popup.

## Change

A shared `QrCode` component (`src/libs/ui/components/qr-code/qr-code.tsx`) is now the single place these parameters live. It pins the symbol to black-on-white in both themes, carries its own 4-module quiet zone, and drops to level `M`. Each constant carries a comment explaining *why*, so that "hardcoded colours in a themed component" doesn't get helpfully reverted later.

Because the quiet zone now lives inside the canvas, the container's background colour no longer affects decoding, and its padding became redundant:

- **`wallet-qr-code`** — the card gives its full content width to the symbol (296 → 328px). It no longer needs a background of its own, since the canvas covers it edge to edge.
- **`account-settings`** — this was the smallest QR we render on the screen with the most room to spare. 120 → 240px, and it is now centred; previously the bare `<canvas>` was `display: inline` inside a plain block, so it sat left-aligned.
- **`receive`** — unchanged apart from adopting the shared component.

Module size per screen, before → after:

| Screen | Before | After | |
|---|---|---|---|
| `wallet-qr-code` | 3.84px | **5.05px** | +31% |
| `account-settings` | 2.45px | **5.33px** | +118% |
| `receive` | 6.04px | **6.58px** | +9% |

These figures were computed by running the payloads through the encoder bundled in `qrcode.react`, not estimated.

## Testing

`npm run ci-check` passes. The QR rendering itself has no unit test — `jest.config.js` runs on `testEnvironment: 'node'` with no jsdom, and `QRCodeCanvas` needs a canvas — so this needs a QA pass on the Android device where the original failure was seen, rather than a "looks fine on my machine".

Worth checking during review: the layout changes on `wallet-qr-code` (full-width card) and `account-settings` (240px, centred) were made without a designer, and are the two things most likely to want adjusting.

[WALLET-1347]: https://make-software.atlassian.net/browse/WALLET-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/popup/pages/account-settings/content.tsx
+++ b/src/apps/popup/pages/account-settings/content.tsx
@@ -1,9 +1,8 @@
-import { QRCodeCanvas } from 'qrcode.react';
 import React, { useCallback } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
-import styled, { useTheme } from 'styled-components';
+import styled from 'styled-components';
 
 import { RouterPath, useTypedNavigate } from '@popup/router';
 
@@ -30,14 +29,23 @@ import { useFetchAccountsInfo } from '@libs/services/account-info';
 import {
   Hash,
   HashVariant,
+  QrCode,
   SvgIcon,
   Tile,
   Typography
 } from '@libs/ui/components';
 
+// At 120px the modules were 2.45px each — the smallest QR we render, on the
+// screen with the most room to spare (WALLET-1347). The tile gives us 296px;
+// 240 takes the modules to 5.33px and still reads as an inset thumbnail rather
+// than taking over the screen.
+const QrCodeContainer = styled(VerticalSpaceContainer)`
+  display: flex;
+  justify-content: center;
+`;
+
 export function AccountSettingsPageContent() {
   const { t } = useTranslation();
-  const theme = useTheme();
 
   const { accountName } = useParams();
   const account = useSelector((state: RootState) =>
@@ -60,16 +68,9 @@ export function AccountSettingsPageContent() {
         <Tile>
           <TileContainer paddingVertical={SpacingSize.XL}>
             <Typography type="header">{account.name}</Typography>
-            <VerticalSpaceContainer top={SpacingSize.XL}>
-              <QRCodeCanvas
-                id="qrCode"
-                value={account.publicKey}
-                size={120}
-                fgColor={theme.color.contentPrimary}
-                bgColor={theme.color.backgroundPrimary}
-                level={'H'}
-              />
-            </VerticalSpaceContainer>
+            <QrCodeContainer top={SpacingSize.XL}>
+              <QrCode value={account.publicKey} size={240} />
+            </QrCodeContainer>
             <VerticalSpaceContainer top={SpacingSize.XL}>
               <FlexColumn gap={SpacingSize.Small}>
                 <Typography type="bodySemiBold">
--- a/src/apps/popup/pages/receive/content.tsx
+++ b/src/apps/popup/pages/receive/content.tsx
@@ -1,8 +1,7 @@
-import { QRCodeCanvas } from 'qrcode.react';
 import React, { useEffect, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
-import styled, { useTheme } from 'styled-components';
+import styled from 'styled-components';
 
 import { formatCep18Tokens } from '@popup/pages/home/components/tokens-list/utils';
 import { useTypedLocation } from '@popup/router';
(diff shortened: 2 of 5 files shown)


ost-ptk:
@comp0te — reposting from #1410 so it isn't lost with the closed PR. One thing this PR deliberately does **not** touch, since it needs a decision on the mobile side.

The real driver of QR density is the payload size, and there is free headroom in it. In `src/background/workers/generate-sync-wallet-qr-data-worker.ts`:

```ts
const qrString = JSON.stringify([b64cipher, b64salt, b64iv]);
const qrBytes  = Uint8Array.from(Buffer.from(qrString));
const qrData   = convertBytesToBase64(qrBytes);   // base64 on top of base64
const qrDataArray = qrData.match(/.{1,200}/g);
```

Two observations:

1. **The outer base64 is redundant.** `qrString` is a JSON array of values that are already base64, i.e. pure ASCII. Encoding it again inflates the payload by ~33% for no benefit. Dropping it would mean ~25% fewer chunks at the same chunk size.

2. **The 200-char chunk size can be lowered.** 200 chars → 65 modules; 120 → 57 (+14% module size); 100 → 49 (+33%). The cost is more frames in the animation, so a longer scan.

On the mobile side I checked `getQrDataFromMultiQrs` (`src/utils/common.ts`) in `casper-wallet-mobile`:

```ts
multiQrContent.map(qr => qr.split('$')).sort((a, b) => Number(a[0]) - Number(b[0])).map(arr => arr[2]).join('')
```

Nothing assumes a chunk size — it splits on `$`, sorts by index and concatenates, and the base64 alphabet contains no `$` — so **(2) is safe to change unilaterally**, the app would not notice.

**(1) is not.** `QRCodePasswordConfirmation/use-screen.ts` mirrors the extension exactly (`toByteArray → Buffer → JSON.parse → [cipher, salt, iv]`), so removing the outer base64 requires a coordinated change in both repos and a compatibility story for older app versions.

Nothing in the mobile repo was modified. Flagging it in case it is worth a follow-up ticket — if the fix in this PR turns out not to be enough on its own, payload size is the next lever, and it is a much bigger one than anything available in the renderer.

ost-ptk:
@comp0te thanks — both points addressed.

**`id="qrCode"`** (your inline comment on #1410): dropped in e7743e1d. You were right that nothing references it; I'd kept it on the theory that QA tooling might, but there are no `getElementById('qrCode')` or `#qrCode` usages anywhere in `src` or `e2e-tests`, so it was carrying no weight.

**On the payload / WALLET-1356**: agreed, and the `[` discriminator is a better answer than the versioned-format compatibility story I had in mind — a raw payload can't be mistaken for base64, so the tolerant parser needs no extra signalling at all.

One correction to my own analysis, since it changes how this PR should be read. You pointed out that light theme scanned fine at the same 3.84px modules with no quiet zone. That's decisive, and I under-weighted it: **density was not the cause**. Of the three changes here, only the polarity fix addresses the reported failure; the level `H` → `M` drop and the 4-module quiet zone are hardening that happens to be free, not part of the fix. My PR description presents the three as roughly co-equal contributors, which overstates the latter two. The changes are still worth keeping on their own merits — a fixed-pixel padding that silently falls below the spec minimum is a real latent bug, and 30% error correction genuinely buys nothing on a screen — but they are not why the scan failed.

Which also means the QA retest is narrower than I framed it: the thing to verify is that dark theme now scans on the Android device where it failed. If it still fails, polarity was the wrong hypothesis and the density work here won't save it — that's when WALLET-1356 becomes relevant.

Note CI is red on this branch for an unrelated reason: the `Audit runtime dependencies` step fails on `brace-expansion <1.1.16` (GHSA-3jxr-9vmj-r5cp), which is being handled separately.

PR #1417, opened 2026-07-22
WALLET-1351 — LavaMoat install-time supply-chain protection (@lavamoat/allow-scripts)

Comp0te:
## What & why

Harden the extension against install-time supply-chain attacks — malicious npm lifecycle scripts (`preinstall`/`install`/`postinstall`) that run arbitrary code on every developer machine and CI runner at install time (the most common vector in recent npm incidents). This is the install-time layer MetaMask uses for the same threat model.

- **`.npmrc`** → `ignore-scripts=true`: no lifecycle scripts run by default, including the `prepare`/husky hook.
- Add **`@lavamoat/allow-scripts`** + **`@lavamoat/preinstall-always-fail`** (a tripwire whose `preinstall` fails loudly if `ignore-scripts` is ever bypassed).
- **`lavamoat.allowScripts` allowlist** — everything disallowed by default; only two approved:
  - `unrs-resolver` → `true` (native binding needed by eslint import-resolver + jest-resolve)
  - `esbuild` → `true` (needed by i18next-parser locale extraction)
  - `geckodriver` → `false` (its binary is fetched lazily at test time by the Firefox smoke test), tripwire → `false`.
- **`npm run setup`** (`npm ci` + `allow-scripts` + `husky`) is the new canonical local install; documented in the README.
- **CI**: `npm run allow-scripts` runs after `npm ci` and fails the build on any install script not in the allowlist.
- `knip.json`: ignore the tripwire package (never imported by design).

Runtime hardening (`@lavamoat/webpack` SES policy + policy-drift CI gate) is tracked separately as follow-up.

## Note

This branch also includes two additional commits: a `keep-alive` ping-rejection fix and the Xcode `project.pbxproj` marketing-version bump (2.5.0 → 2.7.0).

### DIFF
--- a/.github/workflows/ci-cd-manual.yml
+++ b/.github/workflows/ci-cd-manual.yml
@@ -41,6 +41,12 @@ jobs:
     - name: Install
       run: npm install
 
+    # Run approved dependency install scripts (.npmrc sets ignore-scripts=true)
+    # and fail the release build if a dependency ships an install script not in
+    # the `lavamoat.allowScripts` allowlist — same supply-chain gate as ci-check.
+    - name: Allow approved install scripts
+      run: npm run allow-scripts
+
     - name: Build
       run: npm run build:chrome
 
--- a/.github/workflows/ci-check.yml
+++ b/.github/workflows/ci-check.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         node-version: [22.x]
-    
+
     steps:
     - uses: actions/checkout@v4
 
@@ -39,6 +39,11 @@ jobs:
     - name: Install & Check
       run: |
         npm ci
+        # Run approved dependency install scripts (.npmrc sets ignore-scripts=true).
+        # Builds the unrs-resolver native binding needed by eslint/jest, and fails
+        # the build if a dependency ships an install script not in the
+        # `lavamoat.allowScripts` allowlist.
+        npm run allow-scripts
         npm run ci-check
 
     # Blocking: custom wallet rules (.semgrep.yml) + security-focused registry
--- a/.github/workflows/e2e-onboarding-tests.yml
+++ b/.github/workflows/e2e-onboarding-tests.yml
@@ -22,6 +22,12 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # Run approved dependency install scripts (.npmrc sets ignore-scripts=true)
+      # and fail the build if a dependency ships an install script not in the
+      # `lavamoat.allowScripts` allowlist — same supply-chain gate as ci-check.
+      - name: Allow approved install scripts
+        run: npm run allow-scripts
+
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps
 
--- a/.github/workflows/e2e-popup-tests.yml
+++ b/.github/workflows/e2e-popup-tests.yml
@@ -22,6 +22,12 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # Run approved dependency install scripts (.npmrc sets ignore-scripts=true)
+      # and fail the build if a dependency ships an install script not in the
+      # `lavamoat.allowScripts` allowlist — same supply-chain gate as ci-check.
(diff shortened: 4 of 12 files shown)


ost-ptk:
### Code review

Found 1 issue:

1. The `allow-scripts` gate was added only to `ci-check.yml`. The release workflow that builds and publishes the Chrome Web Store artifact still runs a bare `npm install`, and both e2e workflows run `npm ci`, all now under the new root `.npmrc` (`ignore-scripts=true`) — with no `npm run allow-scripts` step. Nothing breaks today (`build:chrome`/`package:chrome` need neither `unrs-resolver` nor `esbuild`), but the supply-chain gate does not cover the pipeline that actually ships the artifact: a future dependency requiring an approved install script would fail loudly in `ci-check.yml` while the release job silently produces a subtly broken zip.

Release workflow: https://github.com/make-software/casper-wallet/blob/517652b4ef424392b1d23c8b14eb711cc2eae02a/.github/workflows/ci-cd-manual.yml#L40-L44

e2e workflows: https://github.com/make-software/casper-wallet/blob/517652b4ef424392b1d23c8b14eb711cc2eae02a/.github/workflows/e2e-popup-tests.yml#L21-L25 and https://github.com/make-software/casper-wallet/blob/517652b4ef424392b1d23c8b14eb711cc2eae02a/.github/workflows/e2e-onboarding-tests.yml#L21-L25

Gate as added: https://github.com/make-software/casper-wallet/blob/517652b4ef424392b1d23c8b14eb711cc2eae02a/.github/workflows/ci-check.yml#L39-L48

---

Verified as correct, for the record: the `lavamoat.allowScripts` allowlist is complete for the current tree. Running the real `@lavamoat/allow-scripts@5.1.0` detector (`loadAllPackageConfigurations`) against an installed macOS checkout reports exactly the three packages the allowlist covers — `eslint-import-resolver-typescript>unrs-resolver#1.12.2`, `geckodriver#6.1.0`, `i18next-parser>esbuild#0.25.12`. Note that `fsevents` shows `hasInstallScript: true` in `package-lock.json` but is **not** flagged: the tool reads `scripts` from the on-disk `package.json`, and the published fsevents 2.3.x tarball ships a prebuilt `fsevents.node` with no install/preinstall/postinstall script and no `binding.gyp`, so no allowlist entry is needed on macOS either.

Comp0te:
Fixed in 9e21c6fd. Added the `npm run allow-scripts` gate right after install in all three uncovered workflows:

- `ci-cd-manual.yml` (release) — after `npm install`
- `e2e-popup-tests.yml` — after `npm ci`
- `e2e-onboarding-tests.yml` — after `npm ci`

Every pipeline that installs under the new `.npmrc` (`ignore-scripts=true`) now enforces the same allowlist, so a future dependency requiring an unlisted install script fails loudly in the release/e2e jobs instead of silently shipping a broken artifact.

Thanks for the verification note on the allowlist completeness — no change needed there.

ost-ptk (approved):
### Code review — resolved

The issue from the previous review is fixed by 9e21c6fd: `npm run allow-scripts` now runs in `ci-cd-manual.yml`, `e2e-onboarding-tests.yml` and `e2e-popup-tests.yml`, in each case after the install step and before anything is built or run.

Re-checked after the `release/2.7.0` merge (dd11c5f6 rewrote ~848 lines of `package-lock.json`), since the version-pinned allowlist patterns could have drifted: they still match the lockfile exactly — `unrs-resolver 1.12.2`, `esbuild 0.25.12`, `geckodriver 6.1.0`, `@lavamoat/preinstall-always-fail 3.0.0` — and the merge introduced no new packages with install scripts, so the gate passes on the Linux runners. `allow-scripts` itself is available as a bin regardless of `ignore-scripts`, since bin linking is unaffected by it.

Non-blocking, and pre-existing: `ci-cd-manual.yml` installs with `npm install` rather than `npm ci`. Now that a fail-closed, version-pinned gate runs immediately after it, a `package.json`/lockfile drift would hard-fail the release build instead of just building through. Worth switching that step to `npm ci` at some point.

LGTM.

PR #1418, opened 2026-07-22
Remediate remaining security alerts on release/2.7.0

Comp0te:
`release/2.7.0`'s dependency upgrades already clear 60 of the 64 open Dependabot alerts. This PR closes the **4 that remain** plus the **10 open code-scanning alerts**, so the branch's Security tab is clean.

### Dependency advisories (all dev/build tooling — none reach the shipped bundle)
Stray transitive copies survived the upgrade; pinned via `overrides` to their patched releases:

| Package | Was | Now | Source |
|---|---|---|---|
| adm-zip | 0.5.18 | 0.6.0 | web-ext / xcode plugin |
| brace-expansion (2.x) | 2.1.1 | 2.1.2 | jest tooling |
| js-yaml (4.x) | 4.2.0 | 4.3.0 | markdownlint-cli |
| shell-quote | 1.8.4 | 1.10.0 | web-ext → fx-runner |

`npm run audit:ci` (runtime deps, high+) stays at **0 vulnerabilities**.

### Code scanning
- **Missing workflow permissions (6 alerts):** added least-privilege top-level `permissions:` blocks to the CI, E2E, manual-release and cache-cleanup workflows. `cleanup-caches` keeps `actions: write` since it purges caches; the rest are `contents: read`.
- **Overly-large regex range (3 alerts):** `/^[\daA-zZ\s]+$/` → `/^[\da-zA-Z\s]+$/` in account- and contact-name validation. The `A-z` span (ASCII 65–122) also accepted `` [ \ ] ^ _ ` ``; the class now matches only the intended alphanumerics + whitespace. Valid names are unaffected.
- **Shell-command injection (1 alert):** the local dev-server script now launches Chrome via `execFileSync` with an argument array instead of an interpolated `execSync` string, so the filesystem-derived extension path can no longer be parsed as a shell command.

Out of scope: `linkify-it` (ReDoS, high) surfaces in the full `npm audit` but is **not** among the 64 Dependabot alerts and is dev-only — left for a separate change.

### DIFF
--- a/.github/workflows/ci-cd-manual.yml
+++ b/.github/workflows/ci-cd-manual.yml
@@ -15,6 +15,9 @@ on:
          - "upload"
          - "publish"
 
+permissions:
+  contents: read
+
 jobs:
   check_version:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-check.yml
+++ b/.github/workflows/ci-check.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master, develop, release/** ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/cleanup-caches.yml
+++ b/.github/workflows/cleanup-caches.yml
@@ -11,6 +11,10 @@ on:
 #   This script deletes the cache which was connected to this branch after PR was closed.
 #   More information can be found here - https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries
 
+permissions:
+  contents: read
+  actions: write
+
 jobs:
   cleanup:
     runs-on: ubuntu-latest
--- a/.github/workflows/e2e-onboarding-tests.yml
+++ b/.github/workflows/e2e-onboarding-tests.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master, develop ]
 
+permissions:
+  contents: read
+
 jobs:
   e2e-onboarding-tests:
     timeout-minutes: 60
--- a/.github/workflows/e2e-popup-tests.yml
+++ b/.github/workflows/e2e-popup-tests.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master, develop ]
 
+permissions:
+  contents: read
+
 jobs:
   e2e-popup-tests:
     timeout-minutes: 60
--- a/package-lock.json
+++ b/package-lock.json
@@ -1842,9 +1842,9 @@
       }
     },
     "node_modules/@jest/reporters/node_modules/brace-expansion": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.1.tgz",
-      "integrity": "sha512-WR1cURNjuvBLMZBMbqM0UoE+WAfdUcEV1ccD8PVBVOI+Z3ND4+SZbN8RsfT2bMuG1qwz5RFvPukSZm5fF2D5eA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
+      "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
       "dev": true,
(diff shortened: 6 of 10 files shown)


ost-ptk:
### Code review

Found 1 blocking issue, plus 2 non-blocking notes.

1. Tightening the name regex blocks renaming accounts/contacts whose stored name was legal under the old rule.

The `A-z` span (ASCII 65-122) additionally admitted `` [ \ ] ^ _ ` ``, and it has been in place since `fb19a717` (2022-05-25), so names such as `Main_Wallet` can already be persisted. The rename form seeds itself with the stored name and validates on every keystroke, so an affected user hits "Account name can't contain special characters" on the first edit and can no longer rename at all. Same for contacts and for the derived-accounts list. No e2e or unit fixture uses these characters, so CI will not catch it.

https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/src/libs/ui/forms/form-validation-rules.ts#L114-L119

https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/src/libs/ui/forms/form-validation-rules.ts#L609-L614

https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/src/libs/ui/components/dynamic-accounts-list-with-select/dynamic-accounts-list-with-select.tsx#L326-L331

Evidence that the stored value is re-validated:

https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/src/libs/ui/forms/rename-account.ts#L26-L32

Suggestion: `/^[\da-zA-Z_\s]+$/` keeps the underscore, drops the genuinely unintended `` [ \ ] ^ ` ``, and still closes the `js/overly-large-range` alert since no large range remains in the class.

---

Non-blocking:

- The shell-injection remediation stops at the Chrome branch. Eight lines below, the Safari branch still interpolates `publicPath` (built from `process.env.PORT`) into an `execSync` string, which is the same construct the Chrome branch just moved away from. `.semgrep.yml` has no `child_process` rule either, so nothing guards against regression here.

  https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/utils/webserver.js#L118-L126

- `adm-zip: ^0.6.0` overrides `firefox-profile`'s declared `~0.5.x`. That subtree is only exercised by `web-ext run`, i.e. `npm run start:firefox`, which CI does not run. Worth confirming manually before merge.

  https://github.com/make-software/casper-wallet/blob/7e493bfd4bd6b087e153ce6f702d11de63bab8b2/package.json#L125-L131

---

Verified clean: all four override floors cover their advisory ranges (`adm-zip <0.6.0`, `shell-quote <=1.8.4`, `js-yaml 4.0.0-4.2.0`, `brace-expansion 2.0.0-2.1.1`); the lockfile retains no vulnerable copies; the `permissions:` blocks cover all five workflows with nothing under-scoped (`cleanup-caches` correctly keeps the `actions: write` its `gh actions-cache delete` needs, and `cws-publish` authenticates via its own secrets rather than `GITHUB_TOKEN`); and the `execFileSync` argv arrays are equivalent to the old shell tokenization, additionally fixing a latent break on build paths containing spaces.

ost-ptk (approved):
### Follow-up

`27a549af` addresses both review points. Verified:

- Name validation is now `/^[\da-zA-Z_\s]+$/` in all three rules, so already-stored names containing `_` re-validate cleanly in the rename forms. The genuinely unintended `` [ \ ] ^ ` `` stay rejected, and the class still contains no large range, so the `js/overly-large-range` alert remains closed. `git grep 'A-z'` over `src/` and `utils/` is empty.

  https://github.com/make-software/casper-wallet/blob/27a549afe1989daecd4ee2f1e56af78e833b5691/src/libs/ui/forms/form-validation-rules.ts#L114-L119

- The Safari dev launch now goes through `execFileSync` with an argv array, so the `PORT`-derived public path no longer reaches a shell. `execSync` is still imported and used by the Firefox branch, which interpolates a build-time constant rather than external input, so leaving that one as-is is consistent.

  https://github.com/make-software/casper-wallet/blob/27a549afe1989daecd4ee2f1e56af78e833b5691/utils/webserver.js#L118-L128

No test or e2e fixture asserts rejection of these characters, so nothing in CI changes behaviour either way.

One item remains open, and it needs no code change: `adm-zip: ^0.6.0` overrides `firefox-profile`'s declared `~0.5.x`, and that subtree is only exercised by `web-ext run`, i.e. `npm run start:firefox`, which CI does not cover. Please confirm it still launches before merge.

Approving.

PR #1419, opened 2026-07-23
fix: WALLET-1353 — cancel the abandoned approval request when the approval window is reused

ost-ptk:
## What & why

Follow-up to WALLET-1344 (PR #1405), implementing "Option A" from its pre-merge review.

The approval window is a single reused slot. When a dapp sends a second approval request while an earlier one is still open, all six SDK handlers reuse the window (navigating the same tab to the new request's page). Before this branch, the abandoned earlier request got **no response** — it stayed `'open'` and was only cancelled later, when the window was finally closed. For the sign→sign overlap this was a latency regression introduced by WALLET-1344 (before it, the signature pages cancelled via `beforeunload`).

This branch cancels the abandoned request **immediately at reuse time**, in the background, delivering the method-correct Cancel response to the dapp.

## How

- **Shared core** (`cancel-requests.ts`): extracted `CANCEL_GRACE_MS`, `buildCancelResponse`, and `cancelRequests(store, initiallyOpen, source, afterMark?)` out of `cancel-open-requests-on-close.ts`, which becomes a thin wrapper. The window-close path's observable behaviour is unchanged (pure refactor).
- **Supersede path** (`cancel-superseded-requests.ts`): `cancelSupersededRequests(store)` snapshots the open requests **synchronously**, then the six SDK handlers dispatch `windowRequestOpened` for the incoming request. Because the snapshot precedes that dispatch with no `await` in between, the incoming request cannot be in the snapshot — so **no `exceptRequestId` flag is needed**, and `openWindow`/`createOpenWindow` are left untouched. The same 250 ms grace + re-snapshot as the close path protects a genuine last-millisecond user response.
- **Housekeeping A**: `windowRequestResponded` now deletes the answered request's `pendingRequests` descriptor; `requests` is removed from the popup broadcast (`selectPopupState`) instead of pruned from the store.
- **Housekeeping B**: removed the dispatch-dead `windowClosed` action and narrowed `RequestStatus` to `'open' | 'responded'`.

## Deviations from the ticket text (intentional — flagging for the reporter)

1. **`exceptRequestId` not used.** The ticket suggested threading `exceptRequestId` through `openWindow`/`createOpenWindow`. We rely on snapshot **ordering** instead — strictly less surface area, and `openWindow` never needs to know about cancellation.
2. **`requests` is not pruned from the store.** The ticket asked to prune terminal `'responded'` entries from **both** maps. But the `'responded'` tombstone in `requests` is exactly what powers the duplicate-response dedup in `sdk-response-to-tab.ts` (`selectRequestStatus(...) === 'responded'`) — pruning it would reopen that dedup hole, whose main scenario is the cancel-vs-real-response race this very ticket concerns. So `pendingRequests` **is** pruned, and for `requests` we eliminate the real cost (per-broadcast serialization into every UI replica, which no UI reads) rather than the negligible heap growth.

## Known edges (accepted; QA awareness)

- **Live-store `requests` growth**: the map still grows for the service-worker lifetime (only broadcast cost was eliminated). Growth is ~tens of bytes per dapp request over a session — negligible; Chrome tears the SW down periodically. Accepted as-is.
- **Approve-then-supersede race**: if a user clicks Approve on request A and its signing is still computing when dapp B supersedes the window, the 250 ms grace can mark A `cancelled` and the genuine signature is silently discarded inside the extension. Security-reviewed: **no secret leak and no double-action** (dedup drops the late real response). The underlying window-reuse navigation race pre-dates this branch. Accepted as-is; noted for QA.

## Testing

- `npm run ci-check` green on the integrated branch (format, lint, tsc, knip, test:coverage; 100% on all touched handler/reducer files).
- New unit tests: supersede cancels the abandoned request but not the incoming one; answered-during-grace is not cancelled; delivery fallback + error surfacing; no-op with no open requests; call-ordering (`invocationCallOrder`) at the SDK handlers; fast-path (already-connected / already-signed) skips supersede.
- Reviews: per-task spec+quality gates, a whole-branch review, plus dedicated security and download-account-keys-interaction reviews — all clean (no Critical/High).
- **Manual extension QA pending.**

Jira: https://make-software.atlassian.net/browse/WALLET-1353

### DIFF
--- a/src/background/handlers/cancel-open-requests-on-close.test.ts
+++ b/src/background/handlers/cancel-open-requests-on-close.test.ts
@@ -6,7 +6,6 @@ import { sdkMethod } from '@content/sdk-method';
 
 import {
   CANCEL_GRACE_MS,
-  buildCancelResponse,
   cancelOpenRequestsForClosedWindow
 } from './cancel-open-requests-on-close';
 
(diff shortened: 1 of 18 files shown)


Comp0te (approved):
👍

Issue #18, opened 2022-01-20
As a Signer user I can review a list of sites that I've previously Connected to


ghost:
As a Casper Wallet user, I can review a list of dapps that I have connected to, broken down by account.
The user opens a Connected Sites screen from the Settings menu of Casper Wallet.
Any dapp that has been connected to with an account will be listed with it's URL and favicon.
A list of the accounts that have connected to that dapp will appear beneath the dapp URL as a sublist
As a Casper Wallet user, when I disconnect an account from a dapp, the account no longer appears beneath that dapp in the Connected Sites list.
As a Casper Wallet user, when I remove disconnect the only connected account from a dapp, the dapp is immediately no longer listed in the Connected Sites list.
As a Casper Wallet user, I can disconnect an individual account from a dapp by clicking on the X button in the Connect Sites screen.
As a Casper Wallet user, I can disconnect ALL accounts from a dapp at the same time by clicking the `Disconnect` button next to the Dapp URL.

Design: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Signer?node-id=101%3A567
![image.png](https://images.zenhubusercontent.com/61e1ac21d1fded72df402cc1/bbecce9c-5e76-40bb-b809-1295b12a87ad)

Issue #12, opened 2022-01-20
As a Signer user I can Connect/Disconnect to compatible dapps/sites


ghost:
Issues for this epic:
* User can Connect one or more accounts to any compatible site such as DotOracle or CSPR.Live. User is prompted to approve the connection before the connection is made.
* Once connected, the User can Disconnect any account from the site.
* As a user I see a status of Connected/disconnected
* Signer should save a list of Connected sites in storage
* UI to view and manage list of Connected Sites

Figma link is here: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Signer?node-id=115%3A855

1 sprint

Issue #11, opened 2022-01-20
As a Signer user I can view and manage a list of my accounts


ghost:
1. Import account
2. Delete an account
3. Rename an Account
4. Switch active account
~5. Create account~ moved to mnemonic phrase epic
~6. Security prompt~ - this is not part of this epic anymore, should be moved to secret phrase work

PR #1427, opened 2026-07-28
fix: WALLET-1364 — tie an approval request to the windows displaying it

ost-ptk:
Ticket: https://make-software.atlassian.net/browse/WALLET-1364 (P0, and P1 items 1, 2, 5–8)

> Items 3 (export-keys window failure feedback) and 4 (`CSP_NONCE` build reproducibility) are **not** in this change — they ship in #1428 and #1429. The earlier "items 1–8" here was wrong.

## The bug

A request's approval UI was not tied to any window, and "the old request is gone" was inferred from a 250 ms wall-clock timer. `cancelRequests` picked victims by `status === 'open'` alone, and `PendingRequestDescriptor` was `{tabId, origin, method}` — no window link at all.

The Ledger permission flow opens a **separate** window carrying a fully functional signing page with the same `requestId`, and device confirmation takes seconds to minutes. So when a second dapp sent any request:

1. dapp A's request was force-cancelled and A received `signResponse({ cancelled: true })`;
2. the device approval then produced a real `signResponse({ signatureHex })`, which hit the dedup guard, saw status `'responded'`, and was **dropped with no log at all**;
3. the UI closed as if it had succeeded. The user approved; the signature was gone; nothing was emitted anywhere.

The same defect existed on the close path: closing the tracked approval window cancelled a request the Ledger window was still displaying.

## The fix

A request stays open while at least one window still displays it. Cancellation is triggered by _"window W stopped displaying requests"_ — never by a timer, never by "a new request arrived".

- `requests` and `pendingRequests` collapse into one discriminated union, and each open request carries `windowIds: number[]`.
- Two events dispatch the same routine: `windows.onRemoved` (now for **any** window, not only the tracked one) and the resolution of `openWindow` with `reused: true`.
- Candidates are only requests that **had** that window and are left with none. A request the Ledger window still shows survives; a request that never had the window can never cancel itself.
- The trigger moves out of the six SDK method branches into `openWindow`, so the old "call cancel before dispatching the open" invariant is structural instead of a matter of statement order.
- The 250 ms grace now means only what its comment always claimed: let an in-flight genuine response land first.

Also fixed, all in the same paths:

- `windows.create` rejection now cancels the incoming request and tells the dapp, instead of leaving its promise pending for the 30-minute timeout; likewise a resolved window with no id, and a window that closes during the reuse round-trips.
- The dedup drop is logged — identifiers only, never the action payload, which carries `signatureHex` / `encryptedMessage`.
- `tabs.update` targets the resolved tab id instead of "the active tab of the current window".
- `isNewWindow` (the import-account flows) no longer retargets the shared approval slot.
- The `sagaError` banner is suppressed when the same-origin fallback actually delivered, so the user no longer reads an untranslated red error on top of the signing screen for a _different_ transaction.
- `PopupState.windowManagement` becomes `Pick<…, 'windowId'>` — fail-closed, so a future slice field is excluded from the replica broadcast by default.

## Review round — 15 follow-up commits

All 26 review comments were verified against the branch; none was a false positive. 24 are fixed here, one was already closed by #1430's `Pick`/broadcast-shape test, and two are taken partially on purpose (see _Deliberately partial_ below).

**The one that mattered most**: `cancelRequests` re-filtered its snapshot on status alone after the grace, so a request that _regained_ a window inside those 250 ms was still cancelled. That is this PR's own P0, reached from inside the fix — the Ledger attach crosses a `runtime.sendMessage` round trip and routinely lands after the candidates were snapshotted. It now re-checks the current descriptor.

The rest, by cluster:

|                         |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Silent skips**        | the Ledger attach could be skipped three ways with nothing in any log (missing `requestId` indistinguishable from the import flow, no `try/catch` around the effect, `dispatchToMainStore` swallowing the cause) — all three closed, the decision extracted into a tested unit, and `params` given a named type so renaming `requestId` is a compile error; a throw in `openWindow`'s success arm was caught by nothing; the `onRemoved` listener had no test at all and is now an exported `handleWindowRemoved` with one. |
| **Error surfaces**      | the cancel banner was suppressed on the close path too, where the supersede rationale does not apply; `failRequestOnWindowError` told the user "cancelled" without checking whether the dapp had been told; `openWindow` logged `error.name` (always the string "Error") instead of a redacted message; a dropped duplicate no longer conflates "a cancel raced a cancel" with "a signature was destroyed".                                                                                                                 |
| **Unstated invariants** | the detach dispatched on **every** window close in the browser, each one a full replica broadcast plus a full `storage.local` rewrite; the tombstone map was unbounded on the two MV2 targets whose background page never restarts; a replayed or reused `requestId` opened a fully functional approval screen for a request that could never be answered.                                                                                                                                                                  |
| **Types that lied**     | `windowIds` was mutable and handed out by reference; `selectOpenRequests` used an unchecked type predicate; the reducer's existence guards were dead code to the compiler; `SagaError.source` was still `string`. Each is now enforced — probes confirm that reverting any of them is a compile error.                                                                                                                                                                                                                      |
| **Trust boundaries**    | the window attach was the one handler path with no `sender` gate; `.match()` does not validate a payload, so a payload-less message threw out of the handler; the liveness probe accepted any live browser window and treated any rejection as "gone"; `windowIdChanged`/`windowIdCleared` were still forwardable with no legitimate UI dispatcher.                                                                                                                                                                         |

### Deliberately partial

- **`Request` union.** `{status: 'open', windowIds: []}` is both the legitimate transient state and the terminal stranded one. The suggested `'awaiting-window'` member with a non-empty tuple is a refactor through the reducer, selectors, both cancel paths, fixtures and every slice test. Here the _entry paths_ are closed instead (required `requestId`, the post-open `.catch`, the duplicate guard), so the stranded state is reachable only under `openWindow`'s control, whose failure branches both lead to `failRequestOnWindowError`. The union itself wants its own ticket.
- **Sender gating.** Applied to the `windowRequestWindowAttached` branch. Gating the whole `FORWARDED_ACTION_TYPES` is a behaviour change for ~60 actions and needs its own verification pass.

## Verification

`npm run ci-check` passes: 66 suites / 473 tests, `knip` clean, coverage floors held.

Mutation probes were run to prove the new tests are load-bearing rather than decorative:

- swapping the supersede and attach blocks in `openWindow` makes `open-window.supersede.test.ts` fail;
- widening the dedup condition from `=== 'responded'` to `!= null` fails exactly the new `'still open'` case, and nothing else;
- collapsing `handleWindowRemoved`'s two independent try/catches fails exactly its two error-handling tests;
- flipping the discriminant in `selectOpenRequests`, or dropping the reducer's null guard, are now compile errors — neither was before.

One bug was caught by its own test during this round: the tombstone eviction used `slice(0, length - CAP)`, and a negative end slices from the _back_, so it evicted almost everything until the map was full.

## Manual QA — cannot be automated

1. **Ledger, long confirmation.** Start a `sign` from dapp A, reach device confirmation, then trigger any request from dapp B. Expected: the Ledger window stays functional and dapp A resolves with its signature. Note the repro **requires** the permission-prompt path — with an already-paired device the confirmation happens inside the main window.
2. Same as (1), but trigger B's request in the narrow gap **while the Ledger window is opening** — this is the race the review found, and the one the grace re-check now covers.
3. Close the approval window via "X" while a Ledger window is open — the request must survive.
4. Ordinary supersede: A is cancelled exactly once, B's screen shows with no error banner over it, and the reused window now genuinely navigates to B.
5. Open "Import account with file" while a dapp approval is open — the approval must not be cancelled.

## Known follow-ups — not filed yet, not caused by this PR

- `closeNewLedgerWindowsAndClearState` closes **all** popup windows, including the main approval window, so after a supersede dapp B is force-cancelled when the Ledger flow finishes. Pre-existing and unchanged by this PR (B died before it too), but it wants its own ticket and its own QA.
- The `Request` union refactor and full forwarding-set gating described under _Deliberately partial_.


src/background/handlers/cancel-requests.ts
```diff
@@ -84,24 +91,93 @@ export async function cancelRequests(
   }
```
Comp0te, line 118: This re-check is unchanged by the PR — but the change is what makes it wrong, because `windowIds` and the "a window stopped displaying it" cancellation model are both new, and this line never consults them.

After `await delay(CANCEL_GRACE_MS)`, `toCancel` is re-filtered against `stillOpenIds` — status only. A request that *gained* a live window during the 250 ms grace is still marked `windowRequestResponded` and still has a cancel delivered to its dapp.

The sequence is the one this PR exists to close. Request R is displayed by shared approval window W (`windowIds: [W]`). The user triggers Ledger, so `openNewSeparateWindow` resolves and the attach begins its `runtime.sendMessage` round trip. If dapp B reuses W inside that gap, `cancelRequestsDisplacedBy` snapshots R as a candidate and detaches W. The Ledger attach then lands — R is `windowIds: [L]`, genuinely displayed, user confirming on the device. 250 ms later R is still `'open'`, so it is cancelled, and the device signature that follows is dropped by the dedup at `sdk-response-to-tab.ts:88`.

Reproducible directly against this function: call `cancelRequestsDisplacedBy(store, 7, 'cancel-on-supersede')` with `getState` returning `{r1: open(tab 3, windowIds: [7])}` at snapshot time and `{r1: open(tab 3, windowIds: [9])}` on every read after it. `tabs.sendMessage` fires once and `windowManagement/windowRequestResponded` is dispatched — for a request holding live window 9.

`attach-window-to-request.ts:12-18` calls this exact round trip "by far the wider race of the two". Suggest re-filtering on the *current* descriptor rather than the snapshot: keep only requests still `'open'` **and** whose `windowIds` is still empty.

src/hooks/use-ledger.ts
```diff
@@ -182,6 +182,18 @@ export const useLedger = ({
 
         if (w.id) {
           dispatchToMainStore(ledgerNewWindowIdChanged(w.id));
+
+          // This window displays the SAME requestId as the approval window that
+          // opened it. Registering it is what keeps the request alive when the
+          // shared approval window is reused or closed while the user is still
+          // confirming on the device.
+          const requestId = askPermissionUrlData.params?.requestId;
+          if (requestId) {
```
Comp0te: This guard is one of three ways the Ledger attach can be skipped with nothing in any log, on what is the fix's only UI path.

1. `if (requestId)` silently does nothing when `askPermissionUrlData.params?.requestId` is absent. That is correct for the import-account flow, which legitimately passes `params: {}` (`import-account-from-ledger/index.tsx:28-32`) — but `params` is `Record<string, string>` (`:39`), so a signing page arriving here without a `requestId` is indistinguishable from it at runtime.
2. The whole effect body is a bare `(async () => {…})()` at `:173-207` with no `try/catch` and no `.catch`. `openNewSeparateWindow` at `:181` is an awaited call that can reject, and a rejection there skips everything below it silently.
3. The dispatch goes through `dispatchToMainStore`, whose `.catch` at `redux/utils.ts:14-18` discards the error and logs only the action type — and a `runtime.sendMessage` to a sleeping or restarting MV3 service worker is a routine failure.

If any of the three fires, the Ledger window is never registered, `windowIds` stays `[approvalWindow]`, and the next dapp request reusing that window cancels this one while the user is confirming on the device — the P0 symptom, with nothing to explain it. `attach-window-to-request.ts` builds a careful liveness repair for a message that may simply never arrive.

Suggest: wrap the effect body in `try/catch` (logging `(error as Error)?.name` only — `url` embeds the plaintext `signMessage` message); branch explicitly on the flow (`askPermissionUrlData.domain !== 'popup.html'` is already used at `:225`) and `console.error` when a dapp-approval flow reaches here without a `requestId`; and give `dispatchToMainStore` an error-bearing log, or await this dispatch specifically and retry.

src/background/index.ts
```diff
@@ -177,18 +174,27 @@ tabs.onUpdated.addListener(async (tabId, changeInfo, tab) => {
...
-  }
+// for ANY window; `cancelOpenRequestsForClosedWindow` itself decides whether
+// the removed window was the last display for any open request (a request
+// the Ledger permission window still shows survives), waits a grace period,
+// then cancels the survivors and dispatches `windowIdCleared` to null the
+// slice `windowId` if it is still tracking the removed window. The `async`
+// IIFE is awaited only internally so a rejection here can never skip the
+// export-keys cleanup below it.
+windows.onRemoved.addListener((removedWindowId: number) => {
+  void (async () => {
```
Comp0te, line 186: This IIFE has no `.catch`, and the `try` below covers less than it looks.

`try/catch` at `:188-192` encloses only `cancelOpenRequestsForClosedWindow`. Both `await getExistingMainStoreSingletonOrInit()` on the next line and the export-keys branch at `:194-196` are outside it, so a rejection from either is discarded by `void` — and there is no `unhandledrejection` handler anywhere in `src/`.

The store init is a real source of one. `get-main-store.ts:68` declares `let storeSingleton: ReturnType<typeof createStore>` with no initializer and a non-nullable type, and `:237-241` catches a `storage.local.get` rejection and returns it anyway — so it can hand back `undefined` typed as `MainStore`, and `store.getState()` at `:194` then throws a `TypeError`.

To be precise about what is unlogged, since the `catch` does cover part of this: in that undefined-store case `cancelOpenRequestsForClosedWindow` throws first, inside the `try`, so `:191` does log it. What is never logged is that the *export-keys cleanup was skipped* as a result — the tracked window id stays set, and the next export-keys open focuses a dead id. A throw originating inside `:194-196` itself is the only wholly silent path.

Suggest `.catch(err => console.error('windows.onRemoved handler failed', err))` on the IIFE — or the `async`-listener shape suggested separately on this hunk, which removes the need for it.

src/background/index.ts
```diff
@@ -177,18 +174,27 @@ tabs.onUpdated.addListener(async (tabId, changeInfo, tab) => {
...
+    const store = await getExistingMainStoreSingletonOrInit();
+
+    try {
```
Comp0te: Separate point from the `.catch` comment above: nothing tests any of this listener.

Grepping `onRemoved` across every test file at this head returns four hits — `redux-actions.parity.test.ts:146,150`, `attach-window-to-request.test.ts:46`, `open-window.test.ts:219` — and all four are inside prose comments. No test file imports the background entry point, so the rewritten listener's shape, this `try/catch`, and the removal of the tracked-id guard are all unverified. `cancel-open-requests-on-close.test.ts` drives the handler directly and is indifferent to every one of them.

Concretely: reverting this listener to the previous `async` form — undoing both the guard removal and the `try/catch` — leaves the suite green. The two things that re-breaks are each the subject of an explicit ticket item.

Suggest extracting the body into an exported `handleWindowRemoved(store, removedWindowId)` and unit-testing it in the style of `cancel-open-requests-on-close.test.ts`.

src/background/handlers/cancel-requests.ts
```diff
@@ -84,24 +91,93 @@ export async function cancelRequests(
...
+  afterMark?: () => void
+): Promise<void> {
+  const candidates = selectOpenRequests(store.getState()).filter(
+    request =>
+      request.windowIds.length === 1 && request.windowIds[0] === windowId
+  );
+
+  // Dispatched synchronously, before the grace: the slice must stop claiming
+  // this window displays anything the moment it stopped doing so.
+  store.dispatch(windowDetachedFromRequests({ windowId }));
```
Comp0te: This dispatch is unconditional, and with the `removedWindowId === selectWindowId(...)` guard removed from the `onRemoved` listener (it was at `index.ts:186` on the base commit), that now means every window the user closes anywhere in the browser costs a full state broadcast and a full state persist.

The candidate filter does run four lines above at `:137-140` — the issue is that this dispatch is not *gated on* its result. It fires even when no open request holds `windowId` and the reducer returns the identical state object.

That still reaches every subscriber. `get-main-store.ts:178-235` does no state-change comparison: each fire performs a `runtime.sendMessage(popupStateUpdated(...))` broadcast to all replicas (`:183-187`) and a full `storage.local.set` re-write of vault cipher, keys, settings, contacts, app events, trusted wasm and CSPR-name expirations (`:216-230`). Redux invokes subscribers per dispatch regardless of what the reducer returned, so the unchanged state reference does not prevent any of it.

Before this change, an unrelated window close produced zero dispatches from this listener.

Suggest dispatching the detach only when some open request actually holds `windowId` — the filtered set at `:137-140` is already right there.

src/background/handlers/cancel-requests.ts
```diff
@@ -84,24 +91,93 @@ export async function cancelRequests(
...
-          })
+      } catch (error) {
+        // NEVER log `action` — it is an SDK response payload.
+        console.error(
+          `${source}: cancel delivery failed`,
+          { requestId, method, tabId },
+          error
         );
+        const delivered = await deliverViaOrigin(origin, action);
+        // Only surface a banner when the response is genuinely lost. On the
```
Comp0te: The `delivered === 0` condition is new — the base commit dispatched a `sagaError` on both outcomes, with the message branching on `delivered > 0`. The comment justifying the change reasons entirely from the supersede path ("this fires while the user is already looking at the NEXT approval screen"), but `cancelRequests` is shared: `cancel-open-requests-on-close.ts:24-29` reaches it with `'cancel-on-close'`, where that rationale does not apply.

It also matters that `deliverViaOrigin` is a weaker signal than the name suggests. `utils.ts:41-71` queries `{active: true}` tabs and counts same-origin sends that did not throw — that is not proof the tab owning the pending dapp promise received anything. So on the close path, "the direct send failed but N other same-origin tabs got the cancel" now suppresses the banner.

To be fair to what is still there: `:99-105` does add a `console.error` with `{requestId, method, tabId}` and the error for this case, so it is not silent. What was lost is the *user-visible* surface on a path where the user is not looking at a replacement screen.

Suggest gating the suppression on `source === 'cancel-on-supersede'` rather than on `delivered > 0` alone.

Comp0te (commented):
Reviewed the window-lifecycle rework: the `windowIds` model, the `onRemoved` and supersede cancellation paths, `attachWindowToRequest`, and the new discriminated union in the `windowManagement` slice.

The through-line of the comments below is that the new model makes "a window stopped displaying this request" the thing that decides a request's fate, while several paths still decide it from status alone, or register a window through a channel that can fail without saying so. A second, smaller cluster is about error surfaces the rework narrowed — a few places now log less, or stop short of telling the user, than the code they replace.

One note that has no line to sit on: the description opens with P0 + P1 items 1–8, but items 3 (export-keys window failure feedback) and 4 (`CSP_NONCE` build reproducibility) are not in this change. `src/background/redux/sagas/export-keys-window-saga.ts` still has the bare `console.error` with its "there is no popup left to show it in" comment, and `webpack.config.js:71` is still `crypto.randomBytes(16)` with the unconditional `DefinePlugin` entry at :212 — `gh pr diff --name-only` lists neither file. Worth narrowing the body to the items actually implemented.

src/background/handlers/cancel-requests.ts
```diff
@@ -84,24 +91,93 @@ export async function cancelRequests(
...
+
+  store.dispatch(windowRequestResponded({ requestId }));
+  store.dispatch(
+    sagaError({
+      source: 'open-window-failed',
+      message: 'Approval window could not be opened; the request was cancelled'
+    })
+  );
+
+  const action = buildCancelResponse(request.method, requestId);
```
Comp0te, line 211: This `await deliverViaOrigin(request.origin, action)` discards the return value, unlike the sibling path at `:106-119` which captures it and dispatches a `sagaError` when `delivered === 0`. `deliver-via-origin.ts:10-20` returns a delivery count precisely so callers can branch on it.

The consequence is a banner that says the wrong thing. `:165-170` dispatches "Approval window could not be opened; the request was cancelled" *before* this delivery is attempted, and `:164` has already written the `windowRequestResponded` tombstone. When both `tabs.sendMessage` and this same-origin fallback fail, the request is marked responded locally, the dapp received nothing and will hang — and the user has been told it is resolved. The tombstone then makes it unrecoverable, since `sdk-response-to-tab.ts:87-103` drops anything that arrives later.

Suggest capturing `delivered` and amending the banner (or dispatching a second `sagaError`) when it is 0.

src/background/handlers/cancel-requests.ts
```diff
@@ -84,24 +91,93 @@ export async function cancelRequests(
...
+              message: `Cancel delivery to tab ${tabId} failed; not delivered`
+            })
+          );
+        }
       }
     })
   );
 }
+
+// The single cancellation trigger: window `windowId` stopped displaying
```
Comp0te, line 160: "The single cancellation trigger" reads as an exhaustiveness claim, and `failRequestOnWindowError` at `:152` — 27 lines below, in this same file — is a second, independent one: it dispatches `windowRequestResponded` at `:164` and delivers a cancel response at `:172-181` without any window event being involved.

This header is the first thing a maintainer sees when scanning the file for "how does a request get cancelled", so it is worth it being right. Suggest "the window-driven cancellation trigger", with a cross-reference to `failRequestOnWindowError`.

src/background/handlers/cancel-requests.ts
```diff
@@ -18,7 +24,8 @@ export const CANCEL_GRACE_MS = 250;
 const delay = (ms: number) =>
   new Promise<void>(resolve => setTimeout(resolve, ms));
 
-export type OpenRequest = ReturnType<typeof selectOpenRequests>[number];
+export type CancelSource =
```
Comp0te: `CancelSource` constrains a parameter that has only two reachable values, and leaves unconstrained the field where sources actually land.

`'open-window-failed'` is in the union but no call site produces it *as a `CancelSource`* — `cancelRequests` is module-private and its only caller passes `'cancel-on-close'` or `'cancel-on-supersede'`. `'open-window-failed'` appears only as a raw literal inside `failRequestOnWindowError`'s `sagaError({ source: … })`. Meanwhile `SagaError.source` is still `string` (`app-events/types.ts:3`) — which is what WALLET-1364 #18 asked to have made a union — and another live source, `'sdk-response-to-tab'`, is not in this union at all.

Suggest moving it to `app-events/types.ts` as a `SagaErrorSource` covering all producers, typing `SagaError.source` with it; `CancelSource` then becomes `Extract<SagaErrorSource, 'cancel-on-close' | 'cancel-on-supersede'>`, or is not needed.

src/background/open-window.ts
```diff
@@ -2,21 +2,102 @@ import {
...
   createOpenWindow({
     windowId: selectWindowId(store.getState()),
     setWindowId: (id: number) => store.dispatch(windowIdChanged(id)),
     clearWindowId: () => store.dispatch(windowIdCleared())
-  })(openWindowProps).catch(error => {
-    // Fire-and-forget: if `windows.create` rejects, surface it instead of an
-    // unhandled rejection. The slice's window id is left cleared (no id was set).
-    console.error('openWindow: failed to open approval window', error);
-  });
+  })(openWindowProps).then(
```
Comp0te, line 69: The two-arm `.then(onFulfilled, onRejected)` form means a throw inside the **success** arm is caught by nothing — there is no trailing `.catch`.

The two-arm shape itself is deliberate and should stay, so the recovery cannot catch itself. The gap is narrower than it might look: the two `store.dispatch` calls at `:49-50` are inside callbacks that `createOpenWindow` invokes from within its own chain (`create-open-window.ts:107,147-149`), so they are already covered by the `onRejected` arm, and both `failIncomingRequest` calls `.catch` internally. What is genuinely unprotected is `attachWindowToRequest` at `:88`.

If it throws, the window opened but was never attached: `windowIds` stays `[]`, and per the `windowIds.length === 1` filter at `cancel-requests.ts:137-140` a request with an empty `windowIds` can never be cancelled by any window event. The dapp hangs for its full 30-minute timeout with nothing logged.

Suggest keeping both arms and appending `.catch(err => console.error('openWindow: post-open handling failed', err))`.

src/background/open-window.ts
```diff
@@ -2,21 +2,102 @@ import {
...
+} from '@background/handlers/cancel-requests';
 import { MainStore } from '@background/redux/get-main-store';
 import {
   windowIdChanged,
   windowIdCleared
 } from '@background/redux/windowManagement/actions';
 import { selectWindowId } from '@background/redux/windowManagement/selectors';
 
-export function openWindow(store: MainStore, openWindowProps: OpenWindowProps) {
+export interface OpenApprovalWindowProps extends OpenWindowProps {
```
Comp0te, line 17: `requestId?: string` is documented here as "absent for internal (non-approval) windows", but there is no such caller. `openWindow` has exactly one non-test importer (`sdk-methods.ts`) and all six call sites pass `requestId` — `grep -rn "@background/open-window" src/` shows this. The UI's internal-window path is a different function entirely (`useWindowManager` → `createOpenWindow`).

The optional is not free, because every branch that saves a request from being stranded is gated on `requestId != null`. A seventh approval flow that forgets the field compiles green, dispatches `windowRequestOpened`, opens a window that is never attached, and leaves `{status: 'open', windowIds: []}` — the state `failRequestOnWindowError` is never reached for.

Suggest making it required. Stronger: have `openWindow` take the `windowRequestOpened` payload and dispatch it itself, so "register the request" and "open its window" cannot be separated at any call site.

src/background/open-window.ts
```diff
@@ -2,21 +2,102 @@ import {
...
+        // update → tabs.update, see create-open-window.ts) before this `.then`
+        // runs, and window creation is itself an awaited round-trip, so the
+        // window can close in that gap. `attachWindowToRequest` owns both the
+        // dispatch and the liveness repair for that case — see the rationale
+        // there, and note the Ledger hook reaches the same helper across a
+        // wider gap still.
+        attachWindowToRequest(store, requestId, windowId);
+      }
+    },
+    error => {
```
Comp0te, line 107: This rejection arm logs `{requestId, errorName: (error as Error)?.name}` and nothing else, and it is the sole cause-bearing diagnostic on the "no approval window could be opened" path — the hard-to-reproduce failure the rest of the PR adds `failRequestOnWindowError` to handle.

The stated reason is right: a `signMessage` window URL embeds the user's plaintext message, and a rejection's text can echo the URL. The concern is that the mitigation drops the whole diagnostic rather than redacting it. `.name` on a `windows.create`/`windows.update`/`tabs.update` rejection is the string `"Error"` in the realistic cases and `undefined` for a non-`Error` rejection, so the message and stack — the only fields that say *why* it failed — are gone.

It is not a straight regression, to be fair: this adds `requestId`, which the base commit could not log because `openWindow` had no such parameter. The trade is "which request failed" gained, "why it failed" lost.

Suggest redacting rather than dropping — strip everything from `?` onward in `error.message`, or log `windowApp` plus a length-capped message. That keeps the secret out and the diagnosis in.

src/background/handlers/attach-window-to-request.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  store.dispatch(windowRequestWindowAttached({ requestId, windowId }));
+
+  // `windows.get` rejects both for an id that never existed (a buggy or hostile
+  // dispatcher) and for one already removed — the window closed during the
+  // round trip, so `onRemoved` ran while this request still had no window,
+  // found no candidates, and nothing else will ever cancel it. Both cases are
+  // repaired identically: run exactly what `onRemoved` would have run. When the
+  // request still has another live window (the Ledger pair), the detach inside
+  // simply shrinks the set and cancels nothing.
```
Comp0te: This liveness probe asks only "does a window with this id exist". `windows.get(windowId)` is called with no `{ populate: true }`, no check that the window belongs to this extension, and no check that it displays this `requestId` — so any live browser window id passes.

The file's header comment says it exists to prevent a permanently uncancellable request; as written it prevents only the dead-id half. With `windowIds: [W, F]` where `F` is an unrelated live window, closing the real approval window `W` is filtered out at `cancel-requests.ts:137-140` (length 2), leaving `[F]` — and nothing will ever fire for `F`. The request stays `'open'` and its dapp's promise hangs.

The tests cover the other cases but not this one: `attach-window-to-request.test.ts:45-72` covers dead and never-existent ids, `:92-112` covers malformed shapes, and no case supplies a live-but-foreign id.

Suggest `windows.get(windowId, { populate: true })` and requiring the first tab's URL to start with `runtime.getURL('')` and carry the same `requestId` search param; otherwise run the same repair as the reject arm.

src/background/handlers/attach-window-to-request.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  store.dispatch(windowRequestWindowAttached({ requestId, windowId }));
+
+  // `windows.get` rejects both for an id that never existed (a buggy or hostile
+  // dispatcher) and for one already removed — the window closed during the
+  // round trip, so `onRemoved` ran while this request still had no window,
+  // found no candidates, and nothing else will ever cancel it. Both cases are
+  // repaired identically: run exactly what `onRemoved` would have run. When the
+  // request still has another live window (the Ledger pair), the detach inside
```
Comp0te: Separate from the "any live window passes" point on the next line: this `.catch` does not bind its error and treats *any* rejection as "the window is gone", running the full repair — `cancelRequestsDisplacedBy(store, windowId, 'cancel-on-close')`.

The comment above asserts that `windows.get` rejects only for a never-existent or already-removed id, but nothing here narrows on the error to check that. If Firefox's or Safari's `windows.get` rejects for any unrelated reason — a transient extension-context error, a WKWebExtension window-type quirk — this cancels a live, user-facing approval and tells the dapp "cancelled" while the window is still on screen.

There is a second cost: the comment names "a buggy or hostile dispatcher" as one of the two causes, and that signal is currently unrecoverable from the code — there is no way to tell it from the benign race.

Suggest binding the error and logging it with `{requestId, windowId}` before running the repair, and narrowing the repair to rejections that actually mean "no such window".

src/background/handlers/redux-actions.ts
```diff
@@ -198,6 +200,20 @@ export async function handleReduxAction(
   action: { type: string },
```
Comp0te, line 199: This signature is unchanged, but the PR is what makes its missing `sender` matter: attaching a window is now what makes a request cancellable, so the new `windowRequestWindowAttached` branch at `:208-215` puts a lifecycle-authority decision on the one handler path with no sender gate.

The asymmetry is deliberate elsewhere in the router. `index.ts:263-268` calls `handleReduxAction` inside the generic loop with no check, while `handleSdkResponseToTab` (`:272-276`, gated at `sdk-response-to-tab.ts:78`) and `handleLegacyImport` (`:283`, gated at `legacy-import.ts:36-38`) are both pulled *out* of that loop specifically to gate on `sender`, each with a comment naming a content-script-world compromise as the reason. `@bringweb3/chrome-extension-kit` runs in exactly that world (`content/index.ts:3,220`).

Both directions are reachable from there. A live-but-unrelated `windowId` gives a set that never shrinks to empty (permanently uncancellable — see the comment on `attach-window-to-request.ts:50`); a dead `windowId` attached in the gap before the real window attaches makes `windowIds` exactly `[dead]`, which `cancel-requests.ts:137-140` then selects and `:79-91` cancels after the grace on status alone.

Not dapp-reachable today — the page-world relay pins `SDK_REQUEST_TYPES` at `content/index.ts:20-33,127-132` — but that allowlist is precisely the defence-in-depth the sibling handlers' comments say they do not rely on.

Suggest threading `sender` into `handleReduxAction` and requiring `isTrustedUiSender(sender)` at least for this branch, ideally for the whole forwarding set.

src/background/handlers/redux-actions.ts
```diff
@@ -198,6 +200,20 @@ export async function handleReduxAction(
...
   store: MainStore
 ): Promise<HandlerResult> {
+  // Intercepted rather than forwarded blindly: attaching a window is what makes
+  // a request cancellable, so a dead or bogus `windowId` would leave it open
+  // forever. `attachWindowToRequest` dispatches AND verifies the window is
+  // alive, which the generic forwarding path cannot do. The UI dispatcher here
+  // is `use-ledger` registering the Ledger permission window.
+  if (windowRequestWindowAttached.match(action)) {
+    attachWindowToRequest(
```
Comp0te, line 229: Small one, on the path the interception was added to harden: `windowRequestWindowAttached.match(action)` is RTK's `isAction(action) && action.type === type` — it does not validate `payload`. So `action.payload.requestId` on the next lines is read *before* `attachWindowToRequest`'s own shape guard runs, and a payload-less message throws a `TypeError` out of `handleReduxAction`, which the router turns into a generic `sendError` (`index.ts:299-301`).

The comment above says this payload "is not trusted to be well-formed", so the guard sitting one function too deep to cover the most obvious malformed shape seems worth closing. Reading via `action.payload?.requestId` / `?.windowId` lets the existing guard reject it and produce the intended `attachWindowToRequest: ignoring malformed attach` log instead.

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -88,6 +88,14 @@ export async function handleSdkResponseToTab(
     requestId != null &&
     selectRequestStatus(store.getState(), requestId) === 'responded'
   ) {
+    // A dropped `cancelled: true` is benign; a dropped signature never is, and
+    // the two are distinguishable here. Log the identifiers ONLY — `action`
```
Comp0te, line 92: The comment here says "A dropped `cancelled: true` is benign; a dropped signature never is, and the two are distinguishable here" — and then the code does not distinguish them. Both take the identical `console.error` with `{requestId, tabId, type}` and both return `{handled: true, response: undefined}` at `:102`. The payload is never read.

Two consequences. The benign case will be the overwhelming majority of these lines, at `console.error` severity, which trains a reader to ignore the one line that means a user's signed transaction was destroyed. And a dropped genuine signature still produces no user-visible feedback at all — no `sagaError`, no banner. P1 #5 asked for a log and got one; the surfacing half is still missing on the path that matters.

Two things worth knowing before implementing the split, since they make it less mechanical than it looks:

- The responses that reach this line are always UI-originated, from `sendSdkResponseToSpecificTab`. The cancel payloads built by `buildCancelResponse` go out directly via `tabs.sendMessage` at `cancel-requests.ts:98` and `:174` and never pass through here — so the discriminator is "a UI-delivered result vs a UI-delivered cancel", not the cancel-path marker.
- `payload.cancelled` only exists on some shapes. `sdk-method.ts:29-38` types `connectResponse` and `switchAccountResponse` payloads as a bare `boolean`, so a branch on `action.payload.cancelled` will not generalise across the union as written.

Suggest branching on whether the dropped response carries a real result, keeping `console.error` (identifiers only) plus a `sagaError` for that case, and downgrading the cancel-duplicate to `console.warn`.

src/background/redux/types.d.ts
```diff
@@ -22,13 +22,12 @@ export type PopupState = {
...
-  windowManagement: Omit<
-    WindowManagementState,
-    'pendingRequests' | 'exportKeysWindowId' | 'requests'
-  >;
+  // Only `windowId` is broadcast. The `requests` map holds each in-flight
+  // request's dapp origin, tabId and window ids; no UI replica reads it, and
+  // broadcasting it would leak dapp origins into every popup update. `Pick`
+  // (not `Omit`) so a newly added slice field is excluded until someone opts
+  // it in.
+  windowManagement: Pick<WindowManagementState, 'windowId'>;
```
Comp0te, line 30: `Pick<WindowManagementState, 'windowId'>` is structurally identical to the `Omit<…>` it replaces, and does not close the spread bypass WALLET-1364 #12 reported. Excess-property checking is suppressed for spread properties regardless of which of the two the target was written with.

Checkable in isolation with the repo's own tsc: `const a: Pick<S,'windowId'> = {...full}` produces no error, while `const b: Pick<S,'windowId'> = {windowId, requests}` produces TS2353 — and `Pick`↔`Omit` are mutually assignable here. So re-introducing `windowManagement: {...state.windowManagement}` still compiles clean and still re-broadcasts every in-flight dapp origin, tabId and `windowIds` into every replica on every store change.

The code comment states the real benefit correctly — future slice fields are excluded by default, which is genuine and worth having. It is the ticket's framing of `Pick` as the fix for the spread bypass that does not hold. With `selectPopupState`/`createMainStoreReplica` also having no tests, that regression currently has no enforcement at compile time or runtime.

Suggest keeping `Pick` and adding the enforcement it cannot provide: export `selectPopupState` and assert `Object.keys(popupState.windowManagement)` equals `['windowId']`, or build the slice through a named helper with an annotated return type — an explicit object literal *is* excess-property checked.

src/background/create-open-window.ts
```diff
@@ -140,7 +140,11 @@ export function createOpenWindow({
               });
 
         return newWindow.then(newWindow => {
-          if (newWindow.id) {
+          // `isNewWindow` opens a deliberately SEPARATE window (import-account
+          // flows). Tracking it would retarget the shared approval slot: a
+          // later close of that separate window would clear the tracked id
+          // and a dapp approval could be cancelled by the wrong event.
+          if (newWindow.id && !isNewWindow) {
```
Comp0te, line 154: `if (newWindow.id && !isNewWindow)` — the `!isNewWindow` clause is new (base was `if (newWindow.id)`), and it makes three of `useWindowManager`'s inputs dead.

Both consumers of `useWindowManager().openWindow` pass `isNewWindow: true` (`account-list.tsx:121-124`, `navigation-menu/index.tsx:175-178`). That also forces `id = null` at `:70`, which skips the reuse block at `:72-77` — so `clearWindowId` at `:107` is unreachable too, and the `useSelector(selectWindowId)` at `use-window-manager.ts:13` feeds nothing. `knip` will not flag any of it, since these are object properties rather than exports.

The part with a live edge is downstream: `windowIdChanged` and `windowIdCleared` remain in `FORWARDED_ACTION_TYPES` (`redux-actions.ts:151-152`) with no legitimate UI dispatcher left — the only ones were `use-window-manager.ts:19-20`, and the background dispatches its own at `open-window.ts:49-50`. Any extension UI page can still `runtime.sendMessage` a `windowIdChanged(<arbitrary id>)` and retarget the shared approval-window slot. The parity test reasons carefully about exactly this hazard for `windowRequestWindowAttached`; these two now meet the same background-only criterion.

Suggest dropping the dead inputs from `useWindowManager`, and moving `windowIdChanged`/`windowIdCleared` into `EXCLUSIONS` with the reason recorded.

src/background/redux/windowManagement/reducer.ts
```diff
@@ -39,44 +41,94 @@ const slice = createSlice({
...
+              id => id !== action.payload.windowId
+            )
+          };
+          changed = true;
+        }
+      }
+
+      return changed ? { ...state, requests } : state;
+    },
+    // The tombstone is deliberately kept: `selectRequestStatus` reading back
```
Comp0te, line 128: This comment claims the map "stays proportional to in-flight requests, not to the lifetime total" and that "an MV3 service-worker restart wipes it". Neither holds on all three build targets.

`windowRequestResponded` writes `{status: 'responded'}` and nothing anywhere deletes a key — the descriptor is dropped, the entry is not. And `src/manifest.v2.json:45` and `src/manifest.v2.safari.json:41` both declare `"persistent": true`, so the Firefox and Safari builds have a background page that is never torn down.

On those two targets `requests` grows by one permanent entry per request for the whole browser session, keyed by a dapp-supplied string. `Record<string, Request>` says nothing about lifetime, so this comment is the only statement of the bound — and it is also the sole justification given for keeping the tombstone at all.

Suggest either correcting it to state the real bound (unbounded on MV2, service-worker-lifetime on MV3), or adding eviction — e.g. FIFO-capping the `responded` entries.

src/background/redux/windowManagement/reducer.ts
```diff
@@ -39,44 +41,94 @@ const slice = createSlice({
...
-      },
-      pendingRequests: {
-        ...state.pendingRequests,
-        [action.payload.requestId]: {
-          tabId: action.payload.tabId,
-          origin: action.payload.origin,
-          method: action.payload.method
-        }
+    ) => {
+      if (state.requests[action.payload.requestId] != null) {
```
Comp0te: This no-op is new (the base reducer overwrote unconditionally), but `sdk-methods.ts:170-194` and its five sibling branches still call `openWindow` afterwards regardless of the outcome — so the reducer declining the request and the window opening anyway are not connected.

Two shapes follow. A dapp replaying a finished `requestId` hits the tombstone: `windowRequestWindowAttached` is also rejected for it (`:72-78`, `status !== 'open'`), the window still opens fully, `deployPayloadReceived` still populates `jsonById` so the screen renders normally — and the user's approval is then silently dropped at `sdk-response-to-tab.ts:87-103`. A dapp reusing a *live* `requestId` across methods keeps the first descriptor, so `buildCancelResponse` later builds the wrong response type (e.g. `connectResponse(false)` delivered against a pending `sign`).

Self-inflicted only, since a v4 UUID from another origin is unguessable and no path leaks one. The comment at `open-window.ts:68-71` acknowledges the case, but it covers the cancellation, not the window opening — and the residual is that the wallet renders a fully functional approval screen for a request it will never answer.

Suggest making the reducer's no-op visible to the caller (or checking `selectRequestStatus` in `sdk-methods.ts`) and responding an error to the dapp immediately instead of opening a window.

src/background/redux/windowManagement/reducer.ts
```diff
@@ -39,44 +41,94 @@ const slice = createSlice({
...
+    // `windows.create`/reuse resolves, and by `use-ledger` for the separate
+    // permission window.
+    windowRequestWindowAttached: (
       state,
-      action: PayloadAction<{ requestId: string }>
+      action: PayloadAction<{ requestId: string; windowId: number }>
     ) => {
-      const pendingRequests = { ...state.pendingRequests };
-      delete pendingRequests[action.payload.requestId];
+      const request = state.requests[action.payload.requestId];
```
Comp0te: Different angle on the two guards in this reducer: the type system believes both are unreachable.

`tsconfig.json` sets `"strict": true` without `noUncheckedIndexedAccess`, and `requests` is `Record<string, Request>` (`types.ts:33`) — so `state.requests[id]` is typed `Request`, never `Request | undefined`. That makes `request == null` on the next line dead code as far as the compiler is concerned, and the same is true of the `!= null` guard at `:45`.

That second one is the P2 #11 fix, i.e. the most important new invariant in the slice — "a repeated dapp-controlled `requestId` must not overwrite a live request nor resurrect a tombstone" — expressed as a runtime check the compiler thinks can never fire. Enabling `@typescript-eslint/no-unnecessary-condition`, or any refactor by someone who trusts the declared type, deletes it. The reducer tests cover the behaviour, but tests do not stop a guard being removed as provably redundant.

Suggest modelling the lookup honestly rather than flipping a repo-wide flag — `requests: Partial<Record<string, Request>>`, or a `getRequest(state, id): Request | undefined` accessor used by both reducer and selectors.

src/background/redux/windowManagement/types.ts
```diff
@@ -8,15 +8,27 @@ export type CancellableMethod =
...
-  tabId: number;
-  origin: string;
-  method: CancellableMethod;
-}
+// One request, one entry. `windowIds` is every window currently DISPLAYING this
+// request — normally one approval window, but the Ledger permission flow opens a
+// second window carrying the same requestId (`src/hooks/use-ledger.ts`), so a
+// request can outlive the loss of any single window. Cancellation is driven by
+// "this window displays it no longer", never by a timer.
+export type Request =
```
Comp0te, line 25: The `'open'` variant permits `windowIds: []`, which is simultaneously the legitimate transient state (registered, window not yet resolved) and the terminal stranded one (nothing will ever display this).

The consequence is that the stranded case has no exit. `cancelRequestsDisplacedBy` filters on `windowIds.length === 1` (`cancel-requests.ts:137-140`), so an empty set can never be cancelled by any window event; the only escape is `failRequestOnWindowError`, and grepping it at this head shows it is reachable only from `openWindow`'s two failure branches (`open-window.ts:57` and `:97`, via `failIncomingRequest`). There is no reaper — once a request reaches `{status: 'open', windowIds: []}` outside `openWindow`'s control, it stays there for the life of the background context.

This union was introduced precisely to make illegal states unrepresentable, and this is the one illegal-ish state it still represents. It is also where two other paths flagged in this review land.

Suggest encoding the displayed case as non-empty — `windowIds: readonly [number, ...number[]]` on a `'displayed'` member, with a separate `{status: 'awaiting-window'}` for the gap between `windowRequestOpened` and the first attach — so the stranded case becomes visible to `selectOpenRequests` consumers instead of indistinguishable from a healthy one.

src/background/redux/windowManagement/types.ts
```diff
@@ -8,15 +8,27 @@ export type CancellableMethod =
...
+  | {
+      status: 'open';
+      tabId: number;
+      origin: string;
+      method: CancellableMethod;
+      windowIds: number[];
```
Comp0te: Narrower point than the empty-set one above: `windowIds: number[]` is mutable, and `selectOpenRequests` hands the store's own array out by reference — `selectors.ts:30` is `{ requestId, ...request }`, a shallow spread. `configureStore` is created with `immutableCheck: false` (`redux/index.ts:24-29`), so nothing catches an in-place mutation at runtime either.

`cancelRequestsDisplacedBy` decides cancellability purely from `windowIds.length` and `windowIds[0]` (`cancel-requests.ts:137-140`), so any consumer doing `request.windowIds.push(id)` on a selector result silently rewrites that decision with no action dispatched.

The reducer itself is correctly copy-on-write (`reducer.ts:86,106-108`) — the type simply does not say that is required. Suggest `readonly windowIds: readonly number[]`, and `readonly` on the other `Request` fields while you are there, since nothing mutates them in place.

src/background/redux/windowManagement/selectors.ts
```diff
@@ -13,22 +13,19 @@ export const selectExportKeysWindowId = (state: RootState): number | null =>
...
-  (requests, pendingRequests) =>
-    Object.entries(requests)
-      .filter(([, status]) => status === 'open')
-      .map(([requestId]) => ({ requestId, ...pendingRequests[requestId] }))
+// Every request still awaiting a response, with its descriptor. Memoized
+// (reselect) so the repeated reads on the cancel path — once to pick candidates,
+// once after the grace — do not rebuild the array while the slice is unchanged.
+export const selectOpenRequests = createSelector(selectRequests, requests =>
+  Object.entries(requests)
+    .filter(
```
Comp0te: This is the single boundary where descriptors are extracted for the cancel path, and it opts out of checked narrowing: `.filter((entry): entry is [string, OpenRequestDescriptor] => entry[1].status === 'open')`.

TypeScript verifies only that the asserted type is assignable to the parameter type; it does not verify the predicate body. Writing `=== 'responded'` here compiles identically, and everything downstream would still be typed as open descriptors.

The rest of the slice does get this for free — `reducer.ts:72-78` and `:99-112` narrow on the discriminant and are compiler-checked. Getting the narrowing from the discriminant here too closes the gap: `Object.entries(requests).flatMap(([requestId, request]) => request.status === 'open' ? [{ requestId, ...request }] : [])`.

src/hooks/use-ledger.ts
```diff
@@ -182,6 +182,18 @@ export const useLedger = ({
 
         if (w.id) {
           dispatchToMainStore(ledgerNewWindowIdChanged(w.id));
+
+          // This window displays the SAME requestId as the approval window that
+          // opened it. Registering it is what keeps the request alive when the
+          // shared approval window is reused or closed while the user is still
+          // confirming on the device.
+          const requestId = askPermissionUrlData.params?.requestId;
```
Comp0te: Distinct from the error-handling point on the line below: this `requestId` read — the only thing in the codebase that ever puts a second entry in `windowIds`, i.e. the UI half of the P0 fix — has no test at all.

At this head, grepping `use-ledger`/`useLedger` across `src` returns 13 files and none of them is a test that imports or exercises the hook; grepping `windowRequestWindowAttached` returns 11, of which the only non-background hit is this file. Deleting lines 190-195 leaves the suite green.

There is a cheaper way to break it than deletion, which is what makes the gap worth closing: rename the `requestId` key at `sign-transaction/index.tsx:294-295` or `sign-message/index.tsx:189-190`. `askPermissionUrlData.params` is `Record<string, string>` (`:39`), so the lookup here silently yields `undefined`, `tsc` does not catch it, and the permission window is never registered. The background half is well covered (`cancel-requests-displaced-by.test.ts:63-80`, `reducer.test.ts:587`); nothing proves the Ledger flow ever produces the second entry.

The repo has no React-hook harness, so suggest either extracting the `requestId → dispatch` decision into a testable unit, or asserting at the page level that `params.requestId` is threaded. Typing `params` as `{ requestId?: string; … }` would additionally make the rename case a compile error.

ost-ptk:
Thanks for the depth here — this is one review rather than 26, so one reply.

I checked every comment against the branch before touching anything. **None was a false positive**, including the claims that needed independent verification: `"persistent": true` really is in both MV2 manifests, `SagaError.source` really was still `string`, both `useWindowManager` consumers really do pass `isNewWindow: true`, `connectResponse`'s payload really is a bare `boolean`, the store subscriber really does no state-change comparison, and `export-keys-window-saga.ts` / `webpack.config.js` really are absent from the diff.

**24 of the 26 are fixed here** (15 commits, `8e0e1747..7368e935`), 2 are on the stacked #1430 branch because they touch files that PR already changed, 1 was already closed by #1430, and 2 are taken partially on purpose.

## The one that mattered

`cancel-requests.ts:91` was right, and it is the reason this PR should not have merged as it stood. The re-check after the grace consulted status alone, so a request that _regained_ a window inside those 250 ms was still cancelled — this PR's own P0, reached from inside the fix. Your repro drove it exactly: `getState` returning `windowIds: [7]` at snapshot time and `[9]` after it produces one `tabs.sendMessage` and one `windowRequestResponded` for a request holding a live window. That is now the first commit, and the re-check runs against the current descriptor.

## Resolution map

| Comment                                                                                                                                         | Resolution                                                                                                                                           |
| ----------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
| `cancel-requests.ts:91` — grace re-check ignores `windowIds`                                                                                    | f9210395                                                                                                                                             |
| `attach-window-to-request.ts:49` — unbound `.catch`, any rejection repairs<br>`redux-actions.ts:209` — `.match()` does not validate the payload | 291d2a99                                                                                                                                             |
| `open-window.ts:51` — throw in the success arm caught by nothing                                                                                | c5fa6dee                                                                                                                                             |
| `cancel-requests.ts:144` — unconditional detach dispatch                                                                                        | 548e7c0f                                                                                                                                             |
| `cancel-requests.ts:107` — banner suppressed on the close path too<br>`cancel-requests.ts:172` — `deliverViaOrigin` result discarded            | 8c9f9e71                                                                                                                                             |
| `use-ledger.ts:191` — three silent skips<br>`use-ledger.ts:190` — the UI half has no test                                                       | 480ae3d3                                                                                                                                             |
| `index.ts:185` — IIFE without `.catch`<br>`index.ts:188` — the listener is untested                                                             | 7c0bd959                                                                                                                                             |
| `redux-actions.ts:200` — no `sender` gate                                                                                                       | 6770b213                                                                                                                                             |
| `attach-window-to-request.ts:50` — any live window passes                                                                                       | 9ebf8f77                                                                                                                                             |
| `types.ts:22` — mutable `windowIds`<br>`selectors.ts:26` — unchecked type predicate<br>`reducer.ts:70` — guards are dead code to the compiler   | f890ec85                                                                                                                                             |
| `cancel-requests.ts:27` — `CancelSource` constrains the wrong thing                                                                             | 1549b0d3                                                                                                                                             |
| `reducer.ts:116` — the map's stated bound is wrong on MV2                                                                                       | 0234e7c3                                                                                                                                             |
| `reducer.ts:45` — reducer no-op invisible to the caller<br>`open-window.ts:17` — `requestId` optional with no such caller                       | 2f8751fb                                                                                                                                             |
| `create-open-window.ts:147` — dead `useWindowManager` inputs                                                                                    | 47ef457f                                                                                                                                             |
| `open-window.ts:91` — `.name` drops the diagnosis                                                                                               | 7368e935                                                                                                                                             |
| `sdk-response-to-tab.ts:92` — the two cases are not distinguished                                                                               | baf6d071 _(on #1430)_                                                                                                                                |
| `cancel-requests.ts:125` — "the single cancellation trigger"                                                                                    | e4417c56 _(on #1430)_                                                                                                                                |
| `types.d.ts:30` — `Pick` does not close the spread bypass                                                                                       | already closed by #1430's `4b4c0929` — the broadcast-shape test asserts the exact key set and that no dapp origin reaches a replica. Nothing to add. |
| review summary — the body claims P1 #3 and #4                                                                                                   | PR description corrected; those two ship in #1428 / #1429.                                                                                           |

## Where I did not take the suggestion as written

**`types.ts:16` — `windowIds: []` means two different things.** Agreed on the diagnosis. But `'awaiting-window'` + `readonly [number, ...number[]]` is a refactor through the reducer, selectors, both cancel paths, fixtures and every slice test, and the actual harm — a request stranded with nothing that can ever cancel it — has identifiable entry paths. Those are closed instead: `requestId` is required, the post-open `.catch` exists, and a duplicate `requestId` no longer opens a window. The stranded state is now reachable only under `openWindow`'s control, whose two failure branches both lead to `failRequestOnWindowError`. The union itself wants its own ticket, and I have not filed it yet.

**`redux-actions.ts:200` — gate the whole forwarding set.** Gated the `windowRequestWindowAttached` branch here. Doing it for all ~60 forwardable actions is a behaviour change that needs its own verification that every UI dispatcher passes `isTrustedUiSender`, so it is also a separate ticket rather than a rider on this one.

## Two implementation notes worth flagging back

**`attach-window-to-request.ts:50` — I did not add the `requestId` match as a verdict.** The ownership check is in (`{ populate: true }` plus an extension-origin prefix), but on the reuse path `tabs.update` resolves when the navigation _starts_, so a legitimate window can be probed while `tab.url` is still the previous request's. A false-positive repair there cancels a live approval — the exact failure this model exists to prevent — so the checks are deliberately asymmetric: a URL that is provably not ours is a verdict, a `requestId` mismatch is a `console.warn`, and anything inconclusive leaves the attach standing.

**`attach-window-to-request.ts:49` — I did not narrow on the error's text.** "Narrow the repair to rejections that actually mean _no such window_" is right, but the only handle is the message, and the wording differs per browser — the same trap #1430's own description flags for `broadcastToReplicas`. It confirms against `windows.getAll()` instead, and fails closed if even that is unavailable: an uncancellable request hangs until its own timeout, which is recoverable, while a wrongly cancelled one destroys a signature the user already approved.

## Verification

`npm run ci-check` is green on both branches (66 suites / 473 tests here, 67 / 488 on #1430). Four mutation probes confirm the new tests are load-bearing, and two of the type changes are now compile-enforced: flipping the discriminant in `selectOpenRequests`, or deleting the reducer's null guard, are both errors now and neither was before.

One bug was caught by its own test while implementing: the tombstone eviction used `slice(0, length - CAP)`, and a negative end slices from the _back_, so it evicted almost everything until the map was full. Worth mentioning as evidence that these tests do something.

The red CI on this PR was unrelated — two advisories published after the branch last ran green (`GHSA-rgw5-rvv9-x895`, `GHSA-7p8r-x3mc-p8w7`). Both now have patched releases inside the existing `overrides` ranges, so `d923973c` bumps them and retires the `GHSA-mh99-v99m-4gvg` allowlist entry, whose "no patched 1.x exists" reasoning the 1.1.18 maintenance release makes false.

src/background/handlers/attach-window-to-request.ts
Comp0te: The "any live browser window passes" half is closed — `{ populate: true }` plus the extension-origin prefix check, and the `tabs` permission is present in all three manifests, so the URL really is readable. Your reason for the asymmetry checks out too: `create-open-window.ts:97,107` fetches the window before `tabs.update`, which resolves at navigation start.

What remains is narrower than the original comment: a `requestId` mismatch is `console.warn` only at `:86-91`, and a URL carrying no `requestId` param produces no signal at all — so the check establishes "this is one of our windows", not "this is the window showing this request". Worth deciding whether that gap is intentional, since the mismatch branch is the only thing that would distinguish them.

src/background/handlers/sdk-methods.ts
```diff
@@ -35,14 +36,42 @@ import { SdkMethod, sdkMethod } from '@content/sdk-method';
...
+  // caller: every branch below went on to open a window anyway. The result was
+  // a fully functional approval screen for a request the wallet could never
+  // answer — a replayed id renders normally, the user approves, and the
+  // response is dropped by the dedup with the dapp none the wiser. Reusing a
+  // LIVE id under another method is worse still: the first descriptor is kept,
+  // so a later cancel is built in the wrong shape. Refuse both here, before
+  // anything is dispatched.
+  if (
+    APPROVAL_REQUEST_TYPES.has(action.type) &&
+    selectRequestStatus(store.getState(), action.meta.requestId) != null
```
Comp0te: This guard and the reducer's "register once" guard read the same dapp-controlled key off a plain object with different operators. `selectRequestStatus` is `requests[requestId]?.status`; the reducer at `windowManagement/reducer.ts:51` is `state.requests[requestId] != null`.

For `requestId` ∈ `__proto__`, `toString`, `constructor`, `valueOf`, `hasOwnProperty` the two disagree. The first sees `undefined` and lets the request through; the second sees an inherited `Object.prototype` member, concludes the id is already registered, and returns `state` unchanged — so nothing is registered. `handleSdkMethod` then calls `openWindow` anyway, `attachWindowToRequest`'s dispatch is likewise a no-op, and `failRequestOnWindowError` returns early because there is no descriptor.

The result is a live `sign` / `signMessage` / `connect` / `decryptMessage` approval window sitting outside the entire lifecycle model this PR builds — not cancellable on close, not cancellable on supersede, not deduped, not recoverable on window-open failure. That is the stranded state the change exists to eliminate, reachable by picking one of five string literals, with no race involved. `isSDKMethod` (`src/content/sdk-method.ts:129`) only checks `typeof requestId === 'string'`, so nothing upstream rejects them.

The divergence on its own: `node -e "const m = {}; console.log(m['__proto__']?.status, m['__proto__'] != null)"` → `undefined true`.

Suggested: validate `action.meta.requestId` against a UUID shape at the `handleSdkMethod` entry, and build the map with `Object.create(null)` or read it via `Object.prototype.hasOwnProperty.call`.

src/background/handlers/attach-window-to-request.ts
```diff
@@ -0,0 +1,126 @@
...
+    );
+  };
+
+  // Two independent things can be wrong with `windowId`, and they are checked
+  // in the two arms below. Note the asymmetry: a provable verdict repairs,
+  // anything inconclusive does not. On the reuse path `tabs.update` resolves
+  // when the navigation STARTS, so a legitimate window can be probed before its
+  // URL settles — repairing on that would cancel a live approval, the exact
+  // failure this model exists to prevent.
+  void windows.get(windowId, { populate: true }).then(
```
Comp0te: This is `void windows.get(windowId, { populate: true }).then(onFulfilled, onRejected)` — a two-arm `.then` with no trailing `.catch`. A throw inside the fulfilled arm becomes an unhandled rejection.

`runtime.getURL('')` at `:74` is a live candidate: it is precisely the API that fails on an invalidated extension context, which the comment a few lines below cites as the reason the *rejected* arm must not trust itself.

This is the same shape you repaired one file over in this push — `open-window.ts:132`, `void chain.catch(...)`. The inner `windows.getAll()` chain here has a `.catch`; the outer one does not. Nothing logs, nothing repairs, and the attach stands with a `windowId` whose liveness was never established — this file's own header calls that the permanently-uncancellable state. There is no `unhandledrejection` handler anywhere in `src/`, so in an MV3 service worker it leaves no trace at all.

src/background/handlers/cancel-requests.ts
```diff
@@ -71,37 +88,151 @@ export async function cancelRequests(
...
+
         store.dispatch(
           sagaError({
             source,
             message:
-              `Cancel delivery to tab ${tabId} failed` +
-              (delivered > 0
-                ? '; delivered via same-origin fallback'
-                : '; not delivered')
+              delivered > 0
```
Comp0te, line 150: The `delivered === 0` arms of both ternaries are covered — `fail-request-on-window-error.test.ts:80-100` asserts `stringContaining('could not be told')` and `cancel-requests-displaced-by.test.ts:161-182` asserts `stringContaining('not delivered')`, both with `deliverViaOrigin` mocked to zero.

The other arm has no assertion anywhere. `'…recovered via the page'` here at `:151` and `'…the request was cancelled'` at `:234` appear nowhere outside the source; the only other hit is a test *title* at `cancel-requests-displaced-by.test.ts:184`, which pins suppression rather than text.

That gap matters because `saga-error-banner.tsx:87` renders `{error.source}: {error.message}` verbatim, and the two arms say opposite things about whether the dapp was told. Collapsing either ternary to its pessimistic constant leaves the whole suite green and makes the banner tell the user the site may still be waiting when it was in fact informed — the direction these commits set out to fix.

Suggested: add a `delivered > 0` case to each, asserting `stringContaining('recovered via the page')` and `stringContaining('the request was cancelled')`.

Comp0te (commented):
Second pass over the push that answers the first round — the incremental diff `8e0e1747..d923973c` plus the state it leaves behind. Most of the earlier comments are closed and I've resolved those threads.

What's left is two ways a request can reach the lifecycle model with nothing registered in it, two new failure paths that log but leave the user or the request stranded, and one banner arm that no test pins.

src/hooks/use-ledger.ts
```diff
@@ -180,20 +207,50 @@ export const useLedger = ({
       ) {
         const w = await openNewSeparateWindow({ url });
 
-        if (w.id) {
-          dispatchToMainStore(ledgerNewWindowIdChanged(w.id));
-          triggeredRef.current = true;
+        if (w.id == null) {
```
Comp0te, line 210: Both new failure paths — this `w.id == null` early return and the effect's `.catch` at `:237-246` — log and return without calling `setLedgerEventStatusToRender` or setting `triggeredRef` (set only at `:228`, on the success path). For the `openNewSeparateWindow` rejection the comment at `:238-239` names, nothing in the dependency array at `:247-253` has changed at that point, so the effect will not re-run and the status stays `LedgerPermissionRequired`.

To be fair to what's already there: that status *is* in the ledger-error map (`errors.ts:22-26`), so the page does render an error view with a working "Got it" CTA (`ledger-footer.tsx:44-48`). The problem is that the error is the wrong one. It tells the user to grant permission in a window that was never opened and never will be, with no signal that the open failed and no retry path — the only exit is the CTA that abandons the approval.

Suggested: set a distinct rendered error status in both branches so the screen reflects "the permission window could not be opened" rather than "waiting for permission".

src/background/redux/windowManagement/reducer.ts
```diff
@@ -39,43 +47,115 @@ const slice = createSlice({
...
+    // service-worker restart wipes it (after which a late duplicate is no
+    // longer deduped), while `manifest.v2.json` and `manifest.v2.safari.json`
+    // both declare `"persistent": true`, so on Firefox and Safari the
+    // background page is never torn down. Nothing here deleted a key, so on
+    // those two targets the map grew by one permanent entry per request —
+    // keyed by a dapp-supplied string — for the whole browser session. Hence
+    // the FIFO cap below: the descriptor is dropped as before, and the oldest
+    // tombstones are evicted once there are more than a dedup could plausibly
+    // need. Open requests are never evicted.
+    windowRequestResponded: (
```
Comp0te, line 140: `windowRequestOpened` (`:51-53`) and `windowRequestWindowAttached` (`:78-84`) both refuse to act on a missing or wrong-status entry. `windowRequestResponded` here is an unguarded upsert, so it writes a tombstone for a `requestId` that has no descriptor at all — reachable from `sdk-response-to-tab.ts:117,134` for any id the extension UI forwards.

The union models `∅ → open → responded`; the reducer permits `∅ → responded`. A concrete route: an MV3 service-worker restart wipes `requests` between registration and the UI's response, and the response then tombstones an id the store has never seen. That entry consumes a slot in the new 50-entry cap, and until it is evicted the `Duplicate requestId` throw at `sdk-methods.ts:68-73` will reject that id — `selectRequestStatus` returns `'open' | 'responded' | undefined`, so a tombstone trips it just as a live entry does.

Bounded, to be clear: the FIFO cap at `:147-156` evicts it after 50 further responses, and `generateRequestId` is `crypto.randomUUID`, so this is not a collision path. The point is the asymmetry itself — two of the three cases make the illegal transition unrepresentable and the third does not.

Suggested: guard the upsert the way its two siblings are guarded — only tombstone a `requestId` that is currently `'open'`.

ost-ptk:
Second round: all five actionable comments fixed, and an answer to the question on `attach-window-to-request.ts:50`. `d923973c..15f8311f`, `ci-check` green (486 tests here, 501 on #1430).

## `sdk-methods.ts:70` — the prototype-key divergence

Reproduced exactly as described, and it is worse in one place than the comment says.

```
selectRequestStatus sees : undefined
reducer guard sees       : true
```

for all five of `__proto__`, `toString`, `constructor`, `valueOf`, `hasOwnProperty`. Fixed by routing every read of the map through a single `getRequest` that uses `hasOwnProperty`, so the two questions can no longer be answered differently — `c9feb3a5`.

But four of the five and `__proto__` are not the same problem. Once reads agree, `toString`/`constructor`/`valueOf`/`hasOwnProperty` store as ordinary own properties and behave like any other id; I have tests for registration, second-registration refusal and window attach on them. `__proto__` cannot be stored at all: building the next map copies entries by assignment, and assigning `__proto__` sets the object's _prototype_ instead of adding an entry, so the descriptor would be inherited by every later lookup in the map. Measured:

```
__proto__        own: false | keys: [] | protoChanged: true
toString         own: true  | keys: ["toString"] | protoChanged: false
```

So `__proto__` is refused rather than stored — in the reducer, and again at the SDK entry with `Invalid requestId` so the dapp gets a definite answer instead of a window nothing can close.

**I did not add the UUID-shape validation.** With the reads agreed and `__proto__` refused, it buys nothing here, and it can reject legitimate traffic: it would be the wallet, not the SDK, deciding the id format, and a dapp on an older bundled SDK has no recourse. If you think the entry should own that constraint anyway, I will take it — but I would rather it be a deliberate compatibility decision than a side effect of this fix.

## The rest

| Comment                                                                      | Fix                                                                                                                                                                                                                                                                                                   |
| ---------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `attach-window-to-request.ts:60` — two-arm `.then` with no trailing `.catch` | `063737e4`. Correct, and I had just repaired the same shape one file over — the inner `getAll` chain had a `.catch` and the outer one did not. `runtime.getURL` is indeed the live candidate.                                                                                                         |
| `reducer.ts:136` — unguarded tombstone upsert                                | `4984ca04`. Guarded on `'open'` like its two siblings, so ∅ → responded is no longer representable. The eviction tests now drive the real ∅ → open → responded path rather than tombstoning ids that were never registered — which is also what let the old shape through unnoticed.                  |
| `cancel-requests.ts:150` — the `delivered > 0` arms are unpinned             | `bd016c0e`. Both now assert the text. Verified by collapsing both ternaries to their pessimistic constants: exactly these two tests fail and nothing else.                                                                                                                                            |
| `use-ledger.ts:210` — the wrong error is rendered                            | `15f8311f`. New `PermissionWindowFailed` status set in both branches, so the screen says the window could not be opened rather than telling the user to grant permission in a window that will never exist. The Ledger error map holds plain strings rather than i18n keys, so no catalog round trip. |

## `attach-window-to-request.ts:50` — the gap is intentional, but it was too quiet

You are right that the check establishes "one of our windows", not "the window showing this request", and right that the mismatch branch is the only thing that would distinguish them.

That is deliberate: a `requestId` mismatch during the reuse round trip is expected, not anomalous, so treating it as a verdict would cancel live approvals. But "no `requestId` param at all" producing _no signal_ was not deliberate — it left the distinction silently unobservable. Both diagnostics now warn (`063737e4`); neither repairs.

Every approval window is opened with `requestId` in its search params, so a window of ours without one is genuinely anomalous and worth a line.

## Still not filed

The two follow-up tickets from the first round — the `Request` union refactor and gating the whole `FORWARDED_ACTION_TYPES` — still need Jira keys.

Comp0te (approved):
👍

PR #1428, opened 2026-07-28
fix: WALLET-1364 — surface a failed export-keys window open

ost-ptk:
Ticket: https://make-software.atlassian.net/browse/WALLET-1364 (P1 item 3)

## The bug

The user clicks "Download account keys" in the popup navigation menu, a saga opens a separate export window, and if `windows.create` rejects the catch only writes to the background console. No window, no message, no feedback of any kind — the click simply does nothing.

The comment sitting in that catch justified the silence and was factually wrong, which is how the gap survived review. It claimed "there is no popup left to show it in". Both halves of that were checked against the code and are false:

- the dispatch comes from the popup navigation menu, whose handler calls `closeNavigationMenu()` — that only navigates router state to hide the menu **overlay**, with no window API call at all;
- the popup mounts `SagaErrorBanner` unconditionally (`src/apps/popup/index.tsx`).

When `windows.create` rejects, no new window takes focus, so the popup is still open and perfectly able to render the error.

## The fix

`yield put(sagaError({ source: 'openExportKeysWindowSaga', … }))` alongside the existing `console.error`, and the incorrect comment is deleted rather than softened — the reasoning it stated was false, so leaving a softened version would preserve the misleading premise.

The surfaced message is a short static sentence with no error text, no paths and no user data: `appEvents` is broadcast to every UI replica and the banner renders `{source}: {message}` verbatim.

## Verification

`npm run ci-check` passes (406 tests). The new test discriminates: removing the `put(sagaError(...))` makes it fail while the other five in the file — including the early-return-on-existing-window case and the watcher test — keep passing.


Comp0te (commented):
Reviewed the export-keys window saga change and its tests against the current head. The change does what the ticket asks — a failure that previously reached only the background console now reaches the user — and the surrounding structure is unusually well documented for a saga (the `takeLeading` rationale and the `type: 'popup'` guard comment both earn their place).

The comments below cluster in two places: what happens to the error entry once it exists (nothing clears or dedupes it, and the export window renders the same banner), and the new test's assertions — one of which cannot fail as written, plus a couple of provider details.

src/background/redux/sagas/export-keys-window-saga.ts
```diff
@@ -67,13 +68,16 @@ export function* openExportKeysWindowSaga() {
...
     }
   } catch (error) {
-    // No UI surface here (the menu banner was removed): a failed open is rare
-    // (windows.create almost never rejects) and there is no popup left to show
-    // it in. The download-failure screen covers the important in-window case.
     console.error(
       'openExportKeysWindowSaga: failed to open export window',
       error
     );
+    yield put(
```
Comp0te, line 106: `sagaError` entries are append-only and nothing clears this one. `app-events/reducer.ts:30-42` does `errors: [...state.errors, entry].slice(-10)` with an incrementing `nextErrorId`, so identical messages never collapse into one. The only removal paths in the tree are the banner's own dismiss button (`saga-error-banner.tsx:70`) and `resetAppEventsDismission` from `resetVaultSaga` (`onboarding-sagas.ts:67`) — nothing on success, on navigation, or on window close.

The sharp case is a retry that **succeeds**. `EXPORT_KEYS_URL` is `popup.html#…` (line 15), `popup.html` builds from `src/apps/popup/index.tsx` (`webpack.config.js:114-116, 334-336`), and that file mounts `<SagaErrorBanner />` at `:72` as a sibling of `<AppRouter />` — so it is route-independent. `appEvents` is broadcast whole (`get-main-store.ts:102`, inside `selectPopupState`), and a freshly mounted window picks up errors appended *before* it opened, via the `popupWindowInit` round-trip (`use-subscribe-to-redux.ts:35` → `get-main-store.ts:179-188`). Net effect: the second click opens the key-download screen with the first failure's red banner pinned over it, and the user cannot tell whether what they are reading is current — at the moment they are looking at their account keys.

Secondary: repeated failures stack byte-identical rows, each needing its own dismiss click, in a `position: fixed; top/left/right: 0; max-height: 100vh` container (`saga-error-banner.tsx:14-27`).

Two things that bound this, neither of which closes it: errors are memory-only (`get-main-store.ts:228` persists only `dismissedEventIds`, `:164-169` rehydrates `errors: []`), so a service-worker restart drops the banner; and append-without-dedupe predates this PR. What is new is that this is the first producer with a *constant* message reachable from a repeatable click — the others pass `errorToMessage(err)`, which varies.

Dispatching `dismissSagaError` for this source on a successful open, or deduping by `source` in the reducer so a repeat replaces rather than appends, would close it.

src/background/redux/sagas/export-keys-window-saga.ts
```diff
@@ -67,13 +68,16 @@ export function* openExportKeysWindowSaga() {
...
+      sagaError({
+        source: 'openExportKeysWindowSaga',
+        message: 'Could not open the export window'
```
Comp0te, line 109: A narrow accuracy point on this string, separate from the comment above on the `put` itself.

The `try` opened at line 20 also wraps the refocus `call` at `:33-36`, so when `windows.update` rejects the user is told the window could not be **opened** — but that path is only reached after `windows.getAll` at `:24` found the window. If it genuinely still exists and only focusing it failed, the message contradicts what is on screen.

Worth being explicit about what this is *not*, since three plausible-sounding consequences don't hold: retaining the window id in that branch is correct (the window exists, and `background/index.ts:184-192` clears the id on `windows.onRemoved`); not falling through to `windows.create` is also correct, since that would open a second export window; and `getAll`/`getCurrent` rejections are fine as-is, because nothing was opened by that point, so the generic message is accurate there.

So it is only the `windows.update` branch, and only when the window really is still there. The likelier way `update` rejects is the window closing between `:24` and `:33`, where "could not open" is roughly accurate anyway and `onRemoved` clears the id. Low value — but if you want it, a separate `catch` around just the refocus with its own message covers it.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -86,6 +87,29 @@ describe('openExportKeysWindowSaga', () => {
       .run();
   });
 
+  it('surfaces a saga error when the export window cannot be opened', () => {
```
Comp0te: This is the only failure test, and it covers `windows.create` rejecting with `exportKeysWindowId: null`. The uncovered path that behaves *differently* is `windows.update` rejecting — the refocus branch, where the message is arguable (see the comment on the saga). `getAll` and `getCurrent` rejections land in the same catch with the same message, so their absence matters less.

There is also no assertion that exactly **one** `sagaError` is emitted: `.put(...)` here is a containment check that removes one matching effect from the recorded set, so a duplicate would pass unnoticed.

The gap is load-bearing rather than cosmetic. Moving the refocus `call` at `:33-36` into a helper generator with its own swallowing catch — i.e. restoring exactly the silent no-op this PR is fixing — keeps all six tests in this file green. An unprovided `call` that yields an iterator gets re-wrapped by redux-saga-test-plan, so the nested `windows.update` call still matches the `matchers.call.fn(windows.update)` provider and is still recorded; nothing in the suite can tell the two behaviours apart.

One more `expectSaga` case — `withState({ windowManagement: { exportKeysWindowId: 42 } })`, `getAll` returning `[{ id: 42, type: 'popup' }]`, and `windows.update` provided a rejection — would pin it.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -86,6 +87,29 @@ describe('openExportKeysWindowSaga', () => {
...
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    return expectSaga(openExportKeysWindowSaga)
+      .withState({ windowManagement: { exportKeysWindowId: null } })
+      .provide([
+        [matchers.call.fn(windows.getAll), []],
```
Comp0te: This provider is never reached. `withState({ windowManagement: { exportKeysWindowId: null } })` on the line above makes the guard at `export-keys-window-saga.ts:23` (`if (windowId != null)`) short-circuit, so `windows.getAll` is never yielded.

The sibling at `:68-77` has the same `null` state and correctly omits it; `:55-66` uses `42`, where the `getAll` provider genuinely is reached — likely where this line was copied from.

Worth deleting rather than leaving: the `.provide([...])` array reads as documentation of the call sequence for the scenario, and this entry implies the failure test walks the stale-id → clear → create path, which it does not.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -86,6 +87,29 @@ describe('openExportKeysWindowSaga', () => {
...
+        [matchers.call.fn(windows.getCurrent), currentWindow],
+        [matchers.call.fn(windows.create), Promise.reject(new Error('denied'))]
```
Comp0te: `Promise.reject(new Error('denied'))` is constructed when this array literal is evaluated, before `.run()` is called. A static provider value is stored untouched and handed to redux-saga only when the effect matches, so nothing attaches a rejection handler until the `windows.create` yield is actually reached.

It is safe today only because that yield is reached within the same turn: `.run()` starts the saga synchronously, `select` resolves off `withState`, and the `getCurrent` provider returns a plain object with no promise hop. Introduce any macrotask boundary before that yield — a provider returning a real promise, a `delay`, a pending mock — and this line emits an `unhandledRejection` that will look unrelated to whatever change introduced it.

`redux-saga-test-plan/providers` exports `throwError(err)` for exactly this case; it throws at the effect site instead of pre-building a rejected promise. It is not used anywhere in `src` yet, so this would be the first — which is also why it is worth setting the precedent here rather than in the next test that copies this one.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -86,6 +87,29 @@ describe('openExportKeysWindowSaga', () => {
...
+      ])
+      .put(
+        sagaError({
+          source: 'openExportKeysWindowSaga',
+          message: 'Could not open the export window'
+        })
+      )
+      .not.put(exportKeysWindowIdChanged(expect.any(Number)))
```
Comp0te: This assertion is unconditionally satisfied — it passes whether or not the saga dispatches the action.

redux-saga-test-plan compares expected effects against recorded ones with `lodash.isEqual`, which knows nothing about jest's asymmetric matchers, so `expect.any(Number)` never equals a real numeric payload. You can see it directly by breaking the *working* positive assertion at `:64`: change `.put(exportKeysWindowIdChanged(77))` to `.put(exportKeysWindowIdChanged(expect.any(Number)))` and run `npx jest src/background/redux/sagas/export-keys-window-saga.test.ts` — that test now fails, even though the saga demonstrably dispatches `exportKeysWindowIdChanged(77)`. The matcher simply never matches; under `.not.put` that failure is silent and the line reads as a guarantee while guaranteeing nothing.

Asserting on the action type works: `.not.put.actionType(exportKeysWindowIdChanged.type)`. Or the idiom already used at `vault-sagas.test.ts:483-496` — destructure `{ effects }` from the run and `expect(putTypes).not.toContain(exportKeysWindowIdChanged.type)`.

src/background/redux/sagas/export-keys-window-saga.ts
ost-ptk: Fixed in 3880c97f — the saga now retracts its own source before each attempt.

Took the `dismissSagaError`-on-retry route over reducer-level dedupe, but dispatching it *before* the attempt rather than on success: one effect then covers both halves of what you described — the stale banner over the key-download screen after a retry that succeeds, and the identical rows stacking after repeated failures. Dedupe in the reducer would only fix the second.

Kept it scoped to one producer instead of changing `sagaError` globally: the other 14 call sites carry `errorToMessage(err)` or an interpolated `tabId`, so collapsing by `source` would be a behaviour change for them with no problem to solve.

The existing `dismissSagaError` takes an id the saga has no way to know, so this is a new reducer case, `dismissSagaErrorsBySource(source)`. It is background-only, so it went into the parity test's `EXCLUSIONS` next to `sagaError` rather than the forwarding set — `redux-actions.parity.test.ts` caught that omission on the first `ci-check` run, which is the guard working as designed.

One deliberate non-change: `nextErrorId` is not rewound on retraction, so a retracted id is never reissued. Pinned by a test.

src/background/redux/sagas/export-keys-window-saga.ts
ost-ptk: Fixed in 3880c97f. Agreed the window-closed-in-between case is the likelier one and roughly accurate as it stood, but the branch now has its own `catch` and its own message ("Could not focus the export window") — cheap, and it is what made the gap below testable.

Both non-changes you flagged are kept and now stated in the code: the tracked id stays (the window exists, `onRemoved` clears it) and control still `return`s rather than falling through to `windows.create`. The new test asserts both — `.not.put.actionType(exportKeysWindowIdCleared.type)` and `.not.call.fn(windows.create)` — so a later refactor cannot quietly turn either into the other behaviour.

`getAll`/`getCurrent` rejections are left in the outer catch as you described.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Both fixed in 3880c97f.

Added the `windows.update`-rejection case roughly as you specified. Ran your falsification directly: replacing the new refocus `catch` with a silent `catch {}` — the exact no-op this PR fixes — fails that test and only that test, 1 failed / 7 passed. Before it existed, the same mutation left the file green.

For "exactly one", both failure tests now count puts of `sagaError.type` alongside the `.put(...)` payload assertion. One detail worth recording: the count has to read `allEffects`, not `effects`. `.put(...)` removes the matched effect from `effects.put` (measured it — length goes 2 → 1), so counting there after asserting yields 0 and the test fails for the wrong reason. `allEffects` is untouched by the expectations. Verified the assertion bites by duplicating the `put` in the outer catch: fails as expected.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Deleted in 3880c97f. The provider list in that test is now `getCurrent` + `create` only, matching the sibling at what was `:68-77`.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Fixed in 3880c97f — `throwError` from `redux-saga-test-plan/providers`, in both failure tests, with the reasoning kept in a comment at the call site so the next copy of this test carries it forward.

Confirmed it is the first use in `src`. The refocus test added below would have been the second place to repeat the pattern, so the timing worked out.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Confirmed by running your reproduction before changing anything: swapping the working positive assertion `.put(exportKeysWindowIdChanged(77))` for `.put(exportKeysWindowIdChanged(expect.any(Number)))` fails, on a saga that demonstrably dispatches exactly that action. The matcher never matches, and under `.not.put` that is silent.

Fixed in 3880c97f with `.not.put.actionType(...)` — in all **three** places, not just this one. The same construct sits at what were `:50` and `:86`, both predating this branch and both equally vacuous. Negative control on the replacement: applied `.not.put.actionType(exportKeysWindowIdChanged.type)` to a test where the action *is* dispatched, and it fails, so the new form genuinely bites.

Left a comment at the first site recording why it is asserted on the type rather than on a built action, since the vacuous form is the one that looks natural.

Comp0te (approved):

PR #1429, opened 2026-07-28
build: WALLET-1364 — emit the CSP nonce only for Chrome production builds

ost-ptk:
Tickets: https://make-software.atlassian.net/browse/WALLET-1364 (P1 item 4) ·
https://make-software.atlassian.net/browse/WALLET-1388

## The problem

`webpack.config.js` generated a random CSP nonce per build and injected it through
`DefinePlugin` for **every** browser target, even though only the Chrome-production CSP arm
pins `style-src` to it. Both readers — `src/set-webpack-nonce.ts` (webpack's
`__webpack_nonce__`, read by style-loader) and `src/libs/ui/style-sheet-manager.tsx`
(styled-components) — are browser-agnostic, so every Firefox and Safari bundle carried a
fresh random base64 literal that nothing ever used.

AMO source review rebuilds the submitted source and compares the result to the uploaded
artifact byte for byte. A per-build random literal makes that impossible.

## The fix

The value is generated only when `isChrome && !isDev`; every other target gets `null`.

That predicate is now spelled out in exactly **one** place. `getCSP()` derives its
`style-src` arm from the value (`CSP_NONCE ? … : …`) rather than re-deriving the condition,
so the manifest CSP and the bundle literal cannot drift apart. A drift would have been
silent and total: `style-src 'self' 'nonce-null'` is syntactically valid, matches nothing,
and blocks every styled-components sheet and style-loader `<style>` in all five apps.

The `DefinePlugin` key stays defined for all targets on purpose — dropping it would leave
the free variable unreplaced and throw `ReferenceError` at runtime. Only the value changes.

Deriving the nonce from the commit hash was considered and rejected: a publicly computable
nonce adds nothing over a static one, whereas gating removes the value from the non-Chrome
bundles altogether.

## Making the null case type-checked

The constant moved off `process.env.CSP_NONCE` to a dedicated `__CSP_NONCE__`, declared in
`src/@types/custom.d.ts` as `string | null`.

`@types/node` types every `ProcessEnv` member as `string | undefined`, so the `null` that
`DefinePlugin` substitutes on non-Chrome targets was outside the declared type. The
null-handling at each reader type-checked clean **with or without it** — a "redundant
nullish coalescing" cleanup would have shipped `nonce={null}` past tsc, eslint and jest,
and styled-components branches on `props.nonce !== undefined`, so a `null` pulls those
components off the shared style sheet.

With the value declared honestly, `?? undefined` in `CspStyleSheetManager` is now required
by the compiler — removing it fails `tsc` with TS2322. `jest.config.js` supplies
`globals: { __CSP_NONCE__: null }` so the free variable resolves under test too.

`src/set-webpack-nonce.ts` also gained `export {}`: without an import or export TypeScript
treats it as a global script, and its top-level `const` leaked into the project-wide scope,
where any file could reference a bare `nonce` and type-check while webpack failed to
resolve it at runtime.

## WALLET-1388, folded in here

`dotenv-webpack` runs with `systemvars: true` and is registered **before** `DefinePlugin`,
so an ambient `CSP_NONCE` — a shell export, CI, a stale local dotenv file — won inside the
bundle while the manifest kept the generated value. Every injected stylesheet blocked, build
green. Same file, same reviewer, so it rides along instead of becoming a conflicting PR.

Moving the constant off `process.env` (above) closes it structurally: `dotenv-webpack` only
ever emits `process.env.*` keys, so it has no way to reach `__CSP_NONCE__`. The ticket's
repro is inert on this branch — with `CSP_NONCE=SHADOWVALUE123 npm run build:chrome`, the
manifest and `popup.bundle.js` carry the same generated nonce and no emitted asset contains
`SHADOWVALUE123`.

The ticket's second suggestion is in as well, because "there happens to be no path for it
today" is not a guarantee: **`AssertCspNonceIntegrity`** (`webpack.config.js`) compares the
artifacts on the last `processAssets` stage. Whatever the emitted `manifest.json` pins must
equal the value this build generated, and every entry that gets the nonce setter prepended
must have a bundle containing that literal — otherwise the build fails.

- `processAssets`, not `afterEmit`: by `afterEmit` webpack has swapped every emitted source
  for a `SizeOnlySource` and reading one throws. Working from the compilation rather than
  from disk also keeps it working under webpack-dev-server.
- It throws rather than pushing to `compilation.errors`: `utils/build.js` rethrows only the
  fatal compiler error and never inspects stats, so a compilation error would not fail the
  build.
- The guarded entries are read back from webpack's normalized entry, so they cannot drift
  from the entries that actually get the setter prepended.

## Making the wiring gated

Nothing in CI evaluated `getCSP()` — `knip` executes the config only to harvest entries and
aliases, and format/lint/tsc/jest never load it at all. Two gates were added:

- **`src/webpack-config.test.ts`** requires the config under all six BROWSER × NODE_ENV
  combinations and asserts the invariant rather than the value: the manifest CSP carries a
  nonce **iff** the `DefinePlugin` literal is a non-empty string, and it is the same string.
  It also drives `AssertCspNonceIntegrity` through a fake compiler/compilation, and pins the
  pre-existing Safari behaviour — `getCSP()` branches on `isFirefox`/`isChrome` only, so the
  Safari manifest ships without a `content_security_policy` key — so that a change to it is
  a deliberate one. 22 cases; mutation-checked rather than assumed:

  | mutation                             | cases red |
  | ------------------------------------ | --------- |
  | predicate → `isChrome && isDev`      | 5         |
  | `if (false && pinned !== CSP_NONCE)` | 3         |
  | `if (false && !carriesNonce)`        | 1         |

- **`e2e-tests/fixtures.ts`** registers a `securitypolicyviolation` listener on the browser
  context and fails the test if anything was blocked. e2e is the only job that builds a
  production Chrome artifact, and its existing `getByRole`/`toBeVisible` assertions all
  pass against an unstyled-but-present DOM, so the page has to report the violation itself.
  Scoped to CSP violations only — page errors are deliberately not asserted, since the
  suite has never been held to that bar.

The e2e workflows now also trigger on `release/**`, matching `ci-check`; previously they ran
only on `master`/`develop`, so the one job exercising a production Chrome build never ran on
a release PR.

## Verification

- Two consecutive `npm run build:firefox` runs are **byte-identical** after this change and
  **differed** before it — the difference was confined to exactly the five bundles that
  prepend the nonce setter, which matches the entry wiring.
- `build/chrome/manifest.json` still carries a real `nonce-<base64>` in its CSP, and the
  Chrome bundle still sets it at runtime.
- No `null`/`undefined` string leaked into the Firefox output where the nonce used to be.
- `npm run ci-check` passes; the one eslint warning the previous revision added
  (`Unused eslint-disable directive` for a `prefer-const` that no longer fires) is gone.
- The emit-time assertion was fault-injected, not assumed: pinning `nonce-${CSP_NONCE}FAULT`
  in `getCSP()` makes `npm run build:chrome` exit 1 with the mismatch reported. Chrome and
  Firefox production builds and `npm run start:chrome` (webpack-dev-server) all pass with it
  in place.
- Both e2e suites pass with the new CSP listener; breaking the predicate makes them fail
  with `[CSP] style-src-elem blocked inline`.


Comp0te (commented):
Reviewed the CSP nonce narrowing across `webpack.config.js`, `src/set-webpack-nonce.ts` and `src/libs/ui/style-sheet-manager.tsx`. The change itself is right — the random base64 really was being inlined into the Firefox/Safari bundles, and building Firefox twice at this head produces byte-identical output, so the reproducibility goal is reached for those artifacts.

The comments below cluster on one theme: the mechanism that makes this correct is narrower than the surrounding prose suggests, and none of it is visible to a gate. The load-bearing piece is the predicate at `webpack.config.js:74-75` together with the requirement that it agree with the CSP arm at `:86-88`; the guards in the other two files are good hygiene but are justified by protections the dependencies already provide; and the one thing that would catch a regression here — a test over the BROWSER/NODE_ENV matrix — doesn't exist.

webpack.config.js
```diff
@@ -64,11 +64,15 @@ if (fileSystem.existsSync(secretsPath)) {
...
-const CSP_NONCE = crypto.randomBytes(16).toString('base64');
+// One nonce per build, and ONLY for Chrome production — that is the sole CSP arm
+// that pins style-src to it (see getCSP below). Emitting a random literal into
+// Firefox/Safari bundles that never read it makes those builds irreproducible,
+// which AMO source review requires (it rebuilds and compares byte-for-byte).
+// Where it does apply, the same value is wired to __webpack_nonce__ (style-loader)
+// and process.env.CSP_NONCE (styled-components' CspStyleSheetManager) below, so
+// the manifest CSP and the injected <style> tags always match.
+const CSP_NONCE =
+  isChrome && !isDev ? crypto.randomBytes(16).toString('base64') : null;
```
Comp0te, line 80: `isChrome && !isDev` now decides two things independently: whether a nonce is generated at all (here), and whether the manifest CSP pins one (`:86-88`). Before this change the nonce was unconditional and only the CSP arm branched, so the two could not disagree.

If they ever drift, `getCSP()` interpolates the literal `null` and emits `style-src 'self' 'nonce-null'; style-src-attr 'unsafe-inline'` — syntactically valid, and nothing can match it. No layer stamps a bogus nonce to compensate either: `__webpack_nonce__` stays unset and styled-components' `getNonce()` finds no meta tag in this repo. So every styled-components sheet and every style-loader `<style>` is blocked in Chrome, and popup, onboarding, connect-to-app, signature-request and import-account all render unstyled.

Nothing catches it. `knip` does execute this file (`knip.json:13-15`), but only to harvest entries, loaders and aliases — it never evaluates `getCSP()`; `format:check`, `lint`, `tsc` and jest never load the config at all.

Deriving the CSP arm from the value instead of re-deriving the predicate makes the desync unrepresentable:

```js
const styleDirectives = CSP_NONCE
  ? `style-src 'self' 'nonce-${CSP_NONCE}'; style-src-attr 'unsafe-inline'`
  : "style-src 'unsafe-inline'";
```

webpack.config.js
```diff
@@ -64,11 +64,15 @@ if (fileSystem.existsSync(secretsPath)) {
...
 
 const getCSP = () => {
```
Comp0te, line 82: Separately from the predicate duplication just above — nothing automated exercises the branch this PR introduces.

Neither changed module has a test file (no `*.test.tsx` exists anywhere under `src`, and nothing references `CspStyleSheetManager`), and `ci-check` never evaluates `getCSP()` — `knip` executes this config (`knip.json:13-15`) but only harvests entries, loaders and aliases.

The one job that would run the live Chrome-production nonce path is `e2e:chrome:headless:popup`, which builds `NODE_ENV=production BROWSER=chrome`. It doesn't run on this PR: `.github/workflows/e2e-popup-tests.yml:4-7` and `e2e-onboarding-tests.yml:4-7` trigger only on `[master, develop]`, while this targets `release/2.7.0`. Only `ci-check.yml` includes `release/**`. And even where they do run, every assertion is `getByText`/`getByRole`/`getByPlaceholder` plus `toBeVisible`/`toBeDisabled`/`toBeEnabled`, with no `console`, `pageerror` or `securitypolicyviolation` listener registered anywhere in `e2e-tests/` or `playwright.config.ts` — `toBeVisible()` only needs a non-empty bounding box, so an unstyled-but-present DOM satisfies all of them. The regression that would slip through is flipping the `:74` predicate.

The cheap version is a root-level `webpack-config.test.ts` — jest's `testRegex` (`jest.config.js:11`) matches it, `tsconfig.json` has `allowJs: true`, and `collectCoverageFrom` is scoped to `src/background/**` so the coverage gate is untouched. Require this config under the six BROWSER/NODE_ENV combinations with `jest.resetModules()` between cases and assert the invariant rather than the value: the manifest CSP contains `nonce-<x>` iff the `DefinePlugin` literal is a non-empty string, and `<x>` equals that literal. Adding `release/**` to the e2e triggers is worth considering too, since e2e is the only gate that runs a production Chrome artifact.

src/set-webpack-nonce.ts
```diff
@@ -2,9 +2,14 @@
...
 // 'self' 'nonce-<value>' with no 'unsafe-inline'). Must be imported FIRST in every
 // app entry, before any style-loader CSS import or styled-components injection runs.
-declare let __webpack_nonce__: string;
-// Ambient webpack global: style-loader's runtime reads it off the global scope (not
-// from this module), so ESLint can't see the "use", and it must stay `let` since it's
-// assigned after declaration rather than initialized inline.
-// eslint-disable-next-line prefer-const, @typescript-eslint/no-unused-vars
-__webpack_nonce__ = process.env.CSP_NONCE as string;
+declare let __webpack_nonce__: string | undefined;
+const nonce = process.env.CSP_NONCE;
```
Comp0te: This file has no `import` and no `export`, so TypeScript treats it as a global script rather than a module — which makes the new `const nonce` at `:6` a project-wide global type binding. Any file under `src` can now reference a bare `nonce` and `tsc` will accept it, while webpack scopes the real binding to this module: a runtime `ReferenceError` that type-checking cannot catch. The pre-PR file created no value binding at all, so this is new.

`nonce` is a fairly collision-prone identifier in a wallet codebase that also deals with cryptographic nonces. `export {};` at the end of the file restores the check — with it, a sibling `src/other.ts` doing `export const leaked: string | undefined = nonce;` (no import) fails with TS2304; without it, that file compiles clean.

src/set-webpack-nonce.ts
```diff
@@ -2,9 +2,14 @@
...
+// Only Chrome-production defines a nonce; elsewhere style-src keeps
+// 'unsafe-inline' and style-loader must not stamp a `nonce="null"`.
```
Comp0te: The requirement this comment states is already guaranteed by the dependency, so the guard below isn't what prevents a `nonce="null"` — which makes the stated reason redundant rather than load-bearing.

Every style-loader runtime path that reads `__webpack_nonce__` checks truthiness before setting the attribute: `style-loader/dist/runtime/setAttributesWithoutAttributes.js:5-8`, `setAttributesWithAttributes.js:5-8`, `injectStylesIntoLinkTag.js:9-14`. And `DefinePlugin` substitutes the literal `null` here, not the string `"null"` (`webpack.config.js:216` wraps the value in `JSON.stringify`). styled-components behaves the same way — `makeStyleTag` does `const resolvedNonce = nonce || getNonce(); if (resolvedNonce) … setAttribute` (`styled-components/dist/styled-components.js:358-360`).

The `if (nonce)` guard is worth keeping, just for other reasons: it keeps `__webpack_require__.nc` unset on non-Chrome targets, and it keeps the `string | undefined` declaration at `:5` honest — assigning the substituted `null` to it wouldn't type-check. Naming those would point the next person auditing this wiring at the layer that actually carries weight.

src/set-webpack-nonce.ts
```diff
@@ -2,9 +2,14 @@
...
+if (nonce) {
+  // Ambient webpack global: style-loader's runtime reads it off the global scope (not
+  // from this module), so ESLint can't see the "use", and it must stay `let` since it's
+  // assigned after declaration rather than initialized inline.
+  // eslint-disable-next-line prefer-const, @typescript-eslint/no-unused-vars
```
Comp0te: Wrapping the assignment in `if (nonce)` stopped `prefer-const` from firing, so the `prefer-const` half of this directive is now dead. `npx eslint src/set-webpack-nonce.ts` reports:

```
13:3  warning  Unused eslint-disable directive (no problems were reported from 'prefer-const')
```

It's the one new warning this PR adds. `lint` is a bare `eslint ./src` with no `--max-warnings`, so it joins the existing backlog and `ci-check` stays green — the gate can't tell you it happened.

The comment above has the same issue: "it must stay `let` since it's assigned after declaration rather than initialized inline" explains a rule that no longer applies here. `@typescript-eslint/no-unused-vars` is still needed.

Narrowing to `// eslint-disable-next-line @typescript-eslint/no-unused-vars` and dropping that trailing clause covers both.

src/libs/ui/style-sheet-manager.tsx
```diff
@@ -11,7 +11,7 @@ export const CspStyleSheetManager = ({ children }: PropsWithChildren) => (
     // styled-components stamps this nonce on its injected <style> tags so they pass the
     // Chrome-prod `style-src 'self' 'nonce-<value>'` CSP; value comes from webpack DefinePlugin
     // (inert on dev + Firefox/Safari, whose CSP keeps 'unsafe-inline'). See webpack.config.js.
```
Comp0te: These three lines are unchanged, but line 14 directly beneath them falsified part of what they say.

"styled-components stamps this nonce on its injected `<style>` tags … value comes from webpack DefinePlugin" was accurate before this PR: a real value was always generated and always passed, just not enforced by the non-Chrome CSP. Now no value reaches this prop on dev/Firefox/Safari, so nothing is stamped from here at all — styled-components falls through to its own `getNonce()` discovery chain instead (`meta[property="csp-nonce"]` → `meta[name="sc-nonce"]` → `__webpack_nonce__`, `styled-components/dist/styled-components.js:328-342`), which finds nothing in this codebase: `grep -rE 'sc-nonce|csp-nonce' src` returns no hits.

The "inert on dev + Firefox/Safari" parenthetical is still true — arguably more true than before. It's the stamping sentence that now describes the old mechanism. Worth saying that no nonce is passed on those targets, and that the fallback chain currently resolves to nothing.

src/libs/ui/style-sheet-manager.tsx
```diff
@@ -11,7 +11,7 @@ export const CspStyleSheetManager = ({ children }: PropsWithChildren) => (
...
-    nonce={process.env.CSP_NONCE}
+    nonce={process.env.CSP_NONCE ?? undefined}
```
Comp0te: `?? undefined` is doing real work here, and nothing in the repo can tell if it's removed.

`process.env.CSP_NONCE` has no declaration in this project — `src/global.d.ts` is a single import and `src/@types/custom.d.ts` only declares asset modules — so its type comes entirely from `@types/node`'s `interface ProcessEnv extends Dict<string>`, i.e. `string | undefined`. `null` isn't in that type and never was, yet `DefinePlugin` substitutes exactly that on every non-Chrome-prod target.

So `nonce={process.env.CSP_NONCE}` without the coalesce type-checks clean, lints clean, and passes the whole jest suite — under jest `DefinePlugin` never runs, so the value genuinely is `undefined` and no test can distinguish the two. A "redundant nullish coalescing" cleanup, or enabling `@typescript-eslint/no-unnecessary-condition`, would ship `nonce={null}` with all five gates green. That isn't cosmetic: styled-components gates on `props.nonce !== undefined` (`dist/styled-components.js:1263`), so a `null` triggers `reconstructWithOptions({ nonce: null })` and takes these components off the shared main sheet.

Worth declaring the value. Either augment `NodeJS.ProcessEnv` in `src/global.d.ts` with `readonly CSP_NONCE: string | null | undefined;`, or — better — stop routing a build constant through `process.env` at all: define `__CSP_NONCE__` in `DefinePlugin` and `declare const __CSP_NONCE__: string | null;`.

webpack.config.js
ost-ptk: Taken, in the shape you wrote — `getCSP()` now derives its `style-src` arm from the value instead of re-deriving the predicate (`b0041363`):

```js
const styleDirectives = CSP_NONCE
  ? `style-src 'self' 'nonce-${CSP_NONCE}'; style-src-attr 'unsafe-inline'`
  : "style-src 'unsafe-inline'";
```

`isChrome && !isDev` is now spelled out exactly once, at `:79`, and `'nonce-null'` is unrepresentable.

That removes the duplication, but "the manifest and the bundles agree" is still only a property of how this file is written — nothing checks it. So `89953bb8` adds `AssertCspNonceIntegrity`, registered right after `DefinePlugin`, which compares artifacts rather than intent: the nonce the emitted `manifest.json` pins must equal the value this build generated, and every entry that gets the nonce setter prepended must have a bundle containing that literal.

Two implementation notes, both found the hard way:

- It taps the last `processAssets` stage, not `afterEmit`. By `afterEmit` webpack has replaced every emitted source with a `SizeOnlySource`, and `.source()` throws. `processAssets` at `PROCESS_ASSETS_STAGE_REPORT` also runs after CopyWebpackPlugin has added the manifest (it emits at `…_ADDITIONAL`), and reading the compilation instead of the disk keeps it working under webpack-dev-server — verified, `npm run start:chrome` compiles clean.
- It throws rather than pushing to `compilation.errors`. `utils/build.js` only rethrows the fatal `err` and never inspects stats, so a compilation error would not fail the build. A throw does: tapable forwards it to the seal callback and webpack hands it over as the compiler error.

Verified by fault injection — pinning `nonce-${CSP_NONCE}FAULT` in `getCSP()` and building:

```
$ npm run build:chrome; echo $?
Error: CSP nonce check: the manifest pins "nt0ygnkuUnH311bo3abfLw==FAULT", but this build
generated "nt0ygnkuUnH311bo3abfLw==". Every injected stylesheet would be blocked.
1
```

This is also what closes WALLET-1388, now pulled into this PR — see the `?? undefined` thread.

webpack.config.js
ost-ptk: Fixed — the branch has three gates now, and the root-level `webpack-config.test.ts` is the one you sketched.

**`src/webpack-config.test.ts`** (`59431dac`, extended in `89953bb8`) requires the config under all six BROWSER × NODE_ENV combinations with `jest.resetModules()` between cases, and asserts the invariant rather than the value: the manifest CSP contains `nonce-<x>` iff the `DefinePlugin` literal is a non-empty string, and `<x>` equals that literal. It also pins the pre-existing Safari behaviour (no `content_security_policy` key at all) so a change to it has to be deliberate. `collectCoverageFrom` is untouched, as you noted.

Flipping `:79` to `isChrome && isDev` turns 5 of the 22 cases red.

**e2e** — `release/**` added to both workflow triggers, so the only job that builds a production Chrome artifact now runs on this PR. And since `toBeVisible()` accepts an unstyled DOM, `e2e-tests/fixtures.ts` registers a `securitypolicyviolation` listener via `addInitScript` and fails the context fixture if anything was blocked. Scoped to `chrome-extension:` documents — the suite also drives Topper (the buy-CSPR redirect) and the playground dapp, and Topper blocks its own analytics on every load.

**The emit-time assertion** from the thread above is the only gate that inspects real artifacts; `89953bb8` covers it with 8 more cases driving it through a fake compiler/compilation (the four things it touches: the two hooks, the normalized entry, the entrypoint file lists, `getAsset`). Its fake entry is the config's real `entry`, so a change to which entries get the setter reaches those cases automatically.

Mutation-checked rather than assumed:

| mutation | red |
| --- | --- |
| `if (false && pinned !== CSP_NONCE)` | 3 |
| `if (false && !carriesNonce)` | 1 |
| `:79` → `isChrome && isDev` | 5 |

src/set-webpack-nonce.ts
ost-ptk: Added — `export {};` closes the file, with a comment saying why it has to stay a module rather than looking like a stray line someone can delete:

```ts
// Keeps this file a module. Without it TypeScript treats it as a global script and
// every top-level binding here leaks into the project-wide scope, where `tsc` would
// accept a bare reference from any other file that webpack then fails to resolve.
export {};
```

The binding it was leaking is gone too, incidentally: `const nonce` was replaced by the `__CSP_NONCE__` global (see the `?? undefined` thread), so there is no longer a `nonce` identifier here at all. `export {}` stays regardless — the next top-level binding added to this file would reintroduce the leak.

src/set-webpack-nonce.ts
ost-ptk: Right — the comment was pointing at the wrong layer. Rewritten to name what the guard actually carries and to credit the dependency for the rest:

```ts
// __CSP_NONCE__ is a real string only on Chrome production; elsewhere it is null
// and style-src keeps 'unsafe-inline'. style-loader already refuses to stamp a
// falsy nonce on its own (dist/runtime/setAttributes*.js), so this guard is not
// what prevents a `nonce="null"` attribute. What it does carry: it leaves
// __webpack_require__.nc unset on non-Chrome targets, and it keeps the
// `string | undefined` declaration above honest — assigning a `string | null`
// to it would not type-check.
```

Both reasons you listed, kept in that order, with the style-loader runtime files named so the next person auditing this does not have to rediscover them.

src/set-webpack-nonce.ts
ost-ptk: Confirmed and fixed. Narrowed to the rule that still fires and dropped the `let` clause from the comment above it:

```ts
// Ambient webpack global: style-loader's runtime reads it off the global scope,
// not from this module, so ESLint cannot see the "use".
// eslint-disable-next-line @typescript-eslint/no-unused-vars
__webpack_nonce__ = __CSP_NONCE__;
```

```
$ npx eslint src/set-webpack-nonce.ts src/libs/ui/style-sheet-manager.tsx src/webpack-config.test.ts; echo $?
0
```

The gap you pointed at underneath is untouched and real: `lint` is a bare `eslint ./src` with no `--max-warnings`, so nothing here could have told me. Turning it on today means clearing 171 pre-existing warnings first (160 of them `import/no-cycle`), which is not this PR — but it is worth its own ticket, since the gate currently cannot report a regression it detects.

src/libs/ui/style-sheet-manager.tsx
ost-ptk: Rewritten — the stamping sentence described the old mechanism, as you said:

```tsx
// On Chrome production styled-components stamps this nonce on its injected <style>
// tags so they pass the `style-src 'self' 'nonce-<value>'` CSP. On dev + Firefox/Safari
// __CSP_NONCE__ is null, so no nonce is passed at all and styled-components falls
// through to its own discovery chain (meta[property="csp-nonce"] → meta[name="sc-nonce"]
// → __webpack_nonce__), which resolves to nothing here — harmless, since those CSPs
// keep 'unsafe-inline'. See webpack.config.js.
```

It now says what actually happens on the non-Chrome targets — nothing is passed, the fallback chain runs and finds nothing — and keeps the "harmless, because those CSPs keep 'unsafe-inline'" half, which is the part that was still true.

src/libs/ui/style-sheet-manager.tsx
ost-ptk: Took the second option — the build constant no longer goes through `process.env` at all (`f5816b75`).

`DefinePlugin` now defines `__CSP_NONCE__`, declared in `src/@types/custom.d.ts` as `string | null`, and `jest.config.js` supplies `globals: { __CSP_NONCE__: null }` so the free variable resolves under test too. The key stays defined for every target on purpose — dropping it on non-Chrome would leave the variable unreplaced and throw `ReferenceError`.

`?? undefined` is now held by the compiler rather than by convention:

```
$ npx tsc --noEmit          # with `nonce={__CSP_NONCE__}`
src/libs/ui/style-sheet-manager.tsx(21,5): error TS2322: Type 'string | null' is not
assignable to type 'string | undefined'.
```

and the comment says why, so the next cleanup pass has the reason in front of it rather than having to reconstruct the styled-components `props.nonce !== undefined` branch.

---

**This also closes WALLET-1388**, which I filed off your review and have pulled into this PR since it is the same file and the same reviewer.

That bug was the mirror image of what you describe here: `dotenv-webpack` runs with `systemvars: true` and is registered _before_ `DefinePlugin`, so an ambient `CSP_NONCE` — a shell export, CI, a stale dotenv file — won inside the bundle while the manifest kept the generated value. Every stylesheet blocked, build green.

Moving off `process.env` closed it structurally: `dotenv-webpack` only ever emits `process.env.*` keys, so it has no way to reach `__CSP_NONCE__`. Verified on this branch with the ticket's repro:

```
$ CSP_NONCE=SHADOWVALUE123 npm run build:chrome
$ grep -o "nonce-[A-Za-z0-9+/=]*" build/chrome/manifest.json
nonce-g/DZXhvsUKBDJV3wupGWqw==
$ grep -c -F "g/DZXhvsUKBDJV3wupGWqw==" build/chrome/popup.bundle.js
2
$ grep -l -F SHADOWVALUE123 build/chrome/*.js
(no matches)
```

I did not want to leave it at "the current wiring happens not to have a path for it", so the ticket's other suggestion is in as well — the emit-time assertion described in the `webpack.config.js:80` thread, which fails the build whenever the manifest and the bundles disagree, whatever the mechanism.

Comp0te (approved):

Issue #19, opened 2022-01-20
As a Signer user I can Review and Sign transactions on the Casper network


ghost:
Tasks:

- Implement UI
- Write business logic to call Casper SDK
- Tests should also be added to playground application so anyone can test transactions

User stories:
As a Casper Wallet user, when I initiate a transaction using a an app/dapp, I can review the details of the transaction on the Casper Network before I sign it.
As a Casper Wallet user, I can cancel a transaction prompt if I don't want to execute the transaction
As a Casper Wallet user, I can sign a transaction on the Casper Network

Issue #14, opened 2022-01-20
As a Signer user I can Remove an account


ghost:
Any imported account (ie one uploaded by file) can be removed from Signer. Once an account is removed it is no longer accessible from within Signer. It is no longer visible in the Accounts list. The only way to bring it back is to reimport the private key file again for that same account.

Figma: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Signer?node-id=1196%3A5966

![image](https://user-images.githubusercontent.com/97249036/172067108-d6f41051-2fa2-495c-8d02-3012eddb17d9.png)

**Requirements**
The account settings screen for imported accounts will include a trash can button. When clicked, the user is directed to a Remove Account? confirmation screen. 
Three buttons are on this screen:
- Back - directs the user back to the Account Settings screen with no change.
- Remove - acts as a confirmation button, removing the account, and navigating the user to the Home screen. The removed account will no longer be present in the account list.
- Cancel - directs the user back to the Account Settings screen with no change.

HD accounts, those which were not imported but rather created via secret phrase, cannot be removed ever. The trash icon will not appear in the Account Settings menu for these accounts.

**Edge cases**
If the account is active when it is removed, the next account in the Home screen's account list will become active.
If the account is active when it is removed, and connected to a dapp, the next account in the Home screen's account list will become active but the connection is severed. The user must re-connect to the Dapp.

Issue #13, opened 2022-01-20
As a Signer user I can Rename my account


ghost:
Any account in Signer can be named. Naming an account is performed within the Account Settings menu

Design: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Signer?node-id=1196%3A5966

![image](https://user-images.githubusercontent.com/97249036/172067221-35e6f47e-7c41-4fa1-b955-bf2a52ba10a5.png)

**Requirements**
A pencil button will appear at the top of the Account Settings screen. When clicked, the user is taken to the "Rename account" screen. This screen includes:

- Back button - directs the user back to the account settings screen with no changes made.
- Input field - where the user may type in their new account name The input field displays the existing account name by default.
- Update button - applies the value entered into the input field as the new account name and directs the user to the Account setting screen. The name change will propagate to all areas of Signer that display the account name including the Account list and account settings screen. The Update button is in a disabled state until the default text in the input field is modified.
- Cancel button - directs the user back to the account settings screen with no changes made.

**Input Validation**
This is an alphanumeric input.

Spaces are allowed.

Special characters are not allowed. If typed, display error message immediately "Please include alphanumeric characters only.

A maximum of 20 characters may be typed.

If user double clicks on the Account name on the Home screen... 
<img width="309" alt="image" src="https://user-images.githubusercontent.com/97249036/172065897-b81771f2-6943-41fa-9a98-84f9e39d47f3.png">
...automatically direct to user to the Rename account prompt screen:
![image](https://user-images.githubusercontent.com/97249036/172065840-34311429-7af9-43ec-a37e-4c25ee4713f3.png)

**Edge cases**
If the user modifies the input field text, but then types in the account's existing name, Update button becomes disabled again.

PR #1430, opened 2026-07-29
fix: WALLET-1364 — surface silent failures, replace assertions that cannot fail, correct drifted comments

ost-ptk:
Closes the cheap P2 findings of [WALLET-1364](https://make-software.atlassian.net/browse/WALLET-1364): make every silent failure visible, replace assertions that cannot fail, and correct comments that drifted from reality.

**Stacked on #1427** (`WALLET-1364-request-lifecycle`). GitHub will retarget the base to `release/2.7.0` once that merges. Review only this branch's 11 commits.

## Ticket items closed

| Item | Change |
| --- | --- |
| #13 | Both replica broadcasts stop swallowing every rejection. Only "Receiving end does not exist" is expected; anything else — a structured-clone failure, a throwing listener, a size limit — now surfaces instead of leaving an open popup silently stale. |
| #14 | `bringweb3`'s `getActivePublicKey` no longer ships `undefined` inside a success envelope. Locked vault or no active account now reports an explicit `null`. |
| #15 | The content script logs a delayed SDK response dropped with no active port, instead of discarding a just-signed deploy in silence. |
| #19 | A rejected password confirmation is logged rather than discarded, so it is no longer indistinguishable from a wrong password. |
| #12 (test half) | The `Pick` narrowing landed in #1427; this adds the test. TypeScript cannot catch this regression — excess-property checking does not apply to spreads, so re-adding `...state.windowManagement` compiles and silently re-broadcasts every in-flight dapp origin and tabId to every replica. |
| P3 rider | Three catch-and-rethrow sites in `sdk-methods.ts` keep the original cause. Dapp-facing messages stay generic; the typo `Desploy` is corrected. |
| Stale comments | Five corrections. Four other items from the ticket's list were already fixed by #1427 and are closed as moot. |
| Blind tests | Exact-count assertions become exact key-set assertions; the deadline storage keys move to a store-free module so a test stops re-declaring their literals. |
| P3 crypto | `validateString`/`validateBytes` report the argument type, never its value. |

## Things worth flagging to a reviewer

**The crypto fix was more than defense-in-depth.** With the old `'arg not valid, got:' + val` concatenation, a `Uint8Array` argument invokes `Array.prototype.toString`, rendering raw bytes as a comma-separated decimal list — a full plaintext dump of key material, not a harmless `[object Object]`. And `errorToMessage` now forwards `err.message` verbatim into `sagaError`, which is broadcast to popup state and rendered as a banner, so the conduit out of the background context is real. The throw is not reachable under the current TypeScript contract, but `decrypt-message` hands this validator a raw account secret key one type-confusion bug away.

**Two dapp-visible changes.** The `sign` rejection message changes from `Desploy json string parse error` to `Deploy json string parse error`. And `bringWeb3Events.getActivePublicKeyResponse`'s payload widens to `{ publicKey: string | null }` — an internal wire-shape change to the bringweb3 integration, whose only consumer (`content/bring.ts`) was updated in the same commit.

**#14 closes only half of P0.6.** The vault-lock check is in; the per-origin connection check is a separate P3 item and is deliberately out of scope. The characterization NOTE in `bringweb3.test.ts` records that distinction rather than claiming P0.6 is closed.

**#19 is log-only by decision.** A form-level message would need a new i18n key and a `locale:extract_pot` / `locale:update_translations` round trip through the catalogs, which is not worth it inside a cleanup PR. The user-facing half is a follow-up.

**One item needs a human check before merge.** `broadcastToReplicas`'s rejection guard matches an English substring. It was modelled on `keep-alive.ts`, but that is installed Chrome-only and fires on a slow ping, while this runs on every store tick on all three targets. Non-`Error` rejections are now normalized with `String(error)`, but whether Firefox and Safari reject with that exact wording is not something the code can prove. A short load-and-watch-the-console on a Firefox build would settle it.

**Two error-logging idioms now coexist.** `download-account-keys` logs `error.name` only; the new password-page log passes the whole error object, which the reachability analysis supports. Worth agreeing which is the house rule.

## Verification

`npm run ci-check` (`format:check` + `lint` + `tsc` + `knip` + `test:coverage`) exits 0 — 64 suites, 447 tests. `src/background/handlers/` holds its coverage floors (96.09% statements / 87.09% branches / 100% functions, floors 95/85/100).

Every new test was proven to fail before its fix. The exceptions are the two characterization tests over existing behaviour, which were instead proven load-bearing by mutating the production code and observing the failure.

The `crypto-security-reviewer` and a whole-branch review both ran. The whole-branch review produced three Important findings, all fixed in `4dff9097`: the cancel-path enumeration under-counted (`failRequestOnWindowError` is a third producer of `windowRequestResponded`, with no grace and outside the shared routine), the broadcast guard needed the non-`Error` normalization above, and two `nosemgrep` tokens sat two lines above their `console.error` where semgrep would not honour them.

[WALLET-1364]: https://make-software.atlassian.net/browse/WALLET-1364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -45,19 +45,43 @@ function deliveryFailedError(tabId: unknown, fallbackDelivered: boolean) {
...
-// 'responded' kills the real duplicate (a cancel racing a successful sign, where
-// the sign already set 'responded') without ever suppressing a first response.
+// It cannot be decided by `payload.cancelled` alone: `connectResponse` and
+// `switchAccountResponse` type their payload as a bare boolean, so that branch
+// does not generalise across the union. And it fails LOUD — anything not
+// recognised as benign is treated as a lost result, because the cost of a
+// missed warning is a line in a log while the cost of a missed error is a
+// signature the user produced and nobody ever received.
+function isBenignDuplicate(payload: unknown): boolean {
+  if (typeof payload === 'boolean') {
```
Comp0te, line 72: `isBenignDuplicate` returns `true` for every boolean payload, but only `false` is the throwaway shape here. `connectResponse` and `switchAccountResponse` are `createSdkAction<boolean>` (`src/content/sdk-method.ts:29,36`), and the wallet only ever synthesises `false` for them — `buildCancelResponse` at `src/background/handlers/cancel-requests.ts:45-48`. `true` is a genuine user approval.

The consequence is an asymmetry rather than one specific race: for these two methods a *lost approval* can never escalate, while the identical loss on `sign` always does. It takes the `console.warn('… dropped a duplicate cancel', …)` branch at :124-127 — a label describing the opposite of what happened — instead of the `console.error` + `sagaError` branch at :129-141. And both approval paths mutate wallet state before they send: `approve-connection/index.tsx:61-67` awaits `connectAccounts(...)`, `switch-account/content.tsx:169-178` awaits `changeActiveAccount(...)`. So where the drop is a real loss, the wallet lists the site as connected while the dapp was told the user rejected, and nothing on that path says so.

Nothing marks these requests responded at click time — `windowRequestResponded` has exactly three producers (`sdk-response-to-tab.ts:177-179`, `cancel-requests.ts:117`, `:211`) and the approval one fires when the message reaches the background, i.e. after the await.

One line, no wire change: `if (typeof payload === 'boolean') return payload === false;`. Worth pairing with `sdk-response-to-tab.test.ts:244-263` — as written it drives `connectResponse(true)` at a `'responded'` request and asserts no `console.error` and no `sagaError`, so a later fix has to delete an assertion rather than add one.

src/background/handlers/sdk-methods.test.ts
```diff
@@ -117,6 +119,98 @@ beforeEach(() => {
...
   } as any);
   selectIsConnectedMock.mockReturnValue(false);
+  consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+});
+
+afterEach(() => {
+  consoleErrorSpy.mockRestore();
+});
+
+describe('a requestId the wallet already registered', () => {
```
Comp0te: This `describe` block and the `'a requestId that is not storable'` block below it are each present twice in the file — `:129-215` and `:216-302` are byte-identical. Against the base, `grep -c "describe('a requestId"` goes 2 → 4 and the `it(` count goes 44 → 48.

Both copies run, so `jest -t "a replayed finished requestId is refused before anything is dispatched"` reports "2 passed". The cost is that a future edit to the duplicate-requestId contract applied to one copy leaves the other asserting the old one — and no repo gate catches duplicated describe/it names, so CI stays green either way.

Deleting either copy fixes it.

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -88,14 +112,34 @@ export async function handleSdkResponseToTab(
...
+      );
+    } else {
+      console.error(
+        'sdk-response-to-tab: dropped a completed response — the result was lost',
+        identifiers
+      );
+      // The log alone is not enough here: the user approved something, the
+      // dapp will never hear about it, and nothing else on this path says so.
+      // No payload, no origin — appEvents is broadcast to every replica.
+      store.dispatch(
```
Comp0te: This `sagaError` is a new independent dispatch — on the base, `sagaError` is constructed only inside `deliveryFailedError` (dispatched at :119 and :145). It's a new user-visible surface, and the description's "Ticket items closed" table doesn't mention it.

`saga-error-banner.tsx:87` renders `{error.source}: {error.message}` verbatim — only the "Something went wrong" heading at :84 is translated — so the user gets the untranslated English `sdk-response-to-tab: A completed response for tab 3 was dropped as a duplicate; the site was not told`, naming an internal integer tabId that identifies no dapp and suggests no next step. `SagaErrorBanner` is mounted in signature-request (`index.tsx:73`) and connect-to-app (`index.tsx:71`), so it can land on top of a replacement approval screen — the UX concern item #9 raises about this banner on this path. (The `tab ${tabId}` phrasing isn't itself new — base :43-44 has it — but this is a new place it reaches the user.)

Either call it out in the description and agree the copy, or keep this path log-only for now, consistent with the #19 decision in this same PR to defer the user-facing half.

src/background/redux/get-main-store.ts
```diff
@@ -104,6 +94,25 @@ const selectPopupState = (state: RootState): PopupState => {
...
+// Only "no receiver" is expected: `runtime.sendMessage` delivers to every
+// extension context except the sender, so with no popup open Chrome rejects
+// with "Receiving end does not exist." Everything else — a structured-clone
+// failure, a throwing listener, a message-size limit — means an OPEN replica
+// just missed an update and is now silently stale, so it must be visible.
+// Same idiom as keep-alive.ts.
+function broadcastToReplicas(message: BackgroundEvent, source: string): void {
+  runtime.sendMessage(message).catch((error: unknown) => {
+    const text = error instanceof Error ? error.message : String(error);
+    if (text.includes('Receiving end does not exist')) {
```
Comp0te, line 106: Is the `'Receiving end does not exist'` wording the same on Firefox and Safari? This runs on every store dispatch (subscriber at :187-242, broadcasting at :192-195 and :202-205) on all three targets — one `background` entry at `webpack.config.js:245`, and `manifest.v2.json:43-46` loads the same bundle — with no `isChromeBuild` / `isSafariBuild` gate. The precedent the comment at :102 names is effectively Chrome-only: `keep-alive.ts` gates alarm registration at :18 and `initKeepAlive` at :93, so `keepAlive()` only ever observes Chrome's wording.

If it differs at all off-Chrome, the miss lands on the normal idle steady state — popup closed, nothing listening — and every dispatch logs `popupStateUpdated broadcast failed:`. That buries the stale-replica errors this code was added to surface, reproducing on the benign case the "trains a reader to skip past it" problem the sibling change to `sdk-response-to-tab.ts` avoids on the dangerous one.

I couldn't settle the off-Chrome wording from source. Loading `./build/firefox` and `./build/safari` with no popup open and driving any dispatch (focusing a window fires `updateOrigin`) would — expecting zero `broadcast failed` lines. Otherwise: also match `'Could not establish connection'`, or pin the expected string per target behind the existing build flags.

Comp0te (commented):
Reviewed the diff against WALLET-1364 and the PR description, focusing on the new failure-surfacing paths and the tests added alongside them. The through-line: the escalate-versus-stay-silent decisions this PR introduces are right in shape, but a few classify the wrong way or aren't pinned by an assertion that could fail — plus one new user-visible banner that isn't in the description. Details inline.

src/background/redux/get-main-store.test.ts
```diff
@@ -82,3 +86,154 @@ describe('getExistingMainStoreSingletonOrInit — keysDoesExist preload derivati
...
+describe('selectPopupState broadcast — replica privacy narrowing', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    storageSet.mockResolvedValue(undefined);
+    storageRemove.mockResolvedValue(undefined);
+    runtimeSendMessage.mockResolvedValue(undefined);
+  });
+
+  it('broadcasts exactly the expected slices, with windowManagement narrowed to windowId', async () => {
+    const store = await initWithKeysSnapshot(undefined);
```
Comp0te, line 133: These tests can't observe a leak in the two slices with the largest secret surface. Across both broadcast tests (:132-149 and :151-172) the assertions are `Object.keys(payload).sort()` at :147, `payload.windowManagement` deep-equal at :148, and `JSON.stringify(...)` not containing `'dapp.example'` at :169-171 — nothing reads inside `payload.keys` or `payload.session`. Both init with `initWithKeysSnapshot(undefined)` (:133, :152), so `passwordHash` / `passwordSaltHash` / `keyDerivationSaltHash` and `session.encryptionKeyHash` are all `null` in the fixture anyway.

Replacing the narrowing at `get-main-store.ts:63-70` with `keys: state.keys, session: state.session` survives all of them: `PopupState` types both slices as the full state type (`types.d.ts:21-22`) so tsc stays green, `KeysState`'s top-level field names are exactly those in the narrowed literal so the asserted key set is unchanged, and neither slice contains an origin. That broadcasts the password hash, both salts and the encryption key hash to every replica on every store change. Nothing else pins it either — `private-state-broadcast.test.ts`, `handlers/private-state.test.ts` and the slice reducer tests never read a `popupStateUpdated` payload.

Seeding with real hash material (`initWithKeysSnapshot({ passwordHash: 'PH', passwordSaltHash: 'PS', keyDerivationSaltHash: 'KD', keysDoesExist: true })` plus an `encryptionKeyHashCreated` dispatch), then asserting `payload.keys` deep-equals the nulled shape and that `JSON.stringify(payload)` contains none of those markers — same style as the `not.toContain('dapp.example')` line — closes it.

src/background/redux/storage-keys.ts
```diff
@@ -0,0 +1,15 @@
...
+// sagas read on `startBackground` to resume the residual delay.
+//
+// They live here, apart from get-main-store, precisely so a consumer (or a test)
+// can read the real values without importing the Redux store.
+//
+// IMPORTANT: once shipped, these key strings are immutable. Existing installs
+// persist deadlines under exactly these strings; renaming them would strand the
+// old entries and break resume-after-restart for already-locked-out users.
+// See docs/architecture/storage-keys.md for the full key inventory.
+export const LOGIN_RETRY_LOCKOUT_DEADLINE_KEY = 'q9Tf3Lm4pRxVne';
```
Comp0te, line 14: The comment three lines up says these strings are immutable once shipped, but nothing executable pins the values. `vault-sagas.test.ts` imports both constants (:15-16) and uses them symbolically on both sides of every assertion — fixtures at :114, :134, :223, :253 and expectations at :128, :199, :278, :303 — so the fixture key and the asserted key are the same symbol and no assertion can observe the string. This PR removed the last test-side copy: the `jest.mock('@background/redux/get-main-store', …)` block that re-declared them at `vault-sagas.test.ts:57-63` on the base is gone. The literals now appear only here and in `docs/architecture/storage-keys.md:54-55`, with no doc/code parity check in `src`, `scripts` or CI.

So changing `'q9Tf3Lm4pRxVne'` to `'q9Tf3Lm4pRxVnf'` ships fully green. The blast radius is narrower than "immutable" suggests — `vault-sagas.ts:227-235` and `:375-383` fall back to a recomputed deadline rather than failing open — but the persisted entry is stranded and the residual deadline is silently replaced by a fresh full one, which is what the comment says must not happen.

A four-line value-pinning test next to the module (`expect(LOGIN_RETRY_LOCKOUT_DEADLINE_KEY).toBe('q9Tf3Lm4pRxVne')`) with a pointer to the immutability rule is the only assertion shape that catches a rename.

src/background/redux/get-main-store.test.ts
```diff
@@ -82,3 +86,154 @@ describe('getExistingMainStoreSingletonOrInit — keysDoesExist preload derivati
...
+
+    expect(consoleErrorSpy).toHaveBeenCalledWith(
+      expect.stringContaining('popupStateUpdated'),
+      expect.objectContaining({
+        message: expect.stringContaining('DataCloneError')
+      })
+    );
+  });
+
+  it('stays silent when the rejection is a bare string, not an Error', async () => {
```
Comp0te, line 224: This test asserts a bare-string rejection is real and must stay silent, and `get-main-store.ts:105` normalises accordingly with `error instanceof Error ? error.message : String(error)`. The sibling the comment at :102 names as the source idiom didn't get the same treatment: `keep-alive.ts:82-88` still requires `error instanceof Error` and otherwise falls through to `console.error('KeepAlive error:', error)`. `keep-alive.test.ts:94-104` and `:106-117` both pass `Error` rejections, so nothing fails.

Under exactly the condition this test asserts, an unlocked wallet with no UI open logs `KeepAlive error:` every 30 seconds (`periodInMinutes: 0.5` at `keep-alive.ts:43`; the ping runs only while unlocked, :67-71) — the recurring false error that buries real ones. The two sites now disagree about whether a bare-string rejection can occur, and only one can be right.

Either apply the same normalisation at `keep-alive.ts:82-87` and add the bare-string case next to `keep-alive.test.ts:94`, or drop this test if a bare-string rejection can't actually occur. (`keepAlive` is Chrome-only — gates at :18 and :93 — so the two sites differ in which targets they run on, too.)

src/apps/popup/pages/password-protection-page/index.tsx
```diff
@@ -129,9 +129,17 @@ export const PasswordProtectionPage = ({
...
-            // and now that the field is read-only-while-submitting, frozen too.
-            .catch(() => setIsSubmitting(false));
+            // Without this a rejected onClick leaves the page stuck submitting
+            // — and now that the field is read-only-while-submitting, frozen
+            // too. The error itself is logged rather than discarded: silently
+            // resetting the button is indistinguishable from a wrong password.
+            .catch((error: unknown) => {
+              // The password is in scope but is deliberately not referenced
+              // here — only a static message and the error object are logged.
+              // nosemgrep: cw-logging-secrets
```
Comp0te: This `nosemgrep` doesn't suppress anything today. `cw-logging-secrets` (`.semgrep.yml:90-103`) is a `pattern-regex` over `privateKey|secretKey|secretPhrase|mnemonic|seed|vault|password|passphrase|recoveryPhrase` with no `(?i)` flag — `cw-no-hardcoded-secrets` at :149 does set one, so the omission looks deliberate — and neither the capital-P string `'Password confirmation failed:'` nor the bare identifier `error` on :140 matches. Same at `src/content/index.ts:61`, whose rule list contains neither `signatureHex` nor `encryptedMessage`.

What it does do is permanently exempt the line. `password` is in scope at :140 through the closure over the `onSubmit` destructure at :100, and `npm run semgrep` is a blocking gate — so this is now the one line in that function where the gate can't fire, at the call site nearest the secret. The annotation makes nothing safer now; it only removes the check that would catch a later edit.

Removing both costs nothing. If either was added because a rule genuinely fired, naming which rule and on what — and narrowing to it — would keep the guard.

src/content/index.ts
```diff
@@ -52,7 +52,22 @@ async function handleSdkMessage(message: unknown) {
       case sdkMethod.encryptMessageError.type:
       case sdkMethod.getActivePublicKeySupportsResponse.type:
         // route the delayed response over the private port (not a window event)
-        activePort?.postMessage(message);
+        if (activePort == null) {
```
Comp0te, line 55: This drop log — the whole of item #15 — isn't exercised by any test. Nothing in the repo imports `src/content/index.ts`: a tree-wide search for `establishSdkPort`, `SDK_REQUEST_TYPES`, `handleSdkMessage` and `cleanupEventType` returns only this file plus two comment mentions (`background/handlers/redux-actions.ts:215`, `background/handlers/sdk-response-to-tab.ts:98`), including under `e2e-tests/` and `utils/`. `src/content/sdk.test.ts` loads `./sdk` only and drives a fake `MessagePort` — the page-world side, never the content-script side.

So reverting :55-68 to `activePort?.postMessage(message);` fails nothing: with a null port the two are behaviourally identical apart from this `console.error`, which nothing asserts. The only record that an approved-and-signed deploy was discarded is a line nobody observes — and the constraint keeping it from later growing to include `signatureHex` is a comment plus a `nosemgrep` token, not an assertion.

The loader pattern at `src/content/sdk.test.ts:17-47` transfers unchanged: stub `global.window`/`global.document`, mock the polyfill's `runtime`, `require('./index')`, dispatch `cleanupEventType` to null the port, then invoke the captured `runtime.onMessage` listener with `signResponse({ cancelled: false, signatureHex: 'deadbeef' })` and assert both that `console.error` got the type and requestId and that the call args don't contain `'deadbeef'` — same shape as `sdk-response-to-tab.test.ts:228`.

ost-ptk:
Thanks for the pass — went through all nine. Six are addressed in six new commits (`eed9082b..37b2677b`); three are deliberately left, reasons below. One of your claims turned out to be wrong on verification, details at the end.

## Addressed

**`isBenignDuplicate` returning `true` for every boolean** — `a1ae29fc`. Now `return payload === false`. Confirmed your enumeration: every `false` producer is a reject or a synthesised cancel (`cancel-requests.ts:46,48`, `approve-connection:72`, `select-account:78`, `switch-account/index:34`, `unconnected-accounts-list:108`), and `disconnectResponse`/`isConnectedResponse` are direct handler return values that never route through the dedupe, so they cannot be mis-escalated. The test at `sdk-response-to-tab.test.ts` was rewritten to drive `connectResponse(false)`, and a new test drives `switchAccountResponse(true)` and asserts the error branch — it fails if the fix is reverted.

**Duplicated `describe` blocks** — `eed9082b`. Deleted the second copy. `describe('a requestId` back to 2, `it(` back to 44.

**The new `sagaError`** — `a1ae29fc`, kept **log-only**, as you suggested. Consistent with the #19 call in this PR: the copy names an internal `tabId` that identifies no dapp, and the banner renders `message` verbatim and untranslated over the approval screens. The user-facing half is a follow-up with a real i18n key.

**`keep-alive.ts` disagreeing with the new guard** — `1f7063fc`. Applied the same `error instanceof Error ? error.message : String(error)` normalisation and added the bare-string case next to the existing one. Agreed that two sites disagreeing about whether a bare-string rejection can occur is the actual defect.

**`nosemgrep` suppressing nothing** — `6fc29324`. Verified empirically rather than by reading the regex: stripped every `cw-logging-secrets` token from a copy of `src/` and rescanned. The rule fires on **exactly one line in the tree** — `get-main-store.ts:239`, `console.error('Persist encrypted vault failed: ', e)`, lowercase `vault`. That token predates this branch and stays. All six added by this branch match nothing and are removed, prose kept. You flagged two; there was a third at `get-main-store.ts:111`. `npm run semgrep` stays at 0 findings.

**Storage-key value pinning** — `f860ae28`. New `storage-keys.test.ts`, three assertions, with a note not to "fix" a failure by updating the expected string.

## Deliberately not done

**Cross-browser wording of `'Receiving end does not exist'`** — still open, and still needs the load-and-watch you describe. I did not add a second substring match speculatively; if the Firefox/Safari check comes back dirty, matching `'Could not establish connection'` as well is the cheap fix, since that half of the message is what actually varies.

**`get-main-store.test.ts` blind to `keys` / `session`** — correct and worth closing, but it is a new test rather than a correction to one, so it is out of this PR's scope. Happy to take it here if you would rather not carry it forward.

**Item #15 untested** — same call. `src/content/` is outside `collectCoverageFrom` entirely, so this is standing up a first test for that entry point, not filling a gap in one.

## One correction to the review

> the persisted entry is stranded and the residual deadline is silently replaced by a fresh full one

This does not happen, and I had repeated it in the test comment before checking. Both resume paths fall back to the same formula they armed with: `vault-sagas.ts:234` recomputes `loginRetryLockoutTime + LOCK_VAULT_TIMEOUT`, which is `:224` verbatim, and `:383` mirrors `:386`. Both operands are themselves persisted slices. So a renamed key yields an **identical** deadline — the residual delay survives.

The real cost of a rename is an orphan key in `storage.local` forever, plus divergence in the one case the recomputation cannot reproduce: an operand that changed since arming (`LOCK_VAULT_TIMEOUT` bumped by an update, or the user editing the auto-lock timeout mid-countdown — the stored deadline is the only record of what was actually promised). The pin is still worth having; the justification is corrected in `37b2677b`.

## Found while fixing #1 — needs your call

`select-account/index.tsx:64-71` sends `connectResponse(true)` **from the render body**, unguarded:

```tsx
if (isActiveAccountConnected) {
  sendSdkResponseToSpecificTab(sdkMethod.connectResponse(true, { requestId }), requestTabId)
    .then(() => { closeCurrentWindow(); });
}
```

The first send marks the request `responded`, which broadcasts `popupStateUpdated`, which re-renders the page, which sends again. The second copy hits the dedupe — and after `a1ae29fc` that is the `console.error('the result was lost')` branch, although the first copy was delivered and nothing was lost. Same on a double-click of Approve/Switch: neither button disables while awaiting.

Not a regression — on `release/2.7.0` this path already logged `console.error`; this PR introduced the warn/error split and `a1ae29fc` moves `true` back to error. But it does undercut what the split is for. The request status alone cannot distinguish "duplicate of a delivered `true`" from "`true` racing a cancel", so the classifier has to keep failing loud; the fix belongs in that page (an effect plus a sent-guard, or disabling the button). Recorded as a `KNOWN NOISE` note in `37b2677b`.

**Do you want a separate ticket for that, or should it ride along in this PR?** It is a side effect in a render body, so it can also double-fire without the broadcast — I would rather not fix it quietly inside a cleanup PR, but I will if you prefer to keep it together.

## Verification

`npm run ci-check` exits 0 — 69 suites, 534 tests. `src/background/handlers/` holds its floors at 96.71% statements / 89.33% branches / 100% functions (floors 95/85/100). `npm run semgrep` — 0 findings. Every new test was proven to fail with its production change reverted.

src/background/redux/get-main-store.ts
Comp0te: Still open at `37b2677b` — `:106` is unchanged and the substring match still runs ungated on all three build targets.

One correction to my original comment: the `keep-alive.ts` precedent is indeed Chrome-only, but the `isChromeBuild` gates sit at `keep-alive.ts:18` and `:93` (alarm registration and `initKeepAlive`), not on the lines I pointed at — `keepAlive()` simply never runs off Chrome, so it has never observed a non-Chrome rejection string either way.

And to be explicit that this is conditional: the impact turns entirely on the off-Chrome wording, which nothing in the tree can settle, since the polyfill short-circuits to the native `browser` object on Firefox and Safari rather than normalising `chrome.runtime.lastError`. Loading `./build/firefox` and `./build/safari` with no popup open and driving a single store dispatch settles it — if the wording differs, the idle steady state emits a `broadcast failed:` line on every dispatch, which is what the change was added to keep readable.

src/background/redux/get-main-store.test.ts
Comp0te: Still open at `37b2677b`. Both broadcast tests still initialise with `initWithKeysSnapshot(undefined)` (`:133` and `:152`), so `keys.passwordHash` / `passwordSaltHash` / `keyDerivationSaltHash` and `session.encryptionKeyHash` are all `null` in the fixture, and the three assertions — `:147`, `:148`, `:169-171` — never read inside `payload.keys` or `payload.session`.

The concrete mutation this cannot catch: in `selectPopupState`, replace `keys: { passwordHash: null, passwordSaltHash: null, keyDerivationSaltHash: null, keysDoesExist: state.keys.keysDoesExist }` with `keys: state.keys`. `tsc` stays green, because `PopupState` types the slice as the full `KeysState` (`types.d.ts:21-22`); the top-level key set is unchanged, so `:147` passes; `windowManagement` is untouched, so `:148` passes; and the payload still contains no `dapp.example`, so `:169-171` passes. All four assertions hold while the password hash, both salts and the encryption key hash go to every replica on every store change. No other test pins the narrowing — `private-state-broadcast.test.ts`, `handlers/private-state.test.ts` and the slice reducer tests never read a `popupStateUpdated` payload.

src/content/index.ts
Comp0te: A correction to my original comment first: "Nothing in the repo imports `src/content/index.ts`" is wrong. `src/content/sdk-channel.test.ts:114` does `require('./index')` — I searched for the path fragment and missed the relative require. Apologies.

The gap itself is unchanged at `37b2677b`, though. That test asserts only that `SDK_REQUEST_EVENT` has no window listener (`:120-122`), and its `webextension-polyfill` mock stubs `runtime.onMessage.addListener` as a bare `jest.fn()` (`:19`) that nothing ever invokes — so `handleSdkMessage` is never entered and the drop log at `:61-65` is never reached. Reverting the guard at `:55-67` to `activePort?.postMessage(message);` still fails nothing.

Comp0te (approved):

PR #1431, opened 2026-07-29
refactor: WALLET-1364 — render api-supplied icons as img and drop the casper-assets host

ost-ptk:
## What

Contract, token and NFT icons that arrive in API responses are now rendered as a plain ``
instead of being fetched and inlined into the extension DOM — which is what finally makes the
`casper-assets.s3.amazonaws.com` host permission and `connect-src` entry genuinely unused, and lets
them be dropped.

The CSP is also single-sourced: `src/csp.json` now holds the shared directives and the `connect-src`
host list, and both consumers derive from it — `getCSP()` in `webpack.config.js` (which builds the
manifest CSP) and `setCSPForSafari()` in `src/utils.ts` (which injects the runtime `<meta>` policy,
because Safari ships no manifest CSP).

Jira: https://make-software.atlassian.net/browse/WALLET-1364

## Why the host looked unused but wasn't

`SvgIcon` is `react-inlinesvg`: it **fetches** the file over HTTP and injects the markup into the
page. A remote `.svg` therefore falls under `connect-src`, not under `img-src` — and `img-src` is
`https: data:`, permissive enough that the host genuinely appeared to have no consumer. Removing it
from `connect-src` broke 27 popup e2e tests with
`[CSP] connect-src blocked https://casper-assets.s3.amazonaws.com/contracts/*.svg`.

The host permission mattered for the same reason: the bucket sends no CORS headers (GET with an
`Origin` returns no `Access-Control-Allow-Origin`; the preflight 403s), so an extension-page `fetch`
needed the permission to bypass CORS. An `` needs neither.

## What changed

- **New `RemoteIcon`** (`src/libs/ui/components/remote-icon/`) — renders an API-supplied url as an
  `` with an `onError` fallback. The fallback deliberately renders through `SvgIcon` rather
  than ``: `img-src` has no `'self'`, so a bundled `chrome-extension:` asset would not load in
  an ``.
- **Eight call sites migrated.** The branch condition moved from "does the url end in `.svg`" to
  "is the url remote or local", via the existing `hasHttpPrefix`. Bundled `assets/icons/*` keep
  going to `SvgIcon`, which needs inlining for its `fill="currentColor"` theming.
- **A regression guard.** `assertLocalIconSrc` logs `[SvgIcon] …` whenever `SvgIcon` receives an
  http(s) src, in dev builds and in the e2e build (`NODE_ENV=production` **plus** `TEST_ENV=true`),
  and `e2e-tests/fixtures.ts` now fails the popup suite on that marker alongside `[CSP]`. It earned
  its keep immediately: it caught two call sites the initial inventory missed, because they passed a
  remote url to `SvgIcon` unconditionally rather than through an `endsWith('.svg')` branch.

## Side effects worth a reviewer's attention

- **Safari's `script-src` changed.** Sourcing `baseDirectives` from `src/csp.json` means Safari's
  runtime `<meta>` policy now carries `'self' 'wasm-unsafe-eval'` where it previously had `'self'`
  only. This is intended — `libsodium-wrappers-sumo` compiles WebAssembly, so Safari was arguably
  under-permissioned before — and it is now pinned by a test (`src/utils.test.ts`) so a future edit
  to `src/csp.json` cannot change Safari's policy unnoticed. The `frame-ancestors` that came along
  with it is inert inside a `<meta http-equiv>` and is present for parity only.
- **A latent bug is fixed as a side effect.** `connect-src` has no wildcard, so a token icon hosted
  anywhere other than the listed hosts was already failing to render silently. Those now work,
  since `img-src https:` covers them.
- **`img-src https: data:` is unchanged but more load-bearing.** It cannot be narrowed without
  breaking dapp favicons and NFT media, which are arbitrary-origin by product design.

## Permission diff

Removed, symmetrically across all three manifests: `https://casper-assets.s3.amazonaws.com/*` —
from `host_permissions` in `manifest.v3.json` and from `permissions` in `manifest.v2.json` and
`manifest.v2.safari.json`, plus the matching `connect-src` entry in the Chrome/Firefox build-time
CSP and the Safari runtime `<meta>` CSP.

Added: none.

## Verification

- `npm run ci-check` — 71 suites, 537 tests, green
- popup e2e — 46 passed, 1 skipped, 0 failed (was 27 failed)
- onboarding e2e — 6 passed
- `build:chrome`, `build:firefox`, `build:safari` — all succeed


src/libs/ui/components/account-info-icon/account-info-icon.tsx
```diff
@@ -57,15 +55,16 @@ export const AccountInfoIcon = ({
   }
 
   if (iconUrl) {
-    return iconUrl.endsWith('.svg') ? (
-      <SvgIcon src={iconUrl || ''} alt={accountName || ''} size={size} />
-    ) : (
-      <LogoImg
+    return hasHttpPrefix(iconUrl) ? (
```
Comp0te: The routing predicate flipped direction here, and the non-http branch is now the inlining one.

Before: `iconUrl.endsWith('.svg') ? SvgIcon : LogoImg` (`:60` on the base) — anything that wasn't a `.svg` path went to an inert ``. After: `hasHttpPrefix(iconUrl) ? RemoteIcon : SvgIcon` — everything that isn't `http(s)` now goes into `SvgIcon`, i.e. into react-inlinesvg. Same flip at `avatar.tsx:126`, `deploy-container.tsx:23`, `token-swticher-row.tsx:29`, `token-plate.tsx:71`.

`iconUrl` is API-supplied and this component renders on both signing screens — `unknown-contract-action-rows.tsx:344-352` passes `action.iconUrl`, and `eip712-display-row.tsx:38-43` passes `row.contractPackage.iconUrl` through `ContractInfoRow` to `common.tsx:333-339`. So the default branch for an untrusted value is now the inliner rather than the inert element, which is the opposite of the reason `RemoteIcon`'s docstring gives for its own existence.

One case is confirmable without leaving the repo: `hasHttpPrefix` is `/^https?:\/\//` at `utils.ts:18-20`, with no `/i`. `HTTPS://host/icon.svg` therefore takes the `SvgIcon` branch — and `assertLocalIconSrc` can't flag it, because `assert-local-icon-src.ts:17` guards with that same predicate and only fires when it returns true. The misroute is precisely the case the guard is blind to, so the `[SvgIcon]` marker stays silent and the icon just fails to render.

The part I can't settle from here: if react-inlinesvg inlines `data:image/svg+xml,…` and raw `<svg …>` strings without fetching, as its docs describe, such a value would reach the popup DOM with no connect-src gate at all. Is either shape reachable from `iconUrl` or `branding.logo` in the API responses?

Making the `SvgIcon` branch require a bundled-asset shape — an `assets/`-prefixed path, or a parsed-URL allowlist — rather than defaulting every non-http value into it would close both; parsing rather than prefix-matching also takes care of the case-sensitivity gap.

src/utils.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  it('locks every fetch-directive default down', () => {
+    const content = getSafariCspContent();
+
+    expect(content).toContain("default-src 'none'");
+    expect(content).toContain("object-src 'none'");
+    expect(content).toContain("base-uri 'none'");
+    expect(content).toContain("form-action 'none'");
+  });
+
+  it('carries every connect-src host from csp.json and nothing else', () => {
```
Comp0te: This assertion can't fail for the thing its name promises. `getSafariCspContent()` builds the string at `utils.ts:176` with `cspConfig.connectSrc.join(' ')`, and the test imports the same `./csp.json` at `:1-2`, so `connectSrc.split(' ')` is compared against its own source.

To be fair to it, it isn't vacuous — it does pin the template literal in `utils.ts` against `csp.json`, so a host hardcoded into the literal, a dropped host, a reordering, or `connect-src` ceasing to be the final directive would all fail it. What it can't do is validate the contents of `csp.json`, and that's now the artefact worth pinning, since the same array feeds all three targets through `webpack.config.js`. Drop a host and every mainnet API call breaks in all three builds; add a typo'd origin and all three widen. Green either way.

`:48` already does this the right way for script-src, against a literal — and the only real host assertion in the file is the negative at `:72`.

Comp0te (commented):
Reviewed this against the two WALLET-1364 items — the casper-assets removal across the manifests and both CSP strings, and the single-sourcing of the Safari policy. The removal itself checks out in all five places, and the shared `csp.json` is a clear improvement on the drift it replaces.

The comments below cluster in three places: the direction of the new local-vs-remote routing predicate and what now falls into its default branch, the half of the CSP that didn't get single-sourced, and the fact that most of the converted call sites and both of RemoteIcon's runtime behaviours aren't observable by any test — so a revert of any one of them would ship green.

src/utils.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(scriptSrcDirective).toBe("script-src 'self' 'wasm-unsafe-eval'");
+  });
+
+  it('never grants eval or inline script', () => {
+    expect(getSafariCspContent()).not.toContain("'unsafe-eval'");
+    expect(getSafariCspContent()).not.toMatch(/script-src[^;]*'unsafe-inline'/);
+  });
+
+  it('locks every fetch-directive default down', () => {
```
Comp0te: These four use `toContain` against the whole policy string, which doesn't pin a directive: `"default-src 'none' https:"` still contains `"default-src 'none'"`, so widening any of them in `csp.json` passes. There's no exhaustiveness assertion in the block either, so a newly appended directive is matched by nothing at all.

`:44-48` right above does it correctly — split on `'; '`, exact `toBe`. The same treatment here would close it.

Worth flagging that these directives now reach the Chrome and Firefox manifests as well as the Safari `<meta>`, so a silent relaxation lands in all three at once.

webpack.config.js
```diff
@@ -90,7 +92,10 @@ const getCSP = () => {
...
+  // `baseDirectives` and `connectSrc` are shared with the Safari runtime <meta>
+  // policy in src/utils.ts — see src/csp.json. Only the style arm differs per
+  // target, so it stays here.
+  const csp = `${cspConfig.baseDirectives}; ${styleDirectives}; img-src https: data:; media-src https: data:; connect-src ${cspConfig.connectSrc.join(' ')}`;
```
Comp0te: Only `baseDirectives` and `connectSrc` ended up shared. `img-src https: data:; media-src https: data:` is still written out verbatim in both consumers — here at `:98` and again at `src/utils.ts:176` — along with the assembly template around it, and nothing compares the two outputs: `getCSP` is a non-exported local in a CommonJS config jest never loads, and `src/utils.test.ts` only checks the Safari string against `csp.json`.

So the drift mode the ticket was filed for is still open in the unshared half: narrow `media-src` or add `frame-src` here, and the Safari runtime `<meta>` keeps the old value indefinitely with nothing failing.

Both literals are byte-identical today and CSP directive order doesn't matter, so moving them into `csp.json`'s `baseDirectives` would leave the style arm as the only per-target difference.

src/manifest.v3.json
```diff
@@ -36,7 +36,6 @@
     "https://cspr-api-gateway.dev.make.services/cspr-node-proxy-rpc-dev-condor/*",
```
Comp0te, line 36: Anchoring on the line below the deletion — the removal itself left no line to comment on.

The host removal landed in five places (three manifests plus both CSP strings), and only the CSP side has a guard: `utils.test.ts:71-73` asserts `casper-assets` is gone from the Safari string, and since both strings now derive from `csp.json`, that covers both. Nothing asserts it's gone from any of the three manifests.

Grepping the branch for `casper-assets` turns up three files, all tests using it as a sample url or as that one negative — `utils.test.ts:6` and `:71-73`, `remote-icon.test.tsx:10`/`:18`, `assert-local-icon-src.test.ts:25`/`:30`.

Re-adding the entry to `host_permissions` — a merge resolution, a copy-pasted permission block — would undo this silently, and on Chrome a host-permission addition can disable the extension for existing users until they re-approve. `webpack-config.test.ts:113-123` already parses the built manifest through the real CopyWebpackPlugin transform, so the assertion has a home and costs about a line per variant.

src/libs/ui/components/remote-icon/remote-icon.tsx
```diff
@@ -0,0 +1,71 @@
...
+  alt,
+  title,
+  fallbackSrc,
+  className
+}: RemoteIconProps) => {
+  const [hasError, setHasError] = useState(false);
+
+  // Rows are recycled across different tokens and contracts. Without this the
+  // error latched from the previous url would hide a perfectly good new icon.
+  useEffect(() => {
```
Comp0te, line 58: Neither of this component's two runtime behaviours is observable by the current setup, so both can regress green.

`jest.config.js:3` is `testEnvironment: 'node'`, and there's no `jsdom` or `jest-environment-jsdom` in `package.json` — `@testing-library/jest-dom` at `:141` supplies matchers, not an environment, and `@testing-library/react` isn't there. `remote-icon.test.tsx` is three `renderToStaticMarkup` cases, none of which touches `onError` or changes `src`; `resolve-icon-src.test.ts` passes `hasError` in as a literal at `:6, :15, :26, :33, :41, :50, :60` and never watches the component set it.

That leaves the `onError → setHasError(true)` chain at `:67` and this reset at `:47-49` untested. Losing the first means a dead icon url renders whatever the browser does with a failed `` (`:65` emits `alt=""`) instead of the fallback, on every Activity row. Losing the reset latches the error onto a recycled row, so every later token or deploy reusing that instance renders the fallback or nothing.

Hoisting the transition the way `resolveIconSrc` was already hoisted — something like `nextIconState(prev, { src })` — would make both rules testable under `testEnvironment: 'node'` without pulling in jsdom.

src/libs/ui/components/deploy-plate/components/deploy-container.tsx
```diff
@@ -1,32 +1,35 @@
...
 export const DeployContainer = ({
   children,
   iconUrl,
   title
 }: DeployContainerProps) => {
   return (
     <AlignedFlexRow gap={SpacingSize.Medium} flexGrow={1}>
-      {iconUrl.endsWith('.svg') ? (
-        <SvgIcon src={iconUrl || ''} alt={title || ''} size={24} />
+      {hasHttpPrefix(iconUrl) ? (
```
Comp0te: This branch and the two like it at `deploy-details/components/common.tsx:109` and `nft-details/content.tsx:347` have no unit test — the only component tests on the branch are `remote-icon`, `resolve-icon-src` and `assert-local-icon-src`.

For two of the three the e2e `[SvgIcon]` guard can't cover the gap either. `DeployContainer` is reached via the Activity tab (`home/index.tsx:153-163`) or the token-details deploy lists, and `common.tsx:109` via deploy details; no spec under `e2e-tests/popup` opens either. `nft-details/content.tsx:347` is better off — `nft-transfer.spec.ts:16-27` does open `NftDetailsPage` — so there it's unit coverage that's missing rather than both.

Reverting any of these to inline-SVG rendering now fetches from a host that has lost both its host permission and its connect-src entry, so the icons disappear and the full gate stays green. A `renderToStaticMarkup` test per call site, as `remote-icon.test.tsx` already does — an `https://` src yields `<img` and never `<svg`, an `assets/icons/...` path yields the SvgIcon branch — would catch it.

src/apps/popup/pages/transfer/components/token-row.tsx
```diff
@@ -34,20 +29,18 @@ export const TokenRow = ({
...
-  const tokenIconFormat = token?.icon?.split('.').pop();
-  const isTokenIconJPG = tokenIconFormat === 'jpg';
-
   return (
     <Container onClick={handleSelect}>
       <AlignedFlexRow gap={SpacingSize.Medium}>
-        {isTokenIconJPG ? (
-          <LogoImg
-            src={token?.icon || undefined}
+        {token?.icon && !hasHttpPrefix(token.icon) ? (
```
Comp0te: The `LogoImg` this replaces carried `border-radius: ${({ theme }) => theme.borderRadius.hundred}px`, and `theme-config.ts:35` has `hundred: 100`, so it was circular. `RemoteIcon`'s `Img` at `remote-icon.tsx:20-25` sets only width, height and `object-fit`, and nothing here passes a `className`, so the radius is gone.

Scoping it precisely: the old branch was `isTokenIconJPG ? LogoImg : SvgIcon`, so only `.jpg` icons were ever round — png and svg went to `SvgIcon` and were square already. For jpg token logos in the "Select token and account" list this is a visible circular → square change, and nothing in the repo would notice it.

Of the six `LogoImg` definitions removed across this change, this is the only one that carried a border-radius; the other five were width/height only and `RemoteIcon`'s `Img` covers them. So a `className` from here, or a radius prop on `RemoteIcon`, is all it needs.

src/libs/ui/components/remote-icon/remote-icon.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  if (resolved == null) {
+    return null;
+  }
+
+  if (resolved.isFallback) {
+    return <SvgIcon src={resolved.src} size={size} className={className} />;
+  }
+
+  return (
```
Comp0te, line 72: This `` takes any `https://` string from an API response with no `referrerPolicy`, under `img-src https: data:` (`webpack.config.js:98`, `utils.ts:176`).

The policy itself isn't something this PR loosened — it's byte-identical on the base. What changes is reach: a `.svg`-suffixed url on an arbitrary host used to go to react-inlinesvg and get dropped by connect-src, and now it loads. In `token-row.tsx` the delta is wider still, since the old branch sent everything except `.jpg` into `SvgIcon`.

So whoever controls a contract's `icon_url` or `branding.logo` now gets a request from the user's browser — IP, timing, and per-row correlation to which contracts and tokens that wallet holds — each time the row renders, including on the deploy-signing screens. I'd stop short of claiming what headers ride along; I haven't checked what an extension-page subresource actually sends, and none of the `declarative_net_request_rules.json` rules bears on it either way (all twelve *set* a Referer, each pinned to one allowlisted host).

Routing these through `image-proxy-cdn.make.services` — already in host_permissions and connect-src — and tightening `img-src` to that host plus `data:` would close it; `referrerPolicy="no-referrer"` here is the cheap partial version.

src/libs/ui/components/remote-icon/remote-icon.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import { resolveIconSrc } from './resolve-icon-src';
+
+export interface RemoteIconProps {
+  src?: string | null;
+  size?: number;
+  alt?: string | null;
+  title?: string | null;
+  /** Bundled `assets/icons/*.svg` path only — it is inlined by SvgIcon. */
+  fallbackSrc?: string;
```
Comp0te: `fallbackSrc` is documented one line above as a bundled `assets/icons/*.svg` path only, but typed `string`, and it goes straight into `SvgIcon` — `resolveIconSrc` at `:51` returns it as `resolved.src` when `isFallback` (`resolve-icon-src.ts:24-25`), and `:57-58` renders it. `<RemoteIcon src={apiUrl} fallbackSrc={otherApiUrl} />` type-checks.

That matters because the fallback is the network-failure path, so a remote one fails in exactly the state it exists to cover: the primary `` 404s, the fallback gets inlined instead, and the user gets nothing rather than the generic icon. It's also the same sink as the routing predicate in `account-info-icon.tsx`.

`DeployIcon` (`constants.ts:176-186`) is already a string enum of bundled paths, and every value reaching `fallbackSrc` today is a member of it — 30 of the 32 `defaultSvg={…}` sites pass a `DeployIcon.*` literal directly, the other two forward one, and `deploy-container.tsx:29` passes `DeployIcon.Generic`. So narrowing the type wouldn't break a call site. It isn't quite a two-line change, though: `defaultSvg?: string` would have to narrow with it at `account-info-icon.tsx:21`, `common.tsx:132` and `common.tsx:283`.

There is a partial runtime guard — `SvgIcon` calls `assertLocalIconSrc` at `svg-icon.tsx:102` — but it returns early in production builds (`assert-local-icon-src.ts:14-15`), so it won't catch this in a shipped build.

src/libs/ui/components/account-info-icon/account-info-icon.tsx
ost-ptk: Confirmed, and the open question resolves worse than the comment supposed — so this is fixed by inverting the predicate rather than by patching the case-sensitivity.

On `data:` and raw `<svg>`: both are reachable, and neither involves a fetch. `react-inlinesvg` (`dist/index.cjs:242-245`):

```js
const dataURI = /^data:image\/svg[^,]*?(;base64)?,(.*)/.exec(src);
if (dataURI) inlineSrc = dataURI[1] ? window.atob(dataURI[2]) : decodeURIComponent(dataURI[2]);
else if (src.includes("<svg")) inlineSrc = src;
```

So it inlines a `data:image/svg+xml` payload *and any string containing `<svg`* directly into the DOM. `connect-src` is not a gate for either. Under `endsWith('.svg')` both shapes went to the inert ``; under `hasHttpPrefix` they went to the inliner.

Fixed as you suggested — the `SvgIcon` branch now requires a bundled-asset shape rather than defaulting every non-http value into it. New predicate in `src/utils.ts`:

```ts
const bundledAssetPathRegex = /^\/?assets\//;
export const isBundledAssetPath = (src: string) => bundledAssetPathRegex.test(src);
```

All eight call sites now read `isBundledAssetPath(x) ? <SvgIcon/> : <RemoteIcon/>`, so `HTTPS://`, `data:`, raw `<svg>` and protocol-relative urls all land in the ``. `assertLocalIconSrc` was widened to the same predicate, which closes the blind spot you identified — it now fires for exactly the cases it used to miss. Verified every icon path in the repo is `assets/…` or `/assets/…` first, so the guard does not flood.

`hasHttpPrefix` itself is unchanged; `src/background/` still uses it for tab-url checks, which is a different question.

src/utils.test.ts
ost-ptk: Fixed, and you are right that the useful half was the template-literal pinning rather than the host list.

`src/utils.test.ts` no longer imports `./csp.json` at all. The policy string is parsed into a directive map and asserted with a single `toEqual` against hand-written literals, including the 15-host `connect-src` list. So the contents of `csp.json` are now pinned, not compared to themselves — dropping a host or adding a typo'd origin fails the test.

Trade-off, stated in a comment next to the assertion: every legitimate host change now needs a test edit. That is the intent — the change becomes visible in review rather than silent.

src/utils.test.ts
ost-ptk: Fixed. All four are gone, replaced by the same treatment as the `script-src` assertion.

The block now parses the policy into a `Record<string, string>` and does one `toEqual` against a literal map. That closes all three gaps at once: each directive is pinned by exact value (so `default-src 'none' https:` fails), a removed directive fails, and a newly appended one fails too — `toEqual` is exhaustive on the key set.

Proved both directions rather than assuming: temporarily widened `default-src` — red; appended a `worker-src` directive — red.

Your point about the blast radius is why this got the stricter treatment rather than a narrower fix.

webpack.config.js
ost-ptk: Fixed. `img-src https: data:; media-src https: data:` moved into `baseDirectives` in `src/csp.json`, so both consumers derive them. The style arm is now the only per-target difference.

Verified the built policy is semantically unchanged: built `build:chrome` before and after, parsed `content_security_policy.extension_pages` into directive→value maps, same 11 directive names with the same values — only the per-build `style-src` nonce differs.

The gap you point at behind this is still real and I did not close it: `csp.json` fields are pulled in by name at each consumer, so adding a *new* field still requires wiring it into both template literals by hand. The `img-src`/`media-src` drift is closed; the general "add a field and both pick it up" property does not exist.

src/manifest.v3.json
ost-ptk: Fixed, and in the place you suggested.

`src/webpack-config.test.ts` now asserts `casper-assets` is absent from the built manifest for all three variants, going through the real CopyWebpackPlugin transform rather than reading `src/manifest.*.json` — the built artefact is what ships, and reading the source would pass even if the build injected it. `it.each(['chrome','firefox','safari'])` gives each variant its own `loadConfig` call against its own source file.

Proved live rather than assumed: re-added the entry to `manifest.v3.json` — the chrome case went red, the other two stayed green; same for `manifest.v2.json` and the firefox case.

The Chrome re-approval consequence you mention is why this got a real assertion rather than a grep in CI.

src/libs/ui/components/remote-icon/remote-icon.tsx
ost-ptk: Fixed, though the first attempt made it worse and that is worth recording.

I hoisted the transition into a pure `nextIconState(prev, {src})` returning a state object. Review of that caught two problems: the assertions were all `.toEqual`, so a change from `return prev` to a structurally-equal fresh object would have passed while breaking the `Object.is` bailout; and worse, the changed-`src` branch allocated a new object unconditionally — including the common case where `hasError` was already `false` — so `setState` no longer bailed out and every ordinary src change forced a wasted second render on exactly the recycled Activity rows this was meant to protect. Output was byte-identical, so no test would ever have shown it.

The root cause was that carrying `src` in state duplicated what the `useEffect` dependency array already does. Final shape drops it:

```ts
export type IconErrorEvent = { type: 'srcChanged' } | { type: 'loadError' };
export const nextHasError = (hasError: boolean, event: IconErrorEvent): boolean => { … }
```

State is a plain boolean again, nothing allocates, and the `Object.is` bailout behaves exactly as the pre-refactor `setHasError(false)` / `setHasError(true)` did.

One honest limit, stated in the module comment: "an unchanged `src` preserves `hasError`" is *not* one of this function's branches — there is no `srcUnchanged` event. That guarantee comes from the dependency array not invoking it at all, which is React's contract, not this module's. The two rules that are real per-event transitions are asserted.

src/libs/ui/components/deploy-plate/components/deploy-container.tsx
ost-ptk: Fixed for all three, and your read on the e2e coverage was correct — `grep -rln "Activity\|deploy-details\|DeployDetails" e2e-tests/popup/` returns nothing, so two of the three had neither unit nor e2e cover.

Each call site now has a `renderToStaticMarkup` test asserting both directions: an `https://` src yields `<img` and never `<svg`; an `assets/icons/…` path takes the `SvgIcon` branch.

One deviation worth flagging. I could not test the branches in place: all three host files reach `webextension-polyfill` at module load through `@libs/layout` → `header` → `close-current-window`, so importing any of them under `testEnvironment: 'node'` throws "This script should only be loaded in a browser extension." The icon branch in each is therefore extracted into a sibling component file, which the original site consumes. That is three new small files rather than three inline ternaries — a real cost, taken because the alternative was no coverage at all.

Each test was verified to fail: flipped each branch to send everything into `SvgIcon`, watched it go red, reverted.

src/apps/popup/pages/transfer/components/token-row.tsx
ost-ptk: Fixed — `RemoteIcon` takes a `borderRadius` prop and `token-row.tsx` passes `100`, matching the `theme.borderRadius.hundred` the removed `LogoImg` used. Confirmed your scoping too: of the six `LogoImg` definitions this migration removed, that was the only one carrying a radius.

One consequence to flag, because it goes beyond restoring the regression and no test will catch it. The old branch was `isTokenIconJPG ? LogoImg : SvgIcon`, so only `.jpg` icons were round. After the routing inversion every non-bundled icon in that list goes through `RemoteIcon`, so `.png` and `.svg` token logos are now round too where they used to be square. More consistent, but it is a visible change — happy to gate it on a `borderRadius` only for the shapes that had it before if design would rather not.

src/libs/ui/components/remote-icon/remote-icon.tsx
ost-ptk: Added `referrerPolicy="no-referrer"`, with a comment recording that it is the partial version and why the full one is blocked: `image-proxy-cdn.make.services` returns `403 requested URL is not allowed` for casper-assets urls, so routing through it needs a backend decision before `img-src` can be tightened.

Your framing of the delta is right and worth keeping in the record: the policy is byte-identical to the base, what changed is reach — a `.svg` on an arbitrary host used to be dropped by `connect-src` and now loads.

Two things surfaced while doing this. First, `referrerPolicy` only suppresses the `Referer` header; IP, timing, `User-Agent`, `Sec-Fetch-*` and any cookies the icon host already holds are untouched, so the correlation you describe is reduced, not removed. Second — relevant to whoever picks up the proxy work — `declarative_net_request_rules.json` rule 1 *sets* `Referer: https://casperwallet.io` for `image-proxy-cdn.make.services` specifically. So once icons are routed through the proxy, that rule will force a domain-level Referer regardless of this attribute. The "full fix" will not be referrerless without also revisiting that rule.

src/libs/ui/components/remote-icon/remote-icon.tsx
ost-ptk: Fixed — `fallbackSrc` is now `DeployIcon`, and the same narrowing carried to `ResolveIconSrcProps.fallbackSrc` and to `defaultSvg` at `account-info-icon.tsx:21`, `common.tsx:132` and `common.tsx:283`. As you predicted, no production call site needed changing: all 32 already passed `DeployIcon` members.

Your point about it being the same sink as the routing predicate is what made this worth doing properly rather than just documenting — and the runtime guard's production early-return is exactly why the type had to carry it.

One follow-on: with the type narrowed, `resolveIconSrc`'s `fallbackSrc ? … : null` could no longer be reached with a defined-but-falsy value, so keeping it meant a test that cast `'' as DeployIcon` to reach its own subject. That is simplified to `== null` and the cast is gone.

src/libs/ui/components/account-info-icon/account-info-icon.tsx
Comp0te: The inversion landed and closes the half this was mostly about — `HTTPS://`, `data:`, protocol-relative and bare `<svg>` all reach the `` now, and widening `assertLocalIconSrc` to the same predicate closes the blind spot where the guard was silent by construction.

One shape still gets through, and it's the same defect rather than a new one. `isBundledAssetPath` is `/^\/?assets\//` (`src/utils.ts:22`) — a prefix test with no constraint on the remainder — so an API-supplied `assets/<svg onload=…></svg>` returns true:

```js
/^\/?assets\//.test('assets/<svg onload=1></svg>')  // true
```

That routes to `SvgIcon`, and it's the `src.includes("<svg")` branch of the snippet you quoted that takes it from there — inlined into the DOM, no fetch, `connect-src` not consulted. Reachability is unchanged from the original report: `unknown-contract-action-rows.tsx:350` and `eip712-display-row.tsx:42`. `assertLocalIconSrc` is silent on this shape by construction — it fires on values that *fail* `isBundledAssetPath` — so the `[SvgIcon]` marker can't catch it either.

Worth noting the docstring you added at `src/utils.ts:24-32` states the intent exactly: "keeps every unrecognised shape out of the inliner by default". Anchoring the remainder is enough to make the code match it — requiring a path segment after `assets/`, or rejecting any src containing `<`.

webpack.config.js
Comp0te: The move landed and the before/after comparison you ran is the right check — `img-src`/`media-src` are in `baseDirectives` and both consumers now carry only a pointer comment.

On the residual you flagged yourself, there's a sharper version of it than "a new field needs wiring by hand". The Chrome arm appends its hosts to the end of the assembled string (`webpack.config.js:107`/`:110`), and that string ends with the `connect-src` arm (`:98`) — so the append is positional, not addressed to `connect-src`. Add `; frame-src 'self'` to the end of the `:98` template and the two bringweb3 hosts extend `frame-src` instead: `connect-src` loses them, `default-src 'none'` blocks every bringweb3 call in Chrome builds, and both test files stay green — `utils.test.ts` pins only the Safari string, and `webpack-config.test.ts` asserts nothing about the Chrome/Firefox directive set beyond the nonce (`:150-184`) and the casper-assets negative (`:219-226`).

Making the Chrome hosts part of the `connectSrc` array rather than a trailing append would remove the positional coupling; asserting the built Chrome/Firefox directive map the way `utils.test.ts` now asserts Safari's would catch it either way.

src/libs/ui/components/remote-icon/remote-icon.tsx
Comp0te: The record of the discarded first attempt is genuinely useful, and the final shape matches it — plain boolean, nothing allocates, and the module comment is honest about "unchanged `src` preserves `hasError`" being React's guarantee rather than the function's.

The two rules are covered now. What the original comment was about — the component's own behaviour — still isn't, and the extraction sharpened one edge of it. The two dispatch sites are interchangeable at the type level: both are `setHasError(prev => nextHasError(prev, …))` taking an `IconErrorEvent`, so writing `{ type: 'srcChanged' }` at `remote-icon.tsx:80` type-checks and leaves the suite green, turning `onError` into a no-op; the symmetric swap at `:59` makes every `RemoteIcon` enter the error state on mount, and since `ContractIcon` (`contract-icon.tsx:26`) and `NftTokenDetailIcon` (`nft-token-detail-icon.tsx:23`) pass no `fallbackSrc`, `resolveIconSrc` returns `null` and they render nothing at all.

Neither swap is observable: `jest.config.js:3` is still `testEnvironment: 'node'`, there is no `jsdom` / `jest-environment-jsdom` / `@testing-library/react` in `package.json`, and `remote-icon.test.tsx` is `renderToStaticMarkup` throughout, including the two new cases. Deleting `onError` at `:79-81` or emptying the dep array at `:60` fails nothing.

Not asking for jsdom in this PR — but the wiring is where the interesting failure now lives, and it is unpinned in both directions.

src/apps/popup/pages/transfer/components/token-row.tsx
Comp0te: The radius is back on the path that lost it — `borderRadius={100}` at `token-row.tsx:43`, matching `theme.borderRadius.hundred`, asserted at `remote-icon.test.tsx:46-63`. Confirming your scoping independently: of the six `LogoImg` definitions this migration removed, `token-row.tsx:19-23` was the only one carrying a radius, and none remain.

On the open question — it's design's call rather than mine, but one technical note bears on it. The old behaviour was gated on `tokenIconFormat === 'jpg'` (base `token-row.tsx:38,43`), i.e. on the file extension. Restoring exactly that shape means reintroducing an extension sniff at the one call site this PR removed one from, and it would sit directly beside `isBundledAssetPath`, which deliberately decides on origin rather than on suffix. If design does want the old split back, better expressed as something the row already knows — the token's own kind — than as a re-read of the url's extension.

Leaving this thread open rather than resolving it, since the question is still yours to get an answer to.

src/libs/ui/components/remote-icon/remote-icon.tsx
Comp0te: `referrerPolicy="no-referrer"` is in and asserted (`remote-icon.test.tsx:38-44`), and both of your caveats check out. Rule 1 of `declarative_net_request_rules.json:1-17` does set `Referer: https://casperwallet.io` for `image-proxy-cdn.make.services`, and it is `urlFilter`-scoped to that host, so it doesn't undo the attribute for today's icon hosts — the warning is for whoever picks up the proxy work, exactly as you put it.

Keeping the open half explicit so it survives this thread: the corrected version of the original point was never about the header, it was about the request. On every converted surface — signing screens included — the browser now makes a direct connection to a contract-controlled host, and `no-referrer` touches none of IP, timing or per-row correlation. The comment at `remote-icon.tsx:83-89` records that accurately.

The one thing I can't check from here is the blocker itself — that `image-proxy-cdn` 403s casper-assets urls is a backend fact. Fine to defer, but it is what the full fix waits on rather than a detail of it.

src/libs/ui/components/account-info-icon/account-info-icon.tsx
```diff
@@ -57,14 +55,15 @@ export const AccountInfoIcon = ({
   }
 
   if (iconUrl) {
-    return iconUrl.endsWith('.svg') ? (
-      <SvgIcon src={iconUrl || ''} alt={accountName || ''} size={size} />
+    return isBundledAssetPath(iconUrl) ? (
+      <SvgIcon src={iconUrl} alt={accountName || ''} size={size} />
```
Comp0te, line 59: Separate from the routing thread on the line above — this one is about what regression-tests the new predicate.

The push converted eight routing sites and gave three of them tests: `contract-icon.tsx:23`, `nft-token-detail-icon.tsx:20` and `deploy-container-icon.tsx:27` were extracted into siblings with two-direction `renderToStaticMarkup` tests. The other five kept the ternary inline and have no test at any level — this one, `avatar.tsx:126`, `token-plate.tsx:71`, `token-row.tsx:35`, `token-swticher-row.tsx:29`. `src/utils.test.ts:57-92` asserts `isBundledAssetPath` itself, never its use.

Two of the five are the surfaces the routing change exists for. `AccountInfoIcon` renders on the signing screen at `src/apps/signature-request/pages/sign-transaction/unknown-contract-action-rows.tsx:350` with `iconUrl={action.iconUrl}` straight off the deploy-parsing response; `avatar.tsx`'s `Logo` gets `accountsInfo[…].brandingLogo` on the signing header via `src/libs/layout/header/header-signing-account.tsx:36,41-49` (rendered from `header-popup.tsx:59`). Swapping the two arms back here type-checks — both consume the same `iconUrl: string` inside the `if (iconUrl)` guard — and leaves the whole unit suite green.

The e2e `[SvgIcon]` marker isn't a backstop for it either. It fires only when a driven screen renders a non-bundled value through `SvgIcon` (`assert-local-icon-src.ts:18`), and no popup spec drives one: the specs that do target icons address bundled `SvgIcon`s by test id (`import-account-with-file.spec.ts:60-62`, `remove-imported-account.spec.ts:59`, `disconnect-account.spec.ts:40-43`), and the only two `getByRole('img')` uses are `settings/change-password.spec.ts:37,82`, both scoped to the bundled network-switcher icon.

Worth the same extraction-plus-`renderToStaticMarkup` treatment you already applied three times, at least for these two.

src/utils.test.ts
```diff
@@ -0,0 +1,168 @@
+import {
+  getSafariCspContent,
+  hasHttpPrefix,
+  isBundledAssetPath
+} from './utils';
+
+// Splits a `directive value; directive value` policy string into a
+// { directiveName: value } map. Tolerant of a trailing semicolon, repeated
+// whitespace inside a directive, and a directive with no value.
+const parseCspDirectives = (policy: string): Record<string, string> =>
```
Comp0te, line 10: `parseCspDirectives` builds its map with `Object.fromEntries`, so a repeated directive name collapses to the last occurrence and the map no longer uniquely determines the policy string it came from.

Concretely: prepend `default-src *; ` to `baseDirectives` in `src/csp.json:2`. `getSafariCspContent()` (`src/utils.ts:189-190`) then yields `default-src *; default-src 'none'; object-src …`; the helper emits `['default-src','*']` and then `['default-src',"'none'"]`, the second overwrites, and the `toEqual` at `:135-162` passes unchanged. The other assertions in the block don't catch it: `:108`'s `not.toContain("'unsafe-eval'")` doesn't match inside `'wasm-unsafe-eval'`, `:109`'s `[^;]*` can't cross a directive boundary, and `:166` is the casper-assets negative. Nothing outside this file pins `default-src` either — `webpack-config.test.ts` asserts only nonce integrity (`:150-184`), `style-src 'unsafe-inline'` (`:158`) and the casper-assets absence (`:224`).

`script-src` is the one directive protected against this, and by a different assertion than it looks: `:100-104` uses `.find(d => d.startsWith('script-src '))`, which returns the *first* occurrence and so compares the prepended one against the literal. No other directive has that form — `style-src` and `connect-src` included.

I'd rather not lean on which occurrence a browser honours for a duplicated directive name; that's a spec question and the answer doesn't change the conclusion. This assertion exists to make a silently-widened policy fail, and a widening expressed as a duplicate is normalised away before the comparison happens. One line fixes it: collect the names into an array first and assert no duplicates, or compare the split array rather than a map.

Comp0te (commented):
Second pass over the icon-routing change, looking only at what moved since the first round. Five of the earlier points are closed and their threads are resolved; five have a named residual or an open question, and I've replied in place rather than opening anything new. The two comments below are new ground from this push. Both are about the same gap between what the change asserts and what it protects: the routing predicate is well covered as a function and untested as a decision at five of its eight call sites, and the CSP assertion that replaced the substring checks normalises away one of the widenings it was added to catch.

PR #1433, opened 2026-07-30
fix: restore the Home tab the user actually left (WALLET-1360)

ost-ptk:
## Description

Going back from a token, an NFT or a deploy could land the user on the wrong Home tab — the ticket reports Activity, but NFTs happens too, depending only on what the user did earlier.

### Root cause

The active Home tab lived in the `Tabs` component's local `useState`, initialised once from `location.state.activeTabId`. React Router binds `state` to a history entry, and nothing wrote the user's tab click back into it. So:

1. A completed send / stake, or Back from NFT or deploy details, pushed a Home entry carrying `activeTabId`.
2. The user then clicked a different tab — that changed only the component's `useState`; the history entry kept the old value.
3. Opening a token pushed a new entry; Back (`navigate(-1)`) returned to that same Home entry, and `Tabs` re-initialised from the **stale** `activeTabId`.

The underlying design problem is that one field served two different jobs: _restore_ the tab the user was on, and _override_ it to Activity after a submission. Because both wrote `location.state.activeTabId`, an override stuck to the history entry permanently and later impersonated a restore.

### Fix

- New `HomeTabProvider` / `useHomeTab` (`src/apps/popup/hooks/use-home-tab.tsx`) owns the active tab for the popup session. It is mounted inside `<HashRouter>` **above** `<AppRoutes>`, which matters: `AppRoutes` branches on `showNavigationMenu` and renders a separate `<Routes>` tree that unmounts Home.
- `Tabs` is now controlled-optional and addresses tabs by **name** instead of position in the children array. `HomePageTabsId` (numeric indices that were never tied to the render order) is gone; `HomePageTabName` is the single source of truth. `RecipientTabs` keeps working uncontrolled and is unchanged.
- The four sites that mean "send the user to Activity" now say so directly via `setActiveHomeTab`. The two that meant "go back" are plain `navigate(-1)`, since the session owner already knows the tab — which made the `navigateHome` prop chain through `DeployPlate` dead, so it is deleted along with `LocationState.activeTabId`.
- The tab stays ephemeral: no Redux slice, no `storage.local`. Closing the popup (or auto-lock) resets it to Tokens, as before.

### Rider: `useNavigationMenu` discarded the rest of `location.state`

Separate pre-existing bug, fixed here because it lives in the same mechanism. The three helpers replaced the whole state object with `{ showNavigationMenu }`. Consequences today, independent of the tab bug:

- **Deploy details renders blank** after opening and closing the burger menu — it reads `location.state.deploy` and its route `/deploys-details` carries no parameter, so there is no fallback and `useFetchSingleDeploy` never fires.
- **NFT transfer loses its preview** mid-flow — `transfer-nft` reads `location.state.nftData`.
- Token details only degrades, recovering via the `:tokenName` param.

The fix spreads the existing state. Purely additive.

### Second rider: the transfer success screen named a tab that does not exist

The success copy read "You can check its status in the **Transactions** / **Deploys** tab on your Wallet home page", branching on `isCasper2Network`. Neither tab exists — the Home tabs are Tokens, NFTs and **Activity**, and have been for some time.

It was also a template literal inside `<Trans>`, so `i18next-parser` never extracted it and no catalog carries a translation for it. Making the sentence static fixes the extraction too, and dropping the network branch leaves `selectIsCasper2Network` unused in the component.

## Notes for reviewers and QA

- **Wider behaviour change than the ticket describes.** Because the tab is session-scoped rather than reset by every stateless `navigate(Home)`, _every_ "go Home" path (header Close/Cancel/Done, contacts, account settings, rename/remove account, RateApp completion, account and network switches) now returns to the last-used tab instead of always Tokens. That is what the design asks for, but it is the thing most likely to surprise manual testing.
- **Please manually verify "Done → Activity".** It has no automated coverage and is unreachable in the e2e environment: on fresh mock state `askForReviewAfter` is `null`, so Done always routes to the RateApp screen. In production any returning user who has already seen the rate prompt takes the Home branch. Repro: dismiss the rate prompt once, then send CSPR → Done should land on Activity.
- **Please retest NFT transfer with the burger menu**, per the rider above.
- **Tab widths: computed to be unchanged, but worth a glance.** `ActiveTabContainer` / `TabContainer` now use `flex: 1` instead of `width: calc(33% - 8px)`. These already inherited `flex-grow: 1` from `TabsContainer`'s `flexGrow={1}`, so with three tabs the old rule resolved to a base of `0.33·w − 8` plus an equal share of the `0.01·w + 24` free space — exactly `w/3`, which is what `flex: 1` gives directly. The change also removes the component's undocumented three-tabs-only constraint. Note this affects the **recipient tab strip** (Recent / My accounts / Contacts) in the transfer flow as well as Home, so it is worth a quick look at both.
- Follow-up worth a ticket: `TransferSuccessScreen` lives in `src/libs/ui/components/` and is re-exported from the barrel all five apps import, but it now requires a `HomeTabProvider` ancestor. All current call sites are popup pages, so it is safe today; moving it under `src/apps/popup/pages/` would make the coupling honest and drop it from the other four bundles.

## Testing

- `npm run ci-check` — passes (405 jest tests, 0 lint errors, tsc clean, coverage gates met).
- New `e2e-tests/popup/common/home-tabs.spec.ts` — four scenarios, all of which failed before the fix and pass after it: the NFTs variant, the Activity variant, the tab surviving a menu toggle, and deploy details surviving a menu toggle.
- Full `e2e-tests/popup` suite — no regressions. One unrelated flake (`signature-request-scenarios › should cancel the signing process`) passed on retry.

## Linked tickets

[WALLET-1360](https://make-software.atlassian.net/browse/WALLET-1360)

## Checklist

- [x] If the PR adds any new text to the UI, make sure they are localized — see the note below

**On localization:** this PR deliberately does **not** regenerate `lang/`. Running `npm run locale:extract_pot` produces **410 new msgids, of which only one is from this PR** — the other 409 are accumulated drift from strings added to the codebase since the last extraction, which is untracked in every one of the 10 locales. Committing that here would bury the actual change. The one new string renders in English (i18next falls back to the key), exactly as the sentence it replaces did — it was never extractable either. **The catalog drift deserves its own ticket.**

- [ ] Include a screenshot or recording if implementing significant UI or user flow change

[WALLET-1360]: https://make-software.atlassian.net/browse/WALLET-1360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/apps/popup/pages/transfer/index.tsx
```diff
@@ -515,15 +517,8 @@ export const TransferPage = () => {
...
-                }
-              };
-
-              // Navigate to "Home" with the pre-defined state
-              navigate(RouterPath.Home, homeRoutesState);
+              setActiveHomeTab(HomePageTabName.Activity);
+              navigate(RouterPath.Home);
             } else {
               // Navigate to "RateApp" when the application has not been rated in the store, and it's time to ask for a review.
               navigate(RouterPath.RateApp);
```
Comp0te, line 529: This line is unchanged by the PR, but the change alters where it puts the user.

`RouterPath.RateApp` is reached from exactly three places — here, `stakes/index.tsx:534` and `transfer-nft/index.tsx:435` — all of them the `else` branch of a Success-step `Done` handler. So every exit from RateApp is a post-submission exit, and RateApp's four exits to Home (`rate-app/index.tsx:98`, `:115`, `:129`, `:153`) don't call `setActiveHomeTab`.

Before this PR that didn't matter: `Tabs` held its own `useState(preferActiveTabId || 0)` inside `HomePageContent`, which is unmounted while RateApp renders, so a bare `navigate(RouterPath.Home)` always came back on Tokens. Now `home/index.tsx:152` passes `activeTabName={activeHomeTab}` from the session-scoped context, which survives the RateApp round trip — so the user lands on whatever tab they were on before starting the transfer.

Concretely: a user who hasn't yet dismissed the rate prompt sends from the NFTs tab, is shown "You can check its status in the Activity tab on your Wallet home page" (`transfer-succeess-screen.tsx:64-68`), taps through RateApp, and lands back on NFTs. `askForReviewAfter == null` makes this the branch that runs on fresh state, so it's the common path for a first submission.

Either call `setActiveHomeTab(HomePageTabName.Activity)` on RateApp's four exits too, or hoist the override so it's applied once at submission time rather than per exit branch.

Comp0te (commented):
Read this through the WALLET-1360 change from index-addressed Home tabs to the named context. The change itself holds up — the context is correctly scoped below the lock gate, and the four post-submission sites now say what they mean instead of smuggling an index through router state. The comments are about the edges of it: one exit path that wasn't brought along with the other four, and three places where the new behaviour, or the invariant it depends on, has nothing standing behind it.

src/apps/popup/pages/transfer/index.tsx
```diff
@@ -515,15 +517,8 @@ export const TransferPage = () => {
...
-              const homeRoutesState = {
-                state: {
-                  // set the active tab to deploys
-                  activeTabId: HomePageTabsId.Deploys
-                }
-              };
-
-              // Navigate to "Home" with the pre-defined state
-              navigate(RouterPath.Home, homeRoutesState);
+              setActiveHomeTab(HomePageTabName.Activity);
```
Comp0te: This pairing — `setActiveHomeTab(HomePageTabName.Activity)` then `navigate(RouterPath.Home)` — is new here and at `transfer-nft/index.tsx:431`, `stakes/index.tsx:529` and `transfer-succeess-screen.tsx:33`, replacing the old `navigate(RouterPath.Home, { state: { activeTabId: HomePageTabsId.Deploys } })`. Nothing asserts it.

The three transfer specs end at their `Done` click and assert nothing after it (`casper-native-transfer.spec.ts:85`, `erc-20-transfer.spec.ts:94`, `nft-transfer.spec.ts:77`); `delegate.spec.ts:64` clicks `Done` then `Close` with nothing in between. The new `home-tabs.spec.ts` covers back-navigation and the navigation menu — none of its four scenarios performs a submission. There's no jest test for any of the four files. Change `HomePageTabName.Activity` to `Tokens` on this line and `ci-check` plus both e2e projects stay green.

One line at the end of a transfer spec closes it: `await popupExpect(popupPage.getByTitle('Activity')).toBeVisible()`. `getByTitle` is a real observation of activeness — only `ActiveTabContainer` carries `title={tabName}` (`tabs.tsx:100`), `TabContainer` doesn't.

src/libs/ui/components/tabs/tabs.tsx
```diff
@@ -51,37 +45,66 @@ interface TabProps {
...
+  const currentTabName = children.some(
+    tab => tab.props.tabName === requestedTabName
+  )
+    ? requestedTabName
+    : firstTabName;
 
   const { t } = useTranslation();
 
+  const handleTabClick = (tabName: string) => {
+    if (activeTabName === undefined) {
```
Comp0te, line 79: This guard is new — before, the tab's `onClick` called `setActiveTabId(index)` unconditionally — and no check ever executes it.

The only uncontrolled `<Tabs>` in the tree is `recipient-tabs.tsx:163`; `home/index.tsx:152` is the controlled one. The transfer specs do mount that strip — they reach the recipient step with the field empty and assert on it (`erc-20-transfer.spec.ts:31-37`, `nft-transfer.spec.ts:37-43`) before filling in a public key, which then swaps in `SearchItemByPublicKey`. What no spec ever does is *click* one of its tabs, so the `activeTabName === undefined` branch never runs. "My accounts" and "Recent" have no hits anywhere in `e2e-tests`, and the "Contacts" hits are all the navigation-menu item rather than the tab. There's no test file in `src/libs/ui/components/tabs/`.

So inverting or dropping this guard freezes the recipient strip on "Recent" — someone sending or delegating couldn't pick from "My accounts" or "Contacts" and would have to paste a raw public key — with every automated check still passing.

There's no jsdom here (`jest.config.js` sets `testEnvironment: 'node'`), so this can't be unit-tested as-is. An e2e step that stops at "Specify recipient" without typing a key, clicks "My accounts", and asserts an account row is listed would cover it.

src/apps/popup/app-router.tsx
```diff
@@ -63,7 +64,9 @@ export function AppRouter() {
 
   return (
     <HashRouter>
-      <AppRoutes />
+      <HomeTabProvider>
```
Comp0te, line 67: The Home tab's ephemerality — the property `use-home-tab.tsx:18-26` is documented around ("destroyed when the popup closes (and when the vault locks), so reopening always starts on Tokens") — rests entirely on this provider sitting below the `if (isLocked) return <LockedRouter popupLayout />;` at `:61-63`, so the whole `HashRouter` subtree unmounts on lock. The state is a plain `useState` (`use-home-tab.tsx:28-30`) with no persistence, and nothing else resets it, so placement is the only thing enforcing it. Nothing asserts that.

Hoisting the provider to wrap the whole `AppRouter` return — an ordinary-looking refactor — compiles and passes lint, tsc, knip, the jest suite and both e2e projects. `wallet.spec.ts:5-26` locks and unlocks but asserts only the lock-screen text and the account name; `change-password.spec.ts:33,78` locks and asserts only the network-switcher icon; `home-tabs.spec.ts` never locks. Those are the only `lockVault` call sites in the suite.

A fifth scenario in `home-tabs.spec.ts` using the existing fixtures covers it: click NFTs, `lockVault()`, `unlockVault()`, expect `getByTitle('Tokens')`.

src/apps/popup/pages/transfer/index.tsx
ost-ptk: Confirmed — `rate-app/index.tsx` has four `navigate(RouterPath.Home)` calls and none of them touch the tab, and `RouterPath.RateApp` is reached only from the three Success-step `Done` handlers.

Took the second option: `setActiveHomeTab(HomePageTabName.Activity)` now runs before the branch in all three (`transfer/index.tsx`, `stakes/index.tsx`, `transfer-nft/index.tsx`), so RateApp stays unaware of Home tabs rather than repeating the override on four exits.

Fixed in c4da10b0.

src/apps/popup/pages/transfer/index.tsx
ost-ptk: Verified the gap — flipping `Activity` to `Tokens` did keep everything green.

One correction to the suggested placement, though: on fresh mock state `askForReviewAfter` is `null`, so `Done` lands on RateApp, not Home. An assertion immediately after the `Done` click sees the rate prompt. The reachable observation point is one step later, on RateApp's exit — which, now that the override is hoisted (see the thread above), exercises the same code.

So the assertion went in at four places:

- `casper-native-transfer.spec.ts` and `nft-transfer.spec.ts` — `Done` → `Close` → `getByTitle('Activity')`
- `delegate.spec.ts` — already clicked `Close` after `Done`; just added the assertion
- `review-flow.spec.ts` "should redirect to home page" — covers the Support → "Maybe later" exit

`getByTitle` as you suggested. Un-hoisting the override fails all four.

I did try to make the direct `navigate(RouterPath.Home)` branch reachable within one spec — first submission → dismiss the prompt (pushes `askForReviewAfter` a month out) → second submission takes the other branch. The second `sendCsprTokens` hung on the recipient step, so I dropped it as too fragile for the coverage it adds on top of the above.

src/libs/ui/components/tabs/tabs.tsx
ost-ptk: Reproduced: inverting the guard to `activeTabName !== undefined` leaves `ci-check` and both e2e projects green while the recipient strip freezes on "Recent".

Added `should switch the recipient tabs` to `casper-native-transfer.spec.ts`, along the lines you described — stop at "Specify recipient" with the field empty, assert `getByTitle('Recent')`, click "My accounts", assert the tab went active and that `Account 2` is listed, then click back to "Recent". The inverted guard fails it.

Agreed on jsdom — `testEnvironment: 'node'` rules out a unit test here, so e2e is the only place this can live today.

src/apps/popup/app-router.tsx
ost-ptk: Added the fifth scenario — `should reset to the Tokens tab after the vault is locked` in `home-tabs.spec.ts`: click NFTs, `lockVault()`, `unlockVault()`, expect `getByTitle('Tokens')`. Hoisting the provider above the `isLocked` gate fails it.

One wrinkle worth recording: `lockVault()` locks through the burger menu, and `showNavigationMenu` outlives the lock in `location.state` — so unlocking lands back in the open menu rather than on Home, and the spec has to close it before asserting. That is the same `useNavigationMenu`/`location.state` mechanism as the rider in the description, and looks like a separate pre-existing bug; I left it alone here and it seems worth its own ticket.

The scenario also replaces the blind 2s sleep the other four use with a `getByTitle('Tokens')` wait, so a slow unlock fails with a readable message instead of timing out on the tab click. Ran it `--repeat-each=3 --retries=0`: 15/15.

src/apps/popup/pages/transfer/index.tsx
Comp0te: Confirmed the four assertions discriminate: `tabs.tsx:100` is unchanged, `title={tabName}` is still carried only by `ActiveTabContainer` and nothing else under `pages/home` carries a `title`, so `getByTitle('Activity')` observes activeness rather than presence. The RateApp leg is reachable as you describe — `src/fixtures/initial-state-for-popup-tests.ts:111-112` seeds `ratedInStore: false, askForReviewAfter: null`, and the `Close` those specs click is the `linkType="close"` link at `rate-app/index.tsx:153`. Between them the write sites at `transfer/index.tsx:523`, `transfer-nft/index.tsx:434` and `stakes/index.tsx:532` are covered.

One site from the original list is still uncovered: `transfer-succeess-screen.tsx:33`, inside the `Enter`-keydown handler at :30-41. That is the one exit from the success screen that bypasses RateApp entirely, and no spec ever presses a key — `press(`, `'Enter'` and `keyboard` have zero hits across `e2e-tests/`. Changing `HomePageTabName.Activity` to `Tokens` on that line alone still leaves `ci-check` and both e2e projects green.

A smaller second gap that the hoist itself created, worth naming: deleting or flipping the hoisted write now does fail the four new specs — but a regression that pushes it back *down* into the `if (ratedInStore || !shouldAskForReview)` leg passes everything, since the fixture only ever exercises the other leg.

Comp0te (approved):

PR #1439, opened 2026-08-07
fix: WALLET-1384 — keep the payload of every in-flight signature request

ost-ptk:
[WALLET-1384](https://make-software.atlassian.net/browse/WALLET-1384)

> **Regression already on `release/2.7.0`.** Base is `release/2.7.0`, not `develop`. If the release branch goes to QA before this merges, the bug is reproducible on the build.

## The bug

`deployPayloadReceived` / `eip712PayloadReceived` built a **new single-entry dict**, so `jsonById` and `eip712ById` never held more than one request. Every new signature request erased the payload of whatever else was in flight.

```ts
// src/background/redux/vault/reducer.ts — before
jsonById: { [payload.id]: payload.json }     // replace, not merge
eip712ById: { [payload.id]: payload.json }
```

The `replace` semantics is old (predates 2.7.0). What made it a live defect is #1427: `cancelRequestsDisplacedBy` now deliberately **spares a request another window still displays** —

```ts
const candidates = displaced.filter(r => r.windowIds.length === 1);
```

— and the one thing that attaches a second window is the Ledger permission window, which carries the same `requestId` (`use-ledger.ts` → `registerLedgerPermissionWindow`). Before #1427 (#1419 / WALLET-1353) every displaced request was cancelled, so no survivor existed.

### Reproduction (Chrome, Ledger account)

1. Dapp calls `sign` — request **A**. Approval window opens.
2. Ledger asks for device permission → a **second** window opens for A; `A.windowIds.length === 2`.
3. Dapp calls `sign` again — request **B**. `sdk-methods.ts` synchronously dispatches `deployPayloadReceived({ id: B })`, wiping A's payload.
4. `openWindow` reuses the main window → after the 250 ms grace, A still has another window → **A is not cancelled** (by design).
5. A's Ledger window re-renders: `deployJsonById['A']` is `undefined` → `useFetchDataForSignatureRequest` stays `enabled: false` → `transaction` never parses → `handleSign` returns on `if (!transaction)`.

**Result:** a live, uncancelled request on screen that cannot be signed. The button silently does nothing and the dapp promise hangs until its own 30-minute timeout.

## The fix

**Merge instead of replace**, and **delete the entry once the request is answered**.

The cleanup keys off the `windowRequestResponded` **action**, not off windowManagement's resulting state: that reducer no-ops the transition unless the request is currently `'open'`, which is precisely the orphan case (an MV3 restart between registration and the response) where a stale payload most needs dropping. When the id is in neither map the **same state object** is returned — otherwise an unrelated `windows.onRemoved` would cost a popupState broadcast to every replica plus a full `storage.local` rewrite (same reasoning as the `displaced.length > 0` gate in `cancel-requests.ts`).

The deletion is also added to the `updateVaultCipher` debounce list, where both payload **writes** already are. Without it the deletion is an in-memory edit only: an MV3 service-worker restart before the next vault write resurrects the entry through `vaultLoaded`, for a requestId whose `windowRequestResponded` has already fired. Most responses carry no payload entry, so this usually re-encrypts an unchanged vault — at ~2 ms behind a 500 ms debounce, cheaper than persisting writes but not deletes.

### The ceiling, and which write loses

A deletion can still be missed. An MV3 service-worker restart wipes `windowManagement.requests` (in-memory only) while the payload maps come back from the encrypted cipher through `vaultLoaded`, so the response for a pre-restart id can never arrive. And nothing else deletes a key: the only other clear is `deploysReseted`, a **full-slice reset**, and `lockVaultSaga` flushes `updateVaultCipher` **before** it — so the payload is already in the cipher and `vaultLoaded` restores it on the next unlock. A leaked entry therefore survives lock/unlock, lives in `storage.local` for good, and rides along in every popup-state broadcast (`selectPopupState` sends `vault` whole). Hence `MAX_STORED_PAYLOADS = 10`.

At capacity the **incoming** write is refused; nothing already stored is evicted. Making room by dropping the oldest entry would put the loss on the request that has waited longest — exactly the one this PR protects, a request being confirmed on a Ledger while a page pushes ten more of its own. `signRequest` has no connected-site precondition, the message handlers are concurrent, and a supersede only frees a slot after `windows.create` plus `CANCEL_GRACE_MS`, so a burst lands in full before any `windowRequestResponded` can. Refusing puts the loss on the request the caller controls. A rewrite of an id already stored still applies, since it cannot grow the map.

Residual and accepted, documented at the decision. A payload leaks whenever its deletion never reaches the cipher, by two routes rather than one: the request is never answered at all (auto-lock while an approval window is open, then the MV3 worker dies with no `windows.onRemoved` handler alive to cancel it), or it **is** answered and the worker dies inside the 500 ms re-encrypt debounce that would have persisted the deletion. Either way the entry outlives lock/unlock, so enough of them fill the map and refuse every later payload.

That state has an exit, written down because nothing in the UI points at it: `vaultLoaded` keeps the in-memory map whenever it is non-empty and discards the cipher's, and `signRequest` has no lock gate — so lock, let a dapp send one request, unlock, and the accumulated entries are gone. Reclaiming the slots without that sequence means reconciling the map against the windows actually open, which is a separate change.

### Prototype-chain reads

Both maps are keyed by `requestId`, which is page-generated (`generateRequestId`, `src/content/sdk.ts`) i.e. dapp-controlled, and both read sites indexed them bare. A request id of `constructor` handed the signature page an inherited `Object.prototype` member instead of transaction JSON, surfacing as `INVALID_TRANSACTION_JSON` for a request the store never held a payload for.

New `vault/payload-map.ts` → `getPayload()`, mirroring `getRequest` in `windowManagement/request-map.ts`, which exists for the same reason on the same key space.

`__proto__` is a separate case, and the write side is now guarded explicitly rather than left to the shape of the write. `tsconfig.json` targets **es2017**, so `{ ...payloads, [id]: json }` is *emitted* as nested `Object.assign` — assignment runs the `__proto__` setter instead of adding an entry:

| | string value | **object** value |
| --- | --- | --- |
| es2017 (`Object.assign`) | silent no-op, entry dropped | **sets the map's prototype** |
| es2018+ (native spread) | own property | own property |

Both of those matter here. The deploy path dispatches an **object** — `sdk-methods.ts` puts `JSON.parse(...)` into a map declared `Record<string, string>` — which is the column that replaces the prototype rather than no-opping. And resting on the emit would make the guarantee a `target` setting. So `storePayload` refuses the id through `isStorableRequestId`; the test carries both value types, and the object case is the one that fails without the guard. Still unreachable in production, since `handleSdkMethod` rejects that id at the message boundary for every approval type — but the two guards answer to different owners, and this map is the one at risk.

### Rendering and gating with no payload

`sign-transaction` guarded its loading branch on `isLoadingSignatureRequest` alone — the query's `isFetching`, which is `false` for a **disabled** query — while both content children below it are guarded on `signatureRequest`. With the payload gone the page therefore rendered an empty pane. Per-request deletion makes that state routine rather than hypothetical, so the branch now carries a `!signatureRequest` term.

`LedgerEventView` is excluded from that term. It is guarded on `signingPageState` alone, carries no `signatureRequest`, and is the only thing the Ledger permission window — the second window, the one that survives a supersede — has to show while the device is being read. Replacing it with a skeleton would blank exactly the screen this fix protects.

The Sign button was gated on `transaction`, local state parsed once and kept, so it stayed enabled over that pane. `!signatureRequest` mirrors the eip712 page. A second click was already dropped by the background response dedup, so nothing reached the dapp, but on the Ledger path it could re-prompt the device for a signature that is then discarded.

## Tests

Two existing tests pinned the buggy behaviour (`replaces the whole jsonById dict with a single-entry dict` and its eip712 twin) and were replaced. Every new test was confirmed red against the code it pins.

- an earlier payload survives a second request (both maps)
- the same id sent twice overwrites only its own entry
- at capacity: the incoming write is refused, a rewrite of a stored id still applies, and a slot freed by an answered request is reusable
- `windowRequestResponded` drops the answered id from both maps and leaves the rest; an id in neither returns the identical state object (`toBe`)
- the four inherited `Object.prototype` names, in both directions, mirroring `windowManagement/reducer.test.ts`; plus `__proto__` with both a string and an object value, asserted on what the reducer owns (empty map, prototype intact, `getPayload` agreeing) so a `target` bump cannot reverse it
- `vault-sagas.test.ts`: the deletion reaches the cipher — driven through the real vault reducer with `withReducer` and asserted on the maps handed to `encryptVault`. Verified red both ways: without the `extraReducers` block, and without the debounce-list entry.
- `payload-map.test.ts`: own keys, missing keys, all five inherited names, and an own key shadowing an inherited name

## Verification

- `npm run ci-check` — green (format, lint, tsc, knip, 542 tests). `redux/vault/reducer.ts` stays at 100% across all four metrics.
- `npm run dependency:circular` — 141 cycles before and after; the new `vault → windowManagement/actions` edge adds none (`windowManagement/{actions,reducer,types,request-map}.ts` import nothing from `vault`).
- `npm run lint` — 171 warnings before and after.
- No new action creator, so `FORWARDED_ACTION_TYPES` and `redux-actions.parity.test.ts` are untouched.

## Known gap

Neither `getPayload` call site nor the disabled direction of the Sign gate is covered: `jest.config.js` sets `testEnvironment: 'node'`, there is no React component test in the repo, and `src/apps/**` is outside `collectCoverageFrom`. Closing it means adding jsdom + Testing Library — infrastructure worth its own ticket rather than this PR.

## Out of scope (noted, separate tickets)

- `sdk-methods.ts` stores the **parsed object** in `jsonById` despite the declared `json: string` — it compiles only because `let deployJson;` is an implicit `any`. This makes `getPayload`'s `string | undefined` return type a runtime lie for the deploy map; it works because `Transaction.fromJSON` accepts objects.
- A refused payload still opens an approval window, so the dapp learns nothing until its own timeout. Rejecting the SDK request at the boundary when the map is full would give it a real error.
- `deploysReseted` is effectively a duplicate of `vaultReseted`; the name does not say so.

[WALLET-1384]: https://make-software.atlassian.net/browse/WALLET-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -403,8 +403,14 @@ export function SignTransactionPage() {
           <Button
             color="primaryRed"
             flexWidth
+            // `!signatureRequest` mirrors the eip712 page. `transaction` is
+            // local state that outlives the payload it was parsed from, so
+            // once the answered request's payload is dropped from the vault
+            // this button would stay enabled over the empty pane the content
+            // branch below renders without a `signatureRequest`.
             disabled={
               !transaction ||
+              !signatureRequest ||
               isLoadingSignatureRequest ||
               (additionalApproveRequired && !wasmApproved)
             }
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -469,7 +469,12 @@ export function SignTransactionPage() {
           );
         }
 
-        return isLoadingSignatureRequest ? (
+        // `!signatureRequest` mirrors sign-eip712. `isLoadingSignatureRequest`
+        // is the query's `isFetching`, which is false for a DISABLED query — so
+        // once the answered request's payload is gone every child below is
+        // guarded away and this branch rendered an empty pane. Per-request
+        // deletion makes that routine, not hypothetical.
+        return isLoadingSignatureRequest || !signatureRequest ? (
           <SignatureRequestLoading />
         ) : (
           <>
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -230,6 +230,23 @@ export function SignTransactionPage() {
     let signature: Uint8Array | null = null;
 
     if (!transaction) {
+      // Every signing route funnels through here, including the Ledger footer's
+      // Connect, which is outside the `disabled` gate on the main footer. If the
+      // payload is gone the request can never be answered from this window, so
+      // returning silently strands both the window and the dapp.
+      // Keyed on `transactionJson`: a payload present but not yet parsed is an
+      // ordinary in-flight state and must not error out.
+      if (!transactionJson) {
+        const error = Error(
+          ErrorMessages.signTransaction.REQUEST_NO_LONGER_AVAILABLE.description
+        );
+        sendSdkResponseToSpecificTab(
+          sdkMethod.signError(error, { requestId }),
+          requestTabId
+        );
+        closeCurrentWindow();
+      }
+
       return;
     }
 
@@ -266,6 +283,7 @@ export function SignTransactionPage() {
     closeCurrentWindow();
   }, [
     transaction,
+    transactionJson,
     signingAccount.hardware,
     signingAccount.derivationIndex,
     signingAccount.publicKey,
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -33,6 +33,7 @@ import {
 } from '@background/redux/trusted-wasm/actions';
 import { selectTrustedWasmByOrigin } from '@background/redux/trusted-wasm/selectors';
 import { dispatchToMainStore } from '@background/redux/utils';
+import { getPayload } from '@background/redux/vault/payload-map';
 import {
   selectConnectedAccountNamesByOrigin,
   selectDeploysJsonById,
@@ -138,8 +139,11 @@ export function SignTransactionPage() {
     [accounts, signingPublicKeyHex]
   );
 
+  // `getPayload`, never a bare index: `requestId` is dapp-controlled, so
+  // `deployJsonById['constructor']` would hand this page an inherited
+  // Object.prototype member instead of transaction JSON.
   const transactionJson = useMemo<string | undefined>(
-    () => deployJsonById[requestId],
+    () => getPayload(deployJsonById, requestId),
     [deployJsonById, requestId]
   );
 
--- a/src/apps/signature-request/pages/sign-transaction/index.tsx
+++ b/src/apps/signature-request/pages/sign-transaction/index.tsx
@@ -469,12 +469,20 @@ export function SignTransactionPage() {
           );
         }
 
-        // `!signatureRequest` mirrors sign-eip712. `isLoadingSignatureRequest`
-        // is the query's `isFetching`, which is false for a DISABLED query — so
-        // once the answered request's payload is gone every child below is
-        // guarded away and this branch rendered an empty pane. Per-request
-        // deletion makes that routine, not hypothetical.
-        return isLoadingSignatureRequest || !signatureRequest ? (
+        // `isLoadingSignatureRequest` is the query's `isFetching`, which is
+        // false for a DISABLED query — so once the answered request's payload
+        // is gone the two content children below are guarded away and this
+        // branch rendered an empty pane. Per-request deletion makes that
+        // routine, not hypothetical.
+        //
+        // `LedgerEventView` is the exception and is excluded from the added
+        // term: it carries no `signatureRequest`, and it is the only thing the
+        // Ledger permission window — the second window, the one that survives a
+        // supersede — has to show while the device is being read. Replacing it
+        // with a skeleton would blank exactly the screen this fix protects.
+        return isLoadingSignatureRequest ||
+          (!signatureRequest &&
+            signingPageState !== SigningPageState.LedgerConfirmation) ? (
           <SignatureRequestLoading />
         ) : (
           <>
--- a/src/background/redux/sagas/vault-sagas.test.ts
+++ b/src/background/redux/sagas/vault-sagas.test.ts
@@ -43,6 +43,7 @@ import {
   selectVaultActiveAccount
 } from '../vault/selectors';
 import { VaultState } from '../vault/types';
+import { windowRequestResponded } from '../windowManagement/actions';
 import { lockVault, startBackground, unlockVault } from './actions';
 import {
   VAULT_REENCRYPT_DEBOUNCE_MS,
@@ -444,6 +445,27 @@ describe('updateVaultCipher debounce', () => {
       .dispatch(accountRenamed({ oldName: 'a', newName: 'b' }))
       .dispatch(accountRenamed({ oldName: 'b', newName: 'c' }))
       .dispatch(accountRenamed({ oldName: 'c', newName: 'd' }))
+      .silentRun(VAULT_REENCRYPT_DEBOUNCE_MS + 200);
+
+    expect(encryptSpy).toHaveBeenCalledTimes(1);
+  });
+
+  // WALLET-1384: the vault reducer deletes an answered request's signing
+  // payload, and that deletion must reach the cipher. Otherwise an MV3
+  // service-worker restart re-loads the entry from a stale blob through
+  // `vaultLoaded`, for a requestId whose `windowRequestResponded` has already
+  // fired — so nothing but the FIFO cap would ever remove it.
+  it('re-encrypts when an answered request drops its payload', async () => {
+    const encryptSpy = jest
+      .spyOn(vaultCryptoModule, 'encryptVault')
+      .mockResolvedValue('cipher-blob');
+
+    await expectSaga(vaultSagas)
+      .provide([
+        [matchers.select.selector(selectEncryptionKeyHash), 'key-hash'],
+        [matchers.select.selector(selectVault), EMPTY_VAULT]
+      ])
+      .dispatch(windowRequestResponded({ requestId: 'answered' }))
       .silentRun(VAULT_REENCRYPT_DEBOUNCE_MS + 200);
 
     expect(encryptSpy).toHaveBeenCalledTimes(1);
--- a/src/background/redux/sagas/vault-sagas.test.ts
+++ b/src/background/redux/sagas/vault-sagas.test.ts
@@ -1,4 +1,5 @@
 import * as matchers from 'redux-saga-test-plan/matchers';
+import { combineReducers } from '@reduxjs/toolkit';
 import { expectSaga } from 'redux-saga-test-plan';
 import { storage, tabs } from 'webextension-polyfill';
 
@@ -37,6 +38,7 @@ import { selectTimeoutDurationSetting } from '../settings/selectors';
 import { vaultCipherCreated } from '../vault-cipher/actions';
 import { selectVaultCipherDoesExist } from '../vault-cipher/selectors';
 import { accountRenamed, vaultLoaded } from '../vault/actions';
+import { reducer as vaultReducer } from '../vault/reducer';
 import {
   selectAccountNamesByOriginDict,
   selectVault,
@@ -454,21 +456,43 @@ describe('updateVaultCipher debounce', () => {
   // payload, and that deletion must reach the cipher. Otherwise an MV3
   // service-worker restart re-loads the entry from a stale blob through
   // `vaultLoaded`, for a requestId whose `windowRequestResponded` has already
-  // fired — so nothing but the FIFO cap would ever remove it.
-  it('re-encrypts when an answered request drops its payload', async () => {
+  // fired — so nothing would ever remove it.
+  //
+  // Driven through the REAL vault reducer (`withReducer`) and asserted on the
+  // value handed to `encryptVault`, not on a call count. Providing `selectVault`
+  // instead would pin only that `windowRequestResponded.type` sits in the
+  // debounce array — the reducer half of the contract would go unobserved, and
+  // deleting the whole `extraReducers` block would leave the test green.
+  it('re-encrypts the vault WITHOUT the answered payload', async () => {
     const encryptSpy = jest
       .spyOn(vaultCryptoModule, 'encryptVault')
       .mockResolvedValue('cipher-blob');
 
+    const seeded: VaultState = {
+      ...EMPTY_VAULT,
+      jsonById: { answered: 'gone', other: 'kept' },
+      eip712ById: { answered: 'gone' }
+    };
+
     await expectSaga(vaultSagas)
+      .withReducer(
+        combineReducers({ vault: vaultReducer }) as never,
+        {
+          vault: seeded
+        } as never
+      )
       .provide([
-        [matchers.select.selector(selectEncryptionKeyHash), 'key-hash'],
-        [matchers.select.selector(selectVault), EMPTY_VAULT]
+        [matchers.select.selector(selectEncryptionKeyHash), 'key-hash']
       ])
       .dispatch(windowRequestResponded({ requestId: 'answered' }))
       .silentRun(VAULT_REENCRYPT_DEBOUNCE_MS + 200);
 
     expect(encryptSpy).toHaveBeenCalledTimes(1);
+    // `toEqual` on each map, not `toMatchObject` on the vault: a recursive
+    // subset match tolerates the very key this test exists to prove is gone.
+    const encrypted = encryptSpy.mock.calls[0][1];
+    expect(encrypted.jsonById).toEqual({ other: 'kept' });
+    expect(encrypted.eip712ById).toEqual({});
   });
 
   it('re-encrypts once per edit when edits are spaced beyond the debounce window', async () => {
--- a/src/background/redux/vault/reducer.test.ts
+++ b/src/background/redux/vault/reducer.test.ts
@@ -452,21 +452,29 @@ describe('vault reducer', () => {
       expect(s.eip712ById).toEqual({ other: 'keep', same: 'fresh' });
     });
 
-    // Measured, not assumed. `__proto__` never reaches here in production —
-    // `handleSdkMethod` rejects it at the message boundary for every approval
-    // type (sdk-methods.ts, `isStorableRequestId`) — but the reducer must be
-    // safe on its own: immer's copy ASSIGNS the returned object's keys, and
-    // assigning a string to `__proto__` is a silent no-op, so the entry is
-    // dropped rather than stored, and the map's prototype is untouched.
-    it('drops a __proto__ payload without touching the prototype', () => {
+    // Asserted on what the reducer owns, so a `target` bump cannot move it.
+    // `tsconfig.json` is es2017 today, which emits the spread as `Object.assign`
+    // and makes `__proto__` run the setter rather than add an entry; at es2018
+    // the native spread would store it as an own property instead. Either way
+    // `storePayload` refuses the id, so the map stays empty and keeps its
+    // prototype — and `getPayload` agrees with the map about both.
+    //
+    // The object case is not a variation for its own sake: the deploy path
+    // dispatches `JSON.parse(...)`, i.e. an object, into a map declared
+    // `Record<string, string>` (sdk-methods.ts), and an object value is what
+    // makes the setter actually replace this map's prototype.
+    it.each([
+      ['a string', 'poison'],
+      ['an object, as the deploy path dispatches', { poisoned: true }]
+    ])('refuses a __proto__ payload carrying %s', (_label, json) => {
       const s = reducer(
         initialState,
-        deployPayloadReceived({ id: '__proto__', json: 'poison' })
+        deployPayloadReceived({ id: '__proto__', json: json as string })
       );
 
       expect(Object.keys(s.jsonById)).toEqual([]);
       expect(Object.getPrototypeOf(s.jsonById)).toBe(Object.prototype);
-      expect(({} as Record<string, unknown>).poison).toBeUndefined();
+      expect(getPayload(s.jsonById, '__proto__')).toBeUndefined();
     });
 
     // The ceiling must never cost an ALREADY STORED request its payload: the
--- a/src/background/redux/vault/reducer.test.ts
+++ b/src/background/redux/vault/reducer.test.ts
@@ -22,6 +22,7 @@ import {
   vaultLoaded,
   vaultReseted
 } from './actions';
+import { getPayload } from './payload-map';
 import { MAX_STORED_PAYLOADS, reducer } from './reducer';
 import { VaultState } from './types';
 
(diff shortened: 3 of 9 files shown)


src/background/redux/vault/reducer.ts
```diff
@@ -1,14 +1,57 @@
...
+// an existing key keeps its original position — so the FIFO eviction below drops
+// the request that has been waiting longest, not one that merely refreshed.
+function storePayload(
+  payloads: PayloadMap,
+  requestId: string,
+  json: string
+): PayloadMap {
+  const next: PayloadMap = { ...payloads, [requestId]: json };
+
+  const overflow = Object.keys(next).length - MAX_STORED_PAYLOADS;
```
Comp0te: `storePayload` evicts from the *head* of the map on overflow, and the key space is dapp-controlled with no per-origin accounting — so the entry dropped is the one that has waited longest, which is exactly the long-lived in-flight request this PR exists to keep.

To see it: seed the map with one entry, then apply `storePayload` ten more times with ids the page chooses. `Object.keys(next).slice(0, overflow)` selects the first key, so the seeded entry is gone and the map holds only the ten newest.

Nothing rate-limits that. `signRequest` has no connected-site precondition (`src/background/handlers/sdk-methods.ts:167-232`), the message handlers are concurrent, and `openWindow` only cancels after `windows.create` plus `CANCEL_GRACE_MS = 250` (`src/background/open-window.ts:52-120`), so all eleven writes land before any `windowRequestResponded` can free a slot. The rationale at `:28-29` ("one approval window means 1-2 in-flight requests") holds for a cooperative page, but the cap is reachable by an uncooperative one.

The displaced request's window then has `transactionJson === undefined`, `useFetchDataForSignatureRequest` goes `enabled: false`, and with the new gate at `sign-transaction/index.tsx:413` its Sign button is permanently disabled over an empty pane while the dapp's promise hangs to its own timeout — the failure #1427 and this PR were written to remove, behind a threshold of ten.

Refusing the *incoming* write at capacity (returning `payloads` unchanged) puts the loss on the request the caller controls rather than on the one the user is confirming.

src/apps/signature-request/pages/sign-transaction/index.tsx
```diff
@@ -399,8 +403,14 @@ export function SignTransactionPage() {
           <Button
             color="primaryRed"
             flexWidth
+            // `!signatureRequest` mirrors the eip712 page. `transaction` is
+            // local state that outlives the payload it was parsed from, so
+            // once the answered request's payload is dropped from the vault
+            // this button would stay enabled over the empty pane the content
```
Comp0te, line 409: This comment, and the PR description ("both pages then render the loading state, and both Sign buttons are disabled"), describe a fallback this page doesn't have. `sign-eip712` guards its loading branch on `isLoadingSignatureRequest || !signatureRequest` (`sign-eip712/index.tsx:321`); this page guards on `isLoadingSignatureRequest` alone (`:472`), and the flag behind it is `isFetching` (`use-fetch-data-for-signature-request.ts:26`), which is `false` for a disabled query. With no payload every child of the fragment is guarded on `signatureRequest` (`:476`, `:480-481`, `:490-491`), so the page renders nothing at all — a blank pane under a greyed-out Sign button, no loading state, no text, and Cancel the only exit the user can guess at.

The render branch itself is unchanged here; what's new is that per-request deletion makes reaching it routine. There's no dapp-facing exit either: the two invalid-state branches that can identify a tab (`:151-160`, `:204-227`) both throw *and* forward an SDK error, while this case does neither, so the dapp's promise hangs to its own timeout. (`:128-135` only throws, since `requestTabId` may be null there — so it isn't a counter-example either way.)

Mirroring the eip712 guard, or rendering an explicit "this request is no longer available", would fix the page and the description together.

src/background/redux/sagas/vault-sagas.test.ts
```diff
@@ -449,6 +450,27 @@ describe('updateVaultCipher debounce', () => {
     expect(encryptSpy).toHaveBeenCalledTimes(1);
   });
 
+  // WALLET-1384: the vault reducer deletes an answered request's signing
+  // payload, and that deletion must reach the cipher. Otherwise an MV3
+  // service-worker restart re-loads the entry from a stale blob through
+  // `vaultLoaded`, for a requestId whose `windowRequestResponded` has already
+  // fired — so nothing but the FIFO cap would ever remove it.
+  it('re-encrypts when an answered request drops its payload', async () => {
```
Comp0te: This test never drops a payload. `expectSaga(vaultSagas)` runs the saga in isolation with no `withReducer`, so the vault reducer is never invoked, and the provided `selectVault → EMPTY_VAULT` short-circuits the only read (`vault-sagas.ts:451`) with `jsonById` and `eip712ById` both `{}` (`:89-97`). The sole assertion is `toHaveBeenCalledTimes(1)`.

What it pins is that `windowRequestResponded.type` appears in the debounce array at `vault-sagas.ts:138` — delete the entire `extraReducers` block at `vault/reducer.ts:390-413` and this test still passes, since neither the saga nor the test imports the vault reducer. Nothing else joins the halves: `encryptVault` appears in no other test file, and the one test that builds a real store (`open-window.supersede.test.ts:52-56`) registers only `{ windowManagement: reducer }`. So the contract the saga entry was added to establish — that the vault handed to `encryptVault` is read *after* the reducer applied — is pinned by nothing.

Providing `selectVault → { ...EMPTY_VAULT, jsonById: { answered: 'json' } }` and asserting on `encryptSpy.mock.calls[0][1]` would observe the value rather than the call count; driving the action through the real root reducer (the `get-main-store.test.ts` pattern) would pin it end to end.

src/background/redux/vault/reducer.test.ts
```diff
@@ -400,22 +402,138 @@ describe('vault reducer', () => {
...
+      );
+      expect(s.eip712ById[`id-${MAX_STORED_PAYLOADS}`]).toBe(
+        `e-id-${MAX_STORED_PAYLOADS}`
+      );
+    });
+
+    // Re-writing an existing key keeps its original insertion position, so a
+    // refreshed payload must NOT be treated as the newest one by the FIFO
+    // eviction above — otherwise a long-lived request could outlive a newer one.
+    it('does not move an entry to the back of the queue on rewrite', () => {
```
Comp0te: `it('does not move an entry to the back of the queue on rewrite')` asserts `Object.keys(s.jsonById)` deep-equals `['first','second']`, which holds only because both ids are non-integer-like. Integer-like string keys enumerate first, ascending, regardless of insertion order — `Object.keys({ victim: 1, 0: 1, 1: 1 })` gives `["0","1","victim"]`. `requestId` is chosen at the message boundary, so integer-like ids are representable.

Since `storePayload` slices its victim from `Object.keys(next)`, eviction order isn't insertion order for that id space, and the comment at `reducer.ts:35-37` ("Insertion order = the order requests first stored a payload") carries the same imprecision. The direction is benign — integer-like keys sort to the front, so a page using numeric ids evicts its own entries first — but the test doesn't establish the FIFO property it's named for.

Seeding past the cap and asserting which id survived would pin the consequence directly, for any id space.

src/background/redux/vault/reducer.ts
```diff
@@ -328,6 +377,40 @@ const slice = createSlice({
...
+  // Keyed off the ACTION, not off `windowManagement`'s resulting state: that
+  // reducer no-ops the transition unless the request is currently 'open', which
+  // is exactly the orphan case (an MV3 restart between registration and the
+  // response) where a stale payload most needs dropping.
+  extraReducers: builder => {
+    builder.addCase(
+      windowRequestResponded,
+      (state, { payload: { requestId } }): State => {
+        if (
+          getPayload(state.jsonById, requestId) == null &&
```
Comp0te, line 412: Three guards this PR adds are each removable with the whole suite still green. Same shape, so one thread rather than three.

**1 — this guard** (`:394-397`) reads both maps through `getPayload` so an inherited `Object.prototype` member isn't mistaken for a stored payload, but no test in `reducer.test.ts` uses such an id: the identity case at `:528-537` uses `'unknown'`, which behaves identically under a bare index, and the branch structure is unchanged by a bare-index rewrite so the 100% gate stays green. The sibling reducer already has the case this one omits — `windowManagement/reducer.test.ts:250-264` has an `it.each(['toString','constructor','valueOf','hasOwnProperty'])` plus a `__proto__` case at `:283-292`.

**2 — the two `getPayload` read sites** (`sign-transaction/index.tsx:146`, `sign-eip712/index.tsx:87`). `payload-map.test.ts` covers the helper in isolation; nothing covers either page using it. No test imports `SignTransactionPage` or `SignEip712Page`, and `src/apps/**` is outside `collectCoverageFrom` (`jest.config.js:42-51`), so the branch gate says nothing here. The popup e2e drives the real SDK, so its ids are always v4 UUIDs (`src/content/sdk.ts:145-152`) and can never collide. `isStorableRequestId` rejects only `__proto__` (`request-map.ts:38-40`), so `toString` / `constructor` / `valueOf` / `hasOwnProperty` all reach these reads.

**3 — the `!signatureRequest` gate** (`sign-transaction/index.tsx:411-416`). The enabled direction is incidentally covered by the e2e clicking Sign on a live request; the disabled direction — the one the change exists for — has none, and nothing reaches `signatureRequest === undefined` with `transaction` set. That state is reachable: `sdk-response-to-tab.ts:133-135` dispatches `windowRequestResponded` *before* `await tabs.sendMessage` at `:138`, the reducer drops the payload, the query key changes and `enabled` goes false, while `transaction` (set only at `index.tsx:201`) is never reset. Transient in the responding window, since `closeCurrentWindow()` follows at `:266`/`:283` — durable in a second window holding the same `requestId`, which is where the gate earns its keep.

A pure seam over the payload read — `selectTransactionJsonForRequest(map, requestId)` consumed by both pages — makes 2 and 3 testable without adding jsdom, the same move `register-ledger-permission-window.ts` already makes for a decision that lived in a hook.

Comp0te (commented):
Read against WALLET-1384 and #1427 — the per-request payload map and the prototype-safe read both look right, and the two-window Ledger case is clearly what the design has in mind. The comments below run along one line: the states that appear once a payload is *absent*. One is a way for a page to force that absence for someone else's request, two are about what the UI does when it happens, and one collects the guards that don't yet have tests which would notice them being removed.

src/background/redux/vault/reducer.ts
ost-ptk: Fixed in 5d90388e — `storePayload` now refuses the incoming write at capacity and evicts nothing.

Confirmed the trigger before changing it: `handleSdkMethod` gates approval types on `isStorableRequestId` and a duplicate-id check only (`sdk-methods.ts:69-81`), so `signRequest` needs no connected site; `openWindow` is fire-and-forget and `cancelRequestsDisplacedBy` awaits `CANCEL_GRACE_MS` before it marks anything responded, so a burst does land in full before a slot can free.

A rewrite of an id already stored still applies — it cannot grow the map, and refusing it would strand a request whose payload the dapp re-sent.

One thing I did not do, and want to flag rather than bury: I looked at removing the ceiling entirely instead, and it is not viable — a request that never gets a window attached has `windowIds: []`, which `cancelRequestsDisplacedBy` can never select (`open-window.ts` calls this out as the one state no window event cancels), so a page that strands enough of those would grow the persisted map without bound.

Refusal has its own residual, now documented at the decision rather than assumed away: leaked entries survive lock/unlock, because `lockVaultSaga` flushes the cipher before `deploysReseted` and `vaultLoaded` restores from it. Enough of them fill the map and refuse every later payload. Reclaiming those slots means reconciling the map against the windows actually open — I would rather do that as its own change than approximate it here.

src/apps/signature-request/pages/sign-transaction/index.tsx
ost-ptk: Fixed in 5d90388e — the branch is now `isLoadingSignatureRequest || !signatureRequest`, matching `sign-eip712/index.tsx:321`.

The description said what the eip712 page does, not what this one does; corrected there too. `isFetching` being false for a disabled query is what makes it an empty pane rather than a spinner, and every child of the fragment is guarded on `signatureRequest`, so nothing rendered at all.

Not adding the dapp-facing exit here. The states that forward an SDK error are ones the page can prove are terminal — a missing signing account, an already-signed transaction. This one is not: with the ceiling now refusing the incoming write rather than evicting a stored one, the reachable route to a missing payload is a request that was already answered, where the response is in flight and a second one would be dropped by the dedup anyway. The genuinely unanswerable case is a refused write, and that is better rejected at the message boundary than diagnosed from the page — listed under out-of-scope in the description.

src/background/redux/sagas/vault-sagas.test.ts
ost-ptk: Fixed in 5d90388e — the test now drives the real vault reducer through `withReducer(combineReducers({ vault: vaultReducer }))`, seeds `jsonById: { answered, other }` plus `eip712ById: { answered }`, and asserts the maps handed to `encryptVault` rather than a call count.

Verified red in both directions separately: with the `extraReducers` block deleted, and with `windowRequestResponded.type` removed from the debounce array. Neither half passes alone now.

Worth recording, since it is the same failure mode one level down: my first version asserted `toMatchObject({ jsonById: { other: 'kept' } })` on the encrypted vault and stayed green with `extraReducers` deleted — a recursive subset match tolerates exactly the key the test exists to prove is gone. It is `toEqual` on each map now.

src/background/redux/vault/reducer.test.ts
ost-ptk: Removed in 5d90388e, along with the property it was named for.

With the ceiling refusing the incoming write, enumeration order no longer selects a victim, so there is no FIFO property left to pin and the `Insertion order = …` comment went with it. What replaces it asserts the consequence directly: at capacity the incoming write is refused and `id-0` survives, a rewrite of a stored id still applies, and a slot freed by `windowRequestResponded` is reusable — none of which depend on the id space.

The integer-like-key point stands on its own regardless and is why I did not simply re-anchor the old test on a different pair of ids.

src/background/redux/vault/reducer.ts
ost-ptk: Two of the three are now covered; the third I want to push back on.

**1 — the guard.** Added in 5d90388e: `it.each(['toString','constructor','valueOf','hasOwnProperty'])` in both directions — a payload stored under such an id is dropped by `windowRequestResponded`, and an id in neither map returns the identical state object — mirroring `windowManagement/reducer.test.ts:250-264`.

`__proto__` turned out to differ from the sibling, so it is pinned separately and it corrected a claim of mine. I had written that a computed key defines an own property even for `__proto__`. Measured through the real reducer: the map ends up empty, because the copy immer makes when it finalizes ASSIGNS the keys and assigning a string to `__proto__` is a silent no-op. Prototype untouched, no pollution, but dropped rather than stored. The `payload-map` docblock now says that instead.

**2 and 3 — the read sites and the disabled direction of the gate.** Correct that nothing covers them. I do not think a `selectTransactionJsonForRequest` seam changes that: `getPayload` is already the pure seam and `payload-map.test.ts` already covers it in isolation — what is uncovered is that the page calls it, and a second pure function is equally uncalled by any test. Covering it means rendering the component, and `jest.config.js` sets `testEnvironment: 'node'` with no React component test anywhere in the repo. That is a jsdom + Testing Library setup, which I would rather land as its own change than smuggle into this one; noted as a known gap in the description.

The `register-ledger-permission-window.ts` comparison is fair for the shape, but that extraction moved a decision out of a hook so the background could make it — the seam earned its keep by changing who runs the code, not only by being pure.

src/apps/signature-request/pages/sign-transaction/index.tsx
Comp0te: Fix confirmed at `index.tsx:477-479`, and the PR description now describes this page correctly.

One correction to what I wrote above: "every child below is guarded away" wasn't right for the first child. `LedgerEventView` at `:481-483` is guarded on `signingPageState === SigningPageState.LedgerConfirmation` alone, with no `signatureRequest` term — in the base at `:466-468` as well as here — so it did render under the old `isLoadingSignatureRequest`-only branch and no longer can. Nothing signable is unmasked either way (`handleSign` still returns without `transaction` at `:229-234`) and the Ledger footer is outside this branch entirely (`:375-381`), so this is a rendering change rather than a defect I can demonstrate a user reaching. Flagging it because I overstated the original and the fix followed my wording.

src/background/redux/vault/reducer.ts
Comp0te: Agreed on all three.

**1 — covered.** The `it.each(['toString','constructor','valueOf','hasOwnProperty'])` pair kills the mutant in both directions: rewriting either arm of the guard as a bare index makes `state.jsonById['toString']` resolve to `Function.prototype.toString`, the `&&` goes false, and the identity return is lost.

**2 and 3 — your call, and the premise holds.** `jest.config.js:3` is `testEnvironment: 'node'` and `package.json:141` carries only `@testing-library/jest-dom`, so there is no harness to write these against. Your rebuttal of the `selectTransactionJsonForRequest` seam is right too — a second pure function would be covered in isolation exactly as `payload-map.test.ts` covers `getPayload`, and the uncovered fact, that the page calls it, would stay uncovered. Landing jsdom + Testing Library as its own change is the right shape; noted in the description is enough for me.

On `__proto__` — the measurement is right and the correction is welcome, but the cause isn't immer. Separate note on `reducer.test.ts:467`.

src/background/redux/vault/reducer.ts
```diff
@@ -1,14 +1,74 @@
...
+// fix exists to protect: a request the user is confirming on a Ledger while a
+// page pushes ten more. `signRequest` has no connected-site precondition, the
+// message handlers are concurrent, and a supersede only frees a slot after
+// `windows.create` plus `CANCEL_GRACE_MS`, so a burst lands entirely before any
+// `windowRequestResponded` can. Refusing instead puts the loss on the request
+// the caller controls.
+//
+// A rewrite of an id already present is always applied — it cannot grow the map.
+//
+// Residual, accepted: entries leak when a request is never answered at all
```
Comp0te, line 49: The ceiling is the right call — but the residual described here is reachable by a route this comment doesn't name, and the framing around it doesn't match the base.

**There was no FIFO here before this PR.** The base wrote `jsonById: { [payload.id]: payload.json }` (`reducer.ts:280-293` at `f5bfee4`) — it replaced the whole map on every write, which is WALLET-1384 itself. So this isn't trading FIFO eviction for refusal; it's turning a single-entry map into an accumulating one with a 10-slot ceiling, and the old behaviour only "self-healed" because every write erased the other nine. Worth correcting because `vault-sagas.ts:134`, added by this PR, still tells the next reader that "nothing but the FIFO cap would ever remove it" — there is no FIFO, and the cap now removes nothing.

**One trigger this comment doesn't name.** `:51-53` covers the lock-flush route accurately. The one it misses is an *answered* request: a worker death inside the 500 ms `windowRequestResponded` re-encrypt debounce (`vault-sagas.ts:89, 111-112, 138`) leaks the payload of a request that already got its response — and this PR's own comment at `vault-sagas.ts:128-137` documents that window. "Never answered at all" understates the set.

**It isn't terminal, but the way out isn't findable.** `signRequest` has no lock gate (`sdk-methods.ts:168-234`; the router reaches `handleSdkMethod` at `index.ts:219` with no lock check, unlike `:399`/`:421`/`:506` for other methods), so a request arriving while the vault is locked writes into the emptied in-memory map, and `vaultLoaded`'s `Object.keys(state.jsonById).length === 0 ? jsonById : state.jsonById` (`reducer.ts:106-111`) then keeps that one entry and discards the cipher's ten. Lock, let a dapp send one request, unlock — and the profile signs again. No UI, message or doc points at that sequence.

**What it looks like when it fires.** A refused write still opens the window: `sdk-methods.ts:208-232` dispatches and calls `openWindow` unconditionally, never re-reading state. With `transactionJson === undefined` the query is disabled (`use-fetch-data-for-signature-request.ts:57`), so the branch you changed at `index.tsx:477` renders `SignatureRequestLoading` — a permanent skeleton with a working Cancel, not a blank pane. Ten of those per map (each map has its own ceiling, `reducer.ts:351`, `:358`) and dapp signing is finished on that profile.

The reconciliation you name at `:55-56` is the fix. This is about its reachability while you schedule it.

src/background/redux/vault/reducer.test.ts
```diff
@@ -400,22 +403,201 @@ describe('vault reducer', () => {
...
+    // safe on its own: immer's copy ASSIGNS the returned object's keys, and
+    // assigning a string to `__proto__` is a silent no-op, so the entry is
+    // dropped rather than stored, and the map's prototype is untouched.
+    it('drops a __proto__ payload without touching the prototype', () => {
+      const s = reducer(
+        initialState,
+        deployPayloadReceived({ id: '__proto__', json: 'poison' })
+      );
+
+      expect(Object.keys(s.jsonById)).toEqual([]);
```
Comp0te, line 467: Following up on the docblock you pointed at — the `__proto__` case is pinned by a compiler flag rather than by the reducer, and the mechanism named above it isn't the one operating.

**The mechanism.** `tsconfig.json:3` is `"target": "es2017"`, so `{ ...payloads, [requestId]: json }` is emitted as nested `Object.assign`, and `[[Set]]` on `__proto__` with a string value is the silent no-op — the key is dropped before immer ever sees the object. Both toolchains read that tsconfig (`jest.config.js:5` via ts-jest, `webpack.config.js:287` via ts-loader), so the test and the shipped bundle share the emit. At `target: es2018` the spread is native, the computed key is a `CreateDataProperty`, and `:467` goes red with the reducer unchanged. `payload-map.ts:22-23` carries the same attribution and asserts both "`{ ...map, [id]: json }` does define an own property" and the drop — under this build those two can't both hold. `windowManagement/request-map.ts:30-40` states a mechanism that survives either target.

**The value type.** The fixture is a string, but the deploy path passes an object: `sdk-methods.ts:181` is a bare `let deployJson;`, `:183` is `JSON.parse(action.payload.deployJson)`, and `:208-213` dispatches that object into a map typed `Record<string, string>` (`types.ts:12-13`; action at `reducer.ts:346-352`). It type-checks only because the uninitialised `let` is an evolving `any`. With the real value, the same reducer line sets the map's own prototype — so `:468` isn't covering the type the deploy path actually uses. Not live today: `sdk-methods.ts:74` rejects `requestId === '__proto__'` before anything is dispatched (`request-map.ts:38-40`). The point is where the guarantee is anchored, not an open hole.

**`:469`** reads a fresh object literal independent of `s`, and the fixture's value is a string, so no implementation of this reducer can make it fail.

Asserting what the reducer owns and what holds at any target — `Object.getPrototypeOf(s.jsonById) === Object.prototype` plus `getPayload` agreeing with the map — would survive a `target` bump; and either applying `isStorableRequestId` inside `storePayload`, or storing `JSON.stringify(deployJson)` so the map matches its declared type, would put the write-side guarantee where the docblock claims it.

Comp0te (commented):
Re-reviewed at `5d90388e`. The four points from the first pass are fixed and verified against this head; the guards thread has a reply. Two notes left, both on the new capacity policy: the residual documented at `reducer.ts:49-56` is reachable by a route the comment doesn't name and recovers only by a sequence a user couldn't find, and the `__proto__` test at `reducer.test.ts:467` is pinned by the `es2017` target rather than by the reducer.

src/apps/signature-request/pages/sign-transaction/index.tsx
ost-ptk: Right, and the correction changes the fix — `LedgerEventView` is now excluded from the term (56ffbe6c).

It is the one child that carries no `signatureRequest`, and it is the only thing the Ledger permission window has to show while the device is being read — the second window, the one `cancelRequestsDisplacedBy` spares, i.e. exactly the screen this PR exists for. A skeleton there would be a worse outcome than the empty pane I was fixing. The branch is now `isLoadingSignatureRequest || (!signatureRequest && signingPageState !== SigningPageState.LedgerConfirmation)`, so the skeleton covers the content states only and a genuine fetch still wins.

I had copied "every child below is guarded away" into the code comment as well; it says "the two content children" now.

src/background/redux/vault/reducer.ts
ost-ptk: Anchored in the reducer in 56ffbe6c — `storePayload` refuses the id through `isStorableRequestId` before anything is built.

The emit explanation is right, and I confirmed it rather than take it on trust — `tsc --target es2017` emits nested `Object.assign` for the spread, `es2018` leaves it native:

| | string value | object value |
| --- | --- | --- |
| es2017 (`Object.assign`) | silent no-op, entry dropped | **sets the map prototype** |
| es2018+ (native spread) | own property | own property |

So the old test was pinned by `tsconfig.json:3` and my docblock named the wrong actor. Both are corrected: `payload-map.ts` now points at the guard instead of claiming the write shape settles it, and the test asserts what the reducer owns — empty map, `Object.getPrototypeOf(s.jsonById) === Object.prototype`, and `getPayload` agreeing — which holds at either target.

The value-type half is the part I would have missed. The test now runs both, and the object case is the one that goes red without the guard; the string case passes either way, which is precisely why the old single-value test proved less than it looked. I did not take the other option you named — `JSON.stringify` at the dispatch site — because that is the deploy path's declared-type problem rather than this map's, and it belongs with the `let deployJson` fix listed under out-of-scope.

`:469` was indeed unfalsifiable — a fresh literal, independent of `s`. Gone.

src/background/redux/vault/reducer.ts
ost-ptk: All three taken (56ffbe6c). Two comment corrections and one that changes what the docblock claims.

**The framing.** Correct that the base had no FIFO — it replaced the whole map, which is the bug. The FIFO existed only between the first commit of this PR and the previous round, so describing the change against it was describing the PR to itself. The concrete casualty was the sentence you point at in `vault-sagas.ts`, which was still telling the next reader that a FIFO cap was the only thing that could remove an entry; it now says the ceiling refuses new writes and removes nothing.

**The second trigger.** Taken — "never answered at all" was the route I had in mind and not the set. A worker death inside the 500 ms re-encrypt debounce leaks the payload of a request that already got its response, and this PR added the comment describing that window, so the omission was mine twice over. Both routes are named now.

**The exit.** This is the part I did not know, and it changes how the residual should be read: `vaultLoaded` keeps a non-empty in-memory map and discards the cipher's, and `signRequest` has no lock gate — so lock, one dapp request, unlock, and the accumulated entries are gone. Verified the gate asymmetry against `sdk-methods.ts:399`/`:421`/`:506`, which do check, while the `signRequest` branch does not. It is written into the docblock precisely because nothing in the UI or in a message points at that sequence — a recovery nobody can find is not much better than none, but it does mean the state is not terminal, and I would rather the next reader know that than rediscover it.

Reconciling the map against the windows actually open is still the fix; this round is about making the interim honest rather than pretending the ceiling is free.

src/background/redux/vault/reducer.ts
ost-ptk: Noted on all three — and the `__proto__` mechanism is answered in the thread you opened on `reducer.test.ts:467`, since the correction landed there.

On 2 and 3: taking "noted in the description is enough" as agreement to land the jsdom + Testing Library harness separately rather than stretch this PR into it.

Comp0te (approved):

PR #1440, opened 2026-08-07
fix: WALLET-1391 — report and time out an untracked export window, and scope the Ledger close listener to its own window (WALLET-1389)

ost-ptk:
Two independent defects, both surfaced while reviewing the WALLET-1364 lifecycle PRs. Both are pre-existing on `release/2.7.0`.

- https://make-software.atlassian.net/browse/WALLET-1391
- https://make-software.atlassian.net/browse/WALLET-1389

### WALLET-1391 — an export window that is created but never tracked

`export-keys-window-saga.ts` had no `else` on `if (created.id != null)`. When `windows.create` **resolves without an id** — as opposed to rejecting, which #1428 already handles — the window is on screen but the store never learns of it: no `exportKeysWindowIdChanged`, no `sagaError`, no log. The reuse guard then sees `selectExportKeysWindowId === null`, skips the `windows.getAll` lookup entirely, and goes straight to creating another window. A second window rendering secret key material opens beside the first untracked one, and the pattern repeats on every subsequent click.

Separately (ticket comment, item 1): none of the four `windows.*` calls was bounded. The `catch` added by #1428 handles a rejection, not a hang — and because the watcher uses `takeLeading`, one wedged call drops every subsequent dispatch, leaving the menu item permanently inert until the service worker restarts.

The test named for this defect passed with the defect live, and would have kept passing after a fix: `.not.put.actionType(...)` cannot fail on a missing branch.

### WALLET-1389 — the Ledger close listener fired for every window

`use-ledger.ts` registered `windows.onRemoved.addListener(handleCloseWindow)` with a zero-arg handler. `Windows.onRemoved` is `Events.Event<(windowId: number) => void>`, and a zero-arg listener is arity-assignable to it, so tsc never saw this. The handler fired for the **first window closed anywhere in the browser** and then removed itself, so the real close of the permission window afterwards cleared nothing. It fired both ways: cleared when it should not, then failed to clear when it should.

`ledgerStateCleared` resets the whole slice, so mid-flow this nulls `windowId`, `deploy`, `transaction` and `recipientToSaveOnSuccess`. In dapp flows the auto-close effect and `closeNewLedgerWindowsAndClearState` are both gated on `windowId` and become no-ops, leaving an orphaned Ledger window on screen. In popup flows the permission window *is* the signing screen: `SignWithLedgerInNewWindowPage` reads the cleared selectors and `ledgerAction` returns early, so pressing Continue does nothing — no error, no log.

## The fix

**WALLET-1391, commit 1** — the missing `else`: log and dispatch a `sagaError`, on the same surface #1428 introduced. `ERROR_SOURCE` and the `sagaError` import already existed and `'openExportKeysWindowSaga'` is already a member of `SagaErrorSource`, so this adds no import and no type change. The message deliberately differs from `'Could not open the export window'` — the window *is* on screen. `created` is not logged: a `Windows.Window` can carry `tabs[].url`.

This is report-only. With no id the saga can neither track nor `windows.remove` the first window, so the duplicate is surfaced, not prevented.

**WALLET-1391, commit 2** — one `race` around the whole flow rather than four per-call timeouts: it covers all four call sites plus any future one, needs one user-facing string instead of four, and targets the actual impact (an inert menu item). The `try` body moves into a module-local `runOpenExportKeysWindow`; errors still propagate through `call` into the existing `catch`, so #1428's behaviour is unchanged. `WINDOWS_API_TIMEOUT_MS = 5000` follows `PRIVATE_STATE_FETCH_TIMEOUT_MS`.

> **Reviewing the saga:** the raw diff is dominated by re-indentation from extracting the worker. Read it with `git diff -w`.

**WALLET-1389, commit 3** — the listener moves into `src/hooks/ledger-window-close-listener.ts` and takes the tracked id as an argument.

Note that the ticket's suggested direction names the wrong binding. It proposes comparing against the redux `windowId` "already in scope … a dependency of the same effect", but the effect body only runs when `!windowId`, and the closure is not updated by the `ledgerNewWindowIdChanged(w.id)` dispatch below it — so `removedWindowId !== windowId` is `!== null`, always true, and the listener would early-return forever. Green CI, dead feature. The correct binding is `w.id`, the id of the window this effect just created. Passing it as an argument also keeps the narrowing straightforward instead of relying on a guard narrowing a property inside a later closure.

Extracting rather than fixing in place is what makes it testable: the repo has no React-hook harness (zero `.test.tsx` files, no `@testing-library/react`, no jsdom) and `use-ledger.ts` pulls in `@libs/services/ledger`, which constructs a `Ledger` and touches `navigator.hid` at module scope. Same reasoning as `registerLedgerPermissionWindow`, which was extracted from this same effect. The honest limit: the seam is covered, the wiring is not.

## Verification

`npm run ci-check` — 68 suites / 518 tests green, knip clean, tsc clean. Coverage moved up: `redux/sagas` aggregate 67.57 / 63.21 / 54.05 / 67.59 against its 45 / 35 / 45 / 45 floor; `export-keys-window-saga.ts` at 100 / 86.66 / 100 / 100.

Each new test was mutation-checked — delete the fix, the test goes red:

| Deletion | Red test |
| --- | --- |
| the `else` branch | `reports a created window that cannot be tracked` |
| `console.error(msg, created)` (i.e. logging the window) | `reports a created window that cannot be tracked` |
| `timedOut: delay(WINDOWS_API_TIMEOUT_MS)` from the race | `bounds a hung windows.* call and leaves the menu item usable` |
| `WINDOWS_API_TIMEOUT_MS` 5000 → 50 | `bounds a hung windows.* call and leaves the menu item usable` |
| `spawn` → `fork` | `bounds a hung windows.* call and leaves the menu item usable` |
| the straggler `windows.remove` guard | `closes a window that arrives after another one was already tracked` |
| `if (removedWindowId !== permissionWindowId) return;` | `ignores a window that is not the tracked permission window` |
| `detach()` inside `arm` | `drops the previous listener when armed again` |

An earlier revision of this table listed the timeout as covered when it was not: the test stubbed the `race` effect itself, and redux-saga-test-plan's static `race` provider is shape-blind, so the worker was never entered. The timeout test now drives the real race and stubs only redux-saga's own `delay`.

The eight pre-existing saga tests needed no changes under the `race` — the winner cancels the losing `delay`, so nothing slowed down. The timeout test uses `redux-saga-test-plan`'s static `race` provider rather than fake timers.

`npm run format:check` does not cover these paths (its glob is `src/*`, one level); formatting was verified with an explicit `npx prettier --check` and by `npm run lint`, which enforces `prettier/prettier` at error severity.

## Left out on purpose

- **`closeNewLedgerWindowsAndClearState`** (`use-ledger.ts`) calls `windows.getAll({ windowTypes: ['popup'] })` and removes every popup window, while its own target is created `type: 'normal'`. It therefore never hits its own target — the explicit `windows.remove(windowId)` on the next line does that — and instead closes unrelated approval windows. Real bug, four lines from this fix, different failure mode; filing separately.
- **`create-open-window.ts`**: `if (newWindow.id && !isNewWindow)` has the same shape as WALLET-1391 (and uses truthiness rather than `!= null`), but is not a live bug — its only tracking importer re-checks `window.id == null` and fails the request.
- Moving the ledger clear into `background/handlers/window-removed.ts`. That route is id-correct by construction and would additionally fix the popup-context hole, but it changes semantics and lands in a directory whose branch/function floors have no headroom. Worth its own ticket.
- `locale:extract_pot`. Saga error messages render untranslated by design (`saga-error-banner.tsx` prints `error.message` raw) and a run would rewrite the catalog wholesale.

---

## Review round 2

All four review comments were reproduced before being acted on, and all four were correct.

**The race abandoned a `windows.create` it could not abort.** Verified in isolation: the losing branch cancels the child and a create that resolves afterwards never reaches `exportKeysWindowIdChanged`. That reintroduced the untracked window from slowness alone. The worker is now `spawn`ed detached — `fork` cannot work here, a forked child keeps its parent alive and `takeLeading` stays wedged — so the entry saga bounds only how long it *waits*. A create that lands after the timeout still tracks, and if a retry already tracked a window the straggler closes itself rather than overwriting the id.

**The timeout test proved nothing.** Deleting the bound left it green. It now drives the real race through the watcher with a permanently pending `windows.getCurrent`, stubs only `delayP`, pins the constant as a literal, and asserts the menu item recovers. This also exposed cross-test mock leakage in the file (only an `afterEach` clear); fixed with a top-level `beforeEach`.

**The Ledger listener outlived its flow.** `ledgerStateCleared()` reaches the store from paths that never close the permission window, after which another `useLedger` instance takes over the slice and the stale listener wipes it. The module is now a tracker with `arm`/`detach` so the lifecycle is covered by tests rather than only fixed. The teardown is deliberately *not* the arming effect's cleanup — `windowId` is one of its dependencies and the arm path dispatches `ledgerNewWindowIdChanged`, so that cleanup would remove the listener it had just registered.

**The `console.error` redaction was unpinned.** Now asserted in the shape `register-ledger-permission-window.test.ts` uses.

### DIFF
--- a/src/background/redux/sagas/export-keys-window-saga.test.ts
+++ b/src/background/redux/sagas/export-keys-window-saga.test.ts
@@ -1,6 +1,7 @@
 import * as matchers from 'redux-saga-test-plan/matchers';
 import { expectSaga } from 'redux-saga-test-plan';
-import { throwError } from 'redux-saga-test-plan/providers';
+import { dynamic, throwError } from 'redux-saga-test-plan/providers';
+import { CallEffectDescriptor } from 'redux-saga/effects';
 import { windows } from 'webextension-polyfill';
 
 import {
@@ -11,9 +12,11 @@ import {
   exportKeysWindowIdChanged,
   exportKeysWindowIdCleared
 } from '@background/redux/windowManagement/actions';
+import { selectExportKeysWindowId } from '@background/redux/windowManagement/selectors';
 
 import { openExportKeysWindow } from './actions';
 import {
+  WINDOWS_API_TIMEOUT_MS,
   exportKeysWindowSaga,
   openExportKeysWindowSaga
 } from './export-keys-window-saga';
@@ -23,7 +26,8 @@ jest.mock('webextension-polyfill', () => ({
     getAll: jest.fn(),
     getCurrent: jest.fn(),
     update: jest.fn(),
-    create: jest.fn()
+    create: jest.fn(),
+    remove: jest.fn()
   }
 }));
 
@@ -42,6 +46,12 @@ const countPutsOfType = (allEffects: unknown, type: string) =>
   ).length;
 
 describe('openExportKeysWindowSaga', () => {
+  // The worker is spawned detached, so a test that abandons one leaks its real
+  // windows.* invocations into whatever runs next. Counts must start at zero.
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
   const currentWindow = {
     id: 1,
     state: 'normal',
@@ -170,6 +180,14 @@ describe('openExportKeysWindowSaga', () => {
         .not.put.actionType(exportKeysWindowIdCleared.type)
         .run();
 
+      // Exactly one argument. The saga deliberately does not log `created`:
+      // a Windows.Window can carry tabs[].url, which here is the
+      // download-account-keys route. Nothing else in the tree pins that
+      // redaction, so a later "improvement" to console.error(msg, created)
+      // would otherwise ship green.
+      expect(consoleError).toHaveBeenCalledWith(
+        'openExportKeysWindowSaga: the export window resolved without an id'
+      );
       expect(countPutsOfType(allEffects, sagaError.type)).toBe(1);
     } finally {
       consoleError.mockRestore();
@@ -208,28 +226,95 @@ describe('openExportKeysWindowSaga', () => {
     }
   });
 
-  it('reports a timeout instead of leaving the menu item inert', async () => {
+  it('closes a window that arrives after another one was already tracked', async () => {
     const consoleError = jest
       .spyOn(console, 'error')
       .mockImplementation(() => {});
 
     try {
-      // A static `race` provider settles the race deterministically. Letting
-      // the real delay run would mean a 5s test, and driving it with fake
-      // timers would fight redux-saga's own scheduler.
+      // Three selects run in this flow: the reuse guard, the pre-create
+      // snapshot, and the post-create re-read. Only the last one sees a
+      // tracked id — i.e. the retry the timeout banner invited won the race
+      // while this create was still in flight.
+      let selects = 0;
+      const trackedByTheRetry = 55;
+
       const { allEffects } = await expectSaga(openExportKeysWindowSaga)
         .withState({ windowManagement: { exportKeysWindowId: null } })
-        .provide({ race: () => ({ timedOut: true }) })
-        .put(
-          sagaError({
-            source: 'openExportKeysWindowSaga',
-            message: 'The export window did not open in time; try again'
-          })
-        )
+        .provide([
+          [
+            matchers.select.selector(selectExportKeysWindowId),
+            dynamic(() => (selects++ < 2 ? null : trackedByTheRetry))
+          ],
+          [matchers.call.fn(windows.getCurrent), currentWindow],
+          [matchers.call.fn(windows.create), { id: 99 }],
+          [matchers.call.fn(windows.remove), undefined]
+        ])
+        // Overwriting the tracked id would strand the retry's window instead —
+        // two windows rendering key material is exactly what 1391 is about.
         .not.put.actionType(exportKeysWindowIdChanged.type)
+        .call([windows, windows.remove], 99)
         .run();
 
-      expect(countPutsOfType(allEffects, sagaError.type)).toBe(1);
+      expect(countPutsOfType(allEffects, sagaError.type)).toBe(0);
+    } finally {
+      consoleError.mockRestore();
+    }
+  });
+
+  it('bounds a hung windows.* call and leaves the menu item usable', async () => {
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+    const delaysRequested: unknown[] = [];
+
+    try {
+      // The hang the bound exists for: windows.getCurrent never settles.
+      (windows.getCurrent as jest.Mock).mockReturnValue(new Promise(() => {}));
+
+      // Only redux-saga's own delay is short-circuited, so the real race runs
+      // against the real WINDOWS_API_TIMEOUT_MS. A static `race` provider would
+      // be shape-blind — it replaces the whole effect, the worker is never
+      // entered, and the bound under test is never exercised.
+      const saga = expectSaga(exportKeysWindowSaga)
+        .withState({ windowManagement: { exportKeysWindowId: null } })
+        .provide({
+          call: (
+            effect: CallEffectDescriptor<unknown>,
+            next: () => unknown
+          ) => {
+            if (effect.fn?.name !== 'delayP') return next();
+
+            delaysRequested.push(effect.args[0]);
+
+            return true;
+          }
+        });
+
+      saga.dispatch(openExportKeysWindow());
+      const runPromise = saga.silentRun(100);
+
+      await Promise.resolve();
+      await Promise.resolve();
+      await Promise.resolve();
+
+      // The wedged first attempt must not swallow this one — that inert menu
+      // item is the whole point of the ticket's timeout item.
+      saga.dispatch(openExportKeysWindow());
+
+      const { allEffects } = await runPromise;
+
+      expect(windows.getCurrent).toHaveBeenCalledTimes(2);
+      // Literal on purpose. Asserting against the imported constant is a
+      // tautology — both sides move together, and dropping the bound to 50ms
+      // sails straight through. 5000 is a product-visible wait; changing it
+      // should have to change this line too.
+      expect(delaysRequested).toEqual([5000, 5000]);
+      expect(WINDOWS_API_TIMEOUT_MS).toBe(5000);
+      expect(countPutsOfType(allEffects, sagaError.type)).toBe(2);
+      expect(countPutsOfType(allEffects, dismissSagaErrorsBySource.type)).toBe(
+        2
+      );
     } finally {
       consoleError.mockRestore();
     }
--- a/src/background/redux/sagas/export-keys-window-saga.test.ts
+++ b/src/background/redux/sagas/export-keys-window-saga.test.ts
@@ -262,6 +262,55 @@ describe('openExportKeysWindowSaga', () => {
     }
   });
 
+  it('closes a window that arrives after a hang in getCurrent, not only in create', async () => {
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    try {
+      // The sibling test above hangs in `create`, so the retry lands BETWEEN the
+      // pre-create snapshot and the post-create re-read. Here the hang is one
+      // call earlier — in `getCurrent` — so the retry lands BEFORE the snapshot
+      // too. That ordering is what makes the position of the snapshot, rather
+      // than the comparison itself, the thing under test: read it downstream of
+      // the hang and it already holds the retry's id, so `trackedNow !==
+      // trackedBeforeCreate` is false exactly when a straggler must be caught.
+      let trackedId: number | null = null;
+      const trackedByTheRetry = 99;
+      const arrivedLate = 100;
+
+      const { allEffects } = await expectSaga(openExportKeysWindowSaga)
+        .withState({ windowManagement: { exportKeysWindowId: null } })
+        .provide([
+          [
+            matchers.select.selector(selectExportKeysWindowId),
+            dynamic(() => trackedId)
+          ],
+          [
+            matchers.call.fn(windows.getCurrent),
+            dynamic(() => {
+              // While this worker was parked here, the entry saga's bound fired,
+              // the banner invited a retry, and that retry ran to completion.
+              trackedId = trackedByTheRetry;
+
+              return currentWindow;
+            })
+          ],
+          [matchers.call.fn(windows.create), { id: arrivedLate }],
+          [matchers.call.fn(windows.remove), undefined]
+        ])
+        // Tracking the late window would strand window 99 — the one the user is
+        // actually looking at — with no id in the store to focus or close it by.
+        .not.put.actionType(exportKeysWindowIdChanged.type)
+        .call([windows, windows.remove], arrivedLate)
+        .run();
+
+      expect(countPutsOfType(allEffects, sagaError.type)).toBe(0);
+    } finally {
+      consoleError.mockRestore();
+    }
+  });
+
   it('bounds a hung windows.* call and leaves the menu item usable', async () => {
     const consoleError = jest
       .spyOn(console, 'error')
--- a/src/background/redux/sagas/export-keys-window-saga.test.ts
+++ b/src/background/redux/sagas/export-keys-window-saga.test.ts
@@ -145,15 +145,35 @@ describe('openExportKeysWindowSaga', () => {
       .run();
   });
 
-  it('does not track when the created window has no id', () => {
-    return expectSaga(openExportKeysWindowSaga)
-      .withState({ windowManagement: { exportKeysWindowId: null } })
-      .provide([
-        [matchers.call.fn(windows.getCurrent), currentWindow],
-        [matchers.call.fn(windows.create), {}]
-      ])
-      .not.put.actionType(exportKeysWindowIdChanged.type)
-      .run();
+  it('reports a created window that cannot be tracked', async () => {
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    try {
+      const { allEffects } = await expectSaga(openExportKeysWindowSaga)
+        .withState({ windowManagement: { exportKeysWindowId: null } })
+        .provide([
+          [matchers.call.fn(windows.getCurrent), currentWindow],
+          [matchers.call.fn(windows.create), {}]
+        ])
+        .put(
+          sagaError({
+            source: 'openExportKeysWindowSaga',
+            message:
+              'Could not track the export window; close it before opening another'
+          })
+        )
+        .not.put.actionType(exportKeysWindowIdChanged.type)
+        // Nothing was tracked, so there is nothing to clear: the new branch
+        // must not invent state.
+        .not.put.actionType(exportKeysWindowIdCleared.type)
+        .run();
+
+      expect(countPutsOfType(allEffects, sagaError.type)).toBe(1);
+    } finally {
+      consoleError.mockRestore();
+    }
   });
 
   it('surfaces a saga error when the export window cannot be opened', async () => {
--- a/src/background/redux/sagas/export-keys-window-saga.test.ts
+++ b/src/background/redux/sagas/export-keys-window-saga.test.ts
@@ -208,6 +208,33 @@ describe('openExportKeysWindowSaga', () => {
     }
   });
 
+  it('reports a timeout instead of leaving the menu item inert', async () => {
+    const consoleError = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    try {
+      // A static `race` provider settles the race deterministically. Letting
+      // the real delay run would mean a 5s test, and driving it with fake
+      // timers would fight redux-saga's own scheduler.
+      const { allEffects } = await expectSaga(openExportKeysWindowSaga)
+        .withState({ windowManagement: { exportKeysWindowId: null } })
+        .provide({ race: () => ({ timedOut: true }) })
+        .put(
+          sagaError({
+            source: 'openExportKeysWindowSaga',
+            message: 'The export window did not open in time; try again'
+          })
+        )
+        .not.put.actionType(exportKeysWindowIdChanged.type)
+        .run();
+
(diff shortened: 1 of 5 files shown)


src/background/redux/sagas/export-keys-window-saga.ts
```diff
@@ -29,74 +128,29 @@ export function* openExportKeysWindowSaga() {
...
-            left: windowWidth + xOffset - SURFACE_WIDTH,
-            top: yOffset
-          }
-    );
+    // A rejection is what the catch below handles; a HANG is not. takeLeading
+    // drops every overlapping trigger while this worker is in flight, so one
+    // wedged windows.* call makes the menu item permanently inert until the
+    // service worker restarts. The race bounds the worker so the next click
+    // gets a fresh attempt.
+    const outcome: { timedOut?: true } = yield race({
```
Comp0te: The `race` at 136-139 bounds the whole flow, but when the `timedOut` branch wins, redux-saga cancels `runOpenExportKeysWindow` while `windows.create()` is still in flight — and that promise is not abortable. If it then resolves with a valid id, the cancelled child can no longer reach `yield put(exportKeysWindowIdChanged(created.id))` at line 101, so the window is on screen and permanently untracked.

That's the state WALLET-1391 was filed to close, now reachable from slowness alone rather than from an id-less resolve. `selectExportKeysWindowId` stays null, so the reuse guard at line 36 short-circuits and the next click calls `windows.create` again; `handleWindowRemoved` (`src/background/handlers/window-removed.ts:28`) never has an id to clear, and refocus-the-existing-window is lost for the rest of the session. The banner this branch raises says "The export window did not open in time; try again" — and following that instruction is what opens a second window rendering secret key material, the opposite of what the sibling branch does.

To reproduce: with redux-saga 1.5.0, replicate this parent/child/race shape with a `create` latency above `WINDOWS_API_TIMEOUT_MS` and trigger three times — three windows are created, `exportKeysWindowIdChanged` is dispatched zero times, and three `sagaError` timeouts go out. The happy path is unaffected: `race` returns `{ completed: undefined }` when the child wins, so `outcome.timedOut` is falsy and no spurious banner is raised.

Suggestion: don't abandon the create. Either have `runOpenExportKeysWindow` return the id and dispatch from a path the race cannot cancel, or `fork` the create so a late-arriving id still dispatches `exportKeysWindowIdChanged`, or bound only `getAll`/`getCurrent`/`update` and treat an abandoned `create` as unrecoverable-and-logged rather than silently droppable.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -188,6 +208,33 @@ describe('openExportKeysWindowSaga', () => {
...
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    try {
+      // A static `race` provider settles the race deterministically. Letting
+      // the real delay run would mean a 5s test, and driving it with fake
+      // timers would fight redux-saga's own scheduler.
+      const { allEffects } = await expectSaga(openExportKeysWindowSaga)
+        .withState({ windowManagement: { exportKeysWindowId: null } })
+        .provide({ race: () => ({ timedOut: true }) })
```
Comp0te: `.provide({ race: () => ({ timedOut: true }) })` replaces the entire `race` effect. redux-saga-test-plan's static `race` provider is shape-blind — it intercepts any `race` and never inspects its branches — so `runOpenExportKeysWindow` is never entered and the bound itself is never exercised.

Deleting `timedOut: delay(WINDOWS_API_TIMEOUT_MS)` from `export-keys-window-saga.ts:138` leaves this test green, and so does changing `WINDOWS_API_TIMEOUT_MS` (line 29 of the same file) from 5000 to 50 or to 0 — every provider in this file resolves synchronously, so the `completed` branch always wins the happy-path tests. `.withState(...)` and the `.not.put.actionType(exportKeysWindowIdChanged.type)` at line 229 are checked against a generator that never runs.

Both mutations are checkable directly against redux-saga-test-plan: the shipped race shape and the same shape with the `delay` branch deleted both pass, with `.not.put({type:'INNER_RAN'})` holding in both; and the happy-path provider shapes pass at 5000, 50 and 0 ms alike.

The mutation table in the PR description is accurate as far as it goes — deleting `if (outcome.timedOut)` does turn this red — but it reads as coverage of the timeout, and the timeout is the one thing it doesn't cover. The acceptance criterion this PR exists to satisfy (the menu item recovers after a hung `windows.*` call) is guarded by a test that passes with the bound removed.

Suggestion: drive the real race instead of stubbing it — exercise the watcher in the style of the existing `takeLeading runs the open worker only once…` test, leave `windows.getCurrent` permanently pending, and drive the delay with fake timers or an injected smaller timeout. Then assert both the `sagaError` and that a second `openExportKeysWindow()` dispatched afterwards reaches `windows.getCurrent` a second time — that second assertion is the one the ticket actually asks for.

src/hooks/use-ledger.ts
```diff
@@ -230,12 +231,7 @@ export const useLedger = ({
 
         triggeredRef.current = true;
 
-        const handleCloseWindow = () => {
-          dispatchToMainStore(ledgerStateCleared());
-          windows.onRemoved.removeListener(handleCloseWindow);
-        };
-
-        windows.onRemoved.addListener(handleCloseWindow);
+        windows.onRemoved.addListener(makeLedgerWindowCloseListener(w.id));
```
Comp0te: The listener returned by `makeLedgerWindowCloseListener(w.id)` isn't retained anywhere and this effect returns no cleanup, so the only removal is the self-removal inside the listener — reachable only past the id guard at `ledger-window-close-listener.ts:24`. A listener whose window is never closed stays armed for the life of the document.

That bites hardest from a `signature-request.html` opener (`src/apps/signature-request/pages/sign-transaction/index.tsx:286-302`, and the sign-message equivalent), which runs in a background-created window that persists rather than in the action popup. There, `ledgerStateCleared()` can be dispatched on a path that never closes permission window A: `LedgerDisconnectedFooter` dispatches it on the Connect click at `ledger-disconnected-footer.tsx:66-68` and `:90-92`, and `ledger-footer.tsx:26-30` renders that footer for `LedgerAskPermission` as well as `Disconnected`. Once `windowId` is null a freshly mounted `useLedger` opens permission window B and takes ownership of the slice — the guard at lines 203-208 is a three-way conjunction, so `triggeredRef` only blocks the same hook instance — while A's listener is still registered. Closing the stale A then resets the whole slice (`src/background/redux/ledger/reducer.ts:20` → `initialState`), wiping B's `deploy`/`transaction`/`recipientToSaveOnSuccess`.

The failure is silent: `sign-with-ledger-in-new-window/index.tsx:41-43` returns with no error and no log when those are null. A sign already awaiting the device does complete, since `ledgerAction` captured its bytes at call time — the damage window is a wipe landing *before* `ledgerAction` runs, including on any retry.

This is also the one case the "Left out on purpose" rationale doesn't cover: the id check makes self-removal *correct*, not *guaranteed*, and the path above never closes the window the listener is waiting on.

Suggestion: keep the returned listener identity in a ref and `windows.onRemoved.removeListener(...)` in this effect's cleanup, and additionally drop it when `windowId` transitions to null.

src/background/redux/sagas/export-keys-window-saga.test.ts
```diff
@@ -145,15 +145,35 @@ describe('openExportKeysWindowSaga', () => {
...
-    return expectSaga(openExportKeysWindowSaga)
-      .withState({ windowManagement: { exportKeysWindowId: null } })
-      .provide([
-        [matchers.call.fn(windows.getCurrent), currentWindow],
-        [matchers.call.fn(windows.create), {}]
-      ])
-      .not.put.actionType(exportKeysWindowIdChanged.type)
-      .run();
+  it('reports a created window that cannot be tracked', async () => {
+    const consoleError = jest
```
Comp0te, line 159: This spy silences `console.error` without asserting its arguments, and this is the test that reaches the `else` at `export-keys-window-saga.ts:102-119` — the branch whose own comment at lines 103-108 records a deliberate decision: `created` is not logged because a `Windows.Window` can carry `tabs[].url`, which here is `popup.html#/download-account-keys`. Nothing pins that redaction; grepping the tree for the message and for the comment turns up only the saga source itself — no test, no lint or semgrep rule.

The convention elsewhere in the repo is the opposite: `src/hooks/register-ledger-permission-window.test.ts:59-62` asserts the message *plus* the exact `{ domain, windowId }` payload, and `sdk-response-to-tab.test.ts:191-194`, `attach-window-to-request.test.ts:225-228`, `open-window.test.ts:258-261` and `window-removed.test.ts:68-71` all pin full argument lists. So a later "improvement" to `console.error(msg, created)` would ship green, and a load-bearing redaction reads as incidental.

Suggestion: assert the `console.error` arguments in this test, in the shape `register-ledger-permission-window.test.ts:59-62` already uses. Worth noting the timeout test's spy at line 212 is a different case — that test never enters `runOpenExportKeysWindow`, so `created` isn't in scope at the `console.error` it reaches, and asserting there would pin nothing about window leakage.

Comp0te (commented):
Reviewed the export-keys window saga's new timeout bound and error reporting, the extracted Ledger close listener and its wiring in `useLedger`, and the new tests for both. The through-line of the comments: both mechanisms are right, but each is bounded slightly short of the failure it was meant to close — the export race cancels a `windows.create` it cannot abort, and the now-correctly-scoped Ledger listener outlives the flow that registered it — and the tests that would catch either are stubbed at the seam that matters.

src/background/redux/sagas/export-keys-window-saga.ts
ost-ptk: Confirmed and fixed in 3132cc81. I reproduced the abandonment in isolation before changing anything — parent races `call(child)` against a delay, the delay wins, the child's `put` after the resolved create never arrives:

```
dispatched === ['TIMED_OUT']    // 'TRACKED' never fires
```

Of your three options, only detachment satisfies this comment and the one on the test together — that one asks for the menu item to demonstrably recover, which means the entry saga has to complete. `fork` cannot do it: a forked child keeps its parent alive until it settles, so `takeLeading` stays wedged exactly as before. So the worker is now `spawn`ed detached and the entry saga bounds only how long it *waits*:

```ts
const task: Task = yield spawn(runOpenExportKeysWindow);
const outcome: { timedOut?: true } = yield race({
  completed: join(task),
  timedOut: delay(WINDOWS_API_TIMEOUT_MS)
});
```

Verified both halves against plain `runSaga`: the detached task still dispatches after the wait is abandoned, and `takeLeading` still gates while the parent is parked on `join`. Being detached, the worker also owns its error reporting now — a throw would otherwise reach nobody once the entry saga has moved on.

That leaves the retry you point out: the banner invites a second click, and two creates can be in flight. The worker snapshots the tracked id before creating and re-reads it after, and the late one closes itself rather than overwriting:

```ts
if (trackedNow != null && trackedNow !== trackedBeforeCreate) {
  yield call([windows, windows.remove], created.id);
  return;
}
```

A snapshot rather than a bare null check because the stale-id path above may have just dispatched `exportKeysWindowIdCleared` — comparing against `null` would silently depend on the reducer having already applied it. Covered by `closes a window that arrives after another one was already tracked`; deleting the guard turns it red.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Confirmed. I ran your mutation before touching anything: deleting `timedOut: delay(WINDOWS_API_TIMEOUT_MS)` left all 9 tests green. The PR description's mutation table was accurate but read as coverage of the timeout, which is exactly what it did not cover — that line is corrected.

Replaced with `bounds a hung windows.* call and leaves the menu item usable`, which drives the watcher with `windows.getCurrent` permanently pending and stubs only redux-saga's own delay, so the real race runs:

```ts
.provide({
  call: (effect: CallEffectDescriptor<unknown>, next: () => unknown) => {
    if (effect.fn?.name !== 'delayP') return next();
    delaysRequested.push(effect.args[0]);
    return true;
  }
})
```

It asserts the second assertion you asked for — a second `openExportKeysWindow()` reaches `windows.getCurrent` again — plus the `sagaError` and the dismiss, one of each per attempt.

On pinning the constant: my first attempt asserted against the imported `WINDOWS_API_TIMEOUT_MS`, which is a tautology — both sides move together and 5000 → 50 sailed through exactly as you described. It is a literal now, with `expect(WINDOWS_API_TIMEOUT_MS).toBe(5000)` alongside it.

Your comment also surfaced something in the test file itself. Once the worker is spawned *before* the race, the old shape-blind provider no longer skips it, so that test began invoking the real `windows.getCurrent` mock — and the count leaked into the watcher test below, which had only an `afterEach` clear. Added a top-level `beforeEach(jest.clearAllMocks())`.

Mutations that are now red and were not before: deleting the delay branch, 5000 → 50, `spawn` → `fork`, deleting the straggler guard.

src/hooks/use-ledger.ts
ost-ptk: Confirmed — I traced the path you describe and it holds: `ledger-disconnected-footer.tsx:66-68` and `:90-92` dispatch `ledgerStateCleared()` on the Connect click when `!ledgerTransport`, and `ledger-footer.tsx:26-30` renders that footer for `LedgerAskPermission`, not only `Disconnected`. So the slice is cleared with window A still open and A's listener still armed.

One correction to the suggested shape: the cleanup cannot live on the effect that arms it. `windowId` is one of that effect's dependencies and the arm path dispatches `ledgerNewWindowIdChanged(w.id)`, so the broadcast round-trip re-runs the effect and its cleanup would remove the listener that had just been registered — silent, and worse than the leak. It is two separate effects instead: unmount-only, and the `windowId → null` transition you asked for.

Rather than hold a bare listener in a ref, the module is now a tracker, which makes the lifecycle testable instead of only fixed:

```ts
export interface LedgerWindowCloseTracker {
  arm(permissionWindowId: number): void;
  detach(): void;
}
```

`arm` detaches first, so a second registration can never outlive the first window. Five tests cover it, including the two failures this comment is about — `detaches a listener whose window is never closed` and `drops the previous listener when armed again`.

Honest limit, unchanged from the original PR text: the seam is covered, the wiring is not. Nothing asserts that `useLedger` arms with `w.id` and detaches in those two effects — there is no React-hook harness in the repo (zero `.test.tsx`, no `@testing-library/react`, no jsdom) and `use-ledger.ts` constructs a `Ledger` touching `navigator.hid` at module scope.

src/background/redux/sagas/export-keys-window-saga.test.ts
ost-ptk: Fixed in 3132cc81. Asserted in the shape `register-ledger-permission-window.test.ts:59-62` uses, with the reason in the test so the single argument does not read as an oversight:

```ts
expect(consoleError).toHaveBeenCalledWith(
  'openExportKeysWindowSaga: the export window resolved without an id'
);
```

Mutation-checked: `console.error(msg, created)` turns it red.

Agreed on the timeout test's spy at the other site — `created` is not in scope at the `console.error` that test reaches, so an assertion there would pin nothing about window leakage. Left as a bare silencer.

src/background/redux/sagas/export-keys-window-saga.ts
Comp0te: Thanks — detachment was the right call and it does close the arm I reported. `spawn` + `join(task)` means the worker survives the wait being abandoned, and you're right that `fork` could not have done it without leaving `takeLeading` wedged.

One arm is still open, and it's the snapshot's *position* rather than the guard's logic. `trackedBeforeCreate` is read at :96-98, ten lines after the unbounded `windows.getCurrent` at :83-86 — so the guard only ever sees a retry when the hang was inside `create`. Move the hang into `getCurrent`:

- Worker A parks at :83-86 with nothing tracked. Its reuse-guard read at :42 already returned null, and there's no second read of that path after the hang, so :44-81 — including the stale-id clear at :80 — can no longer preempt it.
- The 5s bound fires and the banner invites a retry.
- Worker B runs through and tracks window 99.
- A's `getCurrent` resolves. A reads `trackedBeforeCreate` at :96 — and it's already 99, because that read is downstream of the hang.
- A creates window 100 and reads `trackedNow` = 99 at :116. `99 !== 99` is false, so the guard at :118 is skipped.
- A reaches `yield put(exportKeysWindowIdChanged(created.id))` at :132 and overwrites 99 with 100.

Window 99 — the one the user is actually looking at — is now untracked. `handleWindowRemoved` (window-removed.ts:28-30) only clears on a match against the tracked id, so it will never clear 99; closing 100 clears the id and leaves 99 open, and the next open makes a third. Two surfaces rendering key material, which is the invariant this PR is about.

The concurrency it needs isn't hypothetical — your own new test asserts it. `bounds a hung windows.* call and leaves the menu item usable` (test.ts:265-321) holds `windows.getCurrent` pending, dispatches twice, and asserts `expect(windows.getCurrent).toHaveBeenCalledTimes(2)` at :307 with two banners at :314. That's two workers concurrently past :83 under a `getCurrent` hang; the only reason it doesn't surface the overwrite is that neither worker ever gets to resume. The straggler test at :229-263 covers the other ordering — `dynamic(() => selects++ < 2 ? null : trackedByTheRetry)` lands the retry *between* the snapshot and the re-read, which is the `create`-hang case.

To reproduce: hold `windows.getCurrent` on a deferred the test controls, dispatch, let the bound fire, dispatch again and let the second attempt track 99, then resolve the deferred. `windows.create` is called twice, `windows.remove` is never called, and the tracked id flips 99 → 100.

Taking the snapshot before `windows.getCurrent` closes it and keeps your rationale intact — it's still a snapshot rather than a null check, so it still doesn't depend on the clear at :80 having been applied by the reducer. (I'd have suggested gating on `trackedNow != null` alone, but your comment at :93-95 is right that it breaks the stale-id case, so the reorder looks like the better of the two.)

One thing the reorder alone doesn't cover, if you'd rather take the class than the case: a hang inside `windows.getAll` at :45 or `windows.update` at :55 produces a straggler whose stale `all` misses the retry's window, so it falls through to the clear at :80 — wiping the retry's tracked id — and then creates and tracks its own, with `trackedNow` null and the guard silent either way.

Comp0te (approved):

PR #1450, opened 2026-08-10
fix: WALLET-1418 — reclaim leaked signature-payload slots

ost-ptk:
[WALLET-1418](https://make-software.atlassian.net/browse/WALLET-1418)

> Follow-up to [WALLET-1384](https://make-software.atlassian.net/browse/WALLET-1384) (#1439, merged). The ceiling this fixes was introduced there, and the defect was raised during that PR's review and deferred by agreement. Base is `develop`.

## The bug

`MAX_STORED_PAYLOADS = 10` refuses the **incoming** write at capacity and evicts nothing — deliberately, so a page cannot displace the request a user is confirming. The consequence is that **nothing reclaimed the slot of a request that was never answered**. The only path that removed an entry was `windowRequestResponded`; the only other clear is the full-slice `deploysReseted`.

After ten leaks on one profile — counted independently per map — every later dapp signature request is refused at the reducer, and the refusal is invisible to its caller: `sdk-methods` dispatched, never re-read state, and called `openWindow` unconditionally. A full approval window opened with `transactionJson === undefined`, the query stayed disabled, and the page rendered a loading skeleton indefinitely while the dapp was told nothing and hung to the SDK's own 30-minute timeout.

### A leak needs no service-worker death

The ticket, and the comments that shipped with #1439, frame the residual as an MV3-restart edge case. It is not — and the branch corrects those comments. Single live worker, **all three targets**:

1. Dapp signs → `jsonById = { r1: … }`.
2. Auto-lock → `lockVaultSaga` runs `updateVaultCipher()` **before** `vaultReseted()`/`deploysReseted()`. `r1` is in the cipher; the in-memory map is `{}`.
3. User closes the approval window → `windows.onRemoved` → `windowRequestResponded(r1)`.
4. The vault reducer finds no payload for `r1` — the map is already empty — and returns the same state object. Nothing is deleted.
5. The debounced `updateVaultCipher` early-returns: `encryptionKeyHash == null`, the vault is locked.
6. Unlock → `vaultLoaded` restores `r1` from the cipher. Permanent.

Firefox and Safari included: `keep-alive.ts` is gated on `isChromeBuild` and both MV2 manifests declare `"background": { "persistent": true }`, so a restart is not available there as an excuse either.

## The fix

Three changes, one per commit, meant to land together.

### 1 · `vaultLoaded` merges instead of replacing

It took the cipher's maps only when the in-memory map was empty and discarded them wholesale otherwise. One request arriving while the vault is locked — `signRequest` has no lock gate — was enough to destroy the payload of everything in flight when the lock fired. That is the WALLET-1384 data loss reached through a different door.

In-memory wins on conflict: `updateVaultCipher` early-returns while locked, so a payload written during a lock exists **only** in memory.

The cipher's keys are filtered through `isStorableRequestId`. `vaultLoaded` is in `FORWARDED_ACTION_TYPES` and its forwarding branch has no `isTrustedUiSender` gate, unlike `windowRequestWindowAttached` — and until now the cipher branch bypassed `storePayload`, so the dapp-controlled-key guard never ran on this path. Replace could only swap the map; merge can grow it.

The merge deliberately does **not** go through `storePayload`: at capacity it refuses the incoming write, so the merge would silently drop entries. The map can therefore sit above the ceiling between this reducer and the reclaim below; during that window a new request is refused, which is the pre-existing at-capacity behaviour and no worse.

### 2 · Reclaim slots no open window can read

On `vaultLoaded`, a saga reconciles both maps against the request ids that can still be read — the `requestId` query params of actually-open extension windows, **unioned with** `selectOpenRequests` — and drops the rest.

**Both halves are load-bearing.**

`windowManagement` is not persisted, so after a service-worker restart the descriptors are gone while the approval window is still on screen and still signable: `LockedRouter` matches `path='*'` **without navigating**, so the window keeps `?requestId=…&origin=…&tabId=…#/sign-deploy` across the lock, and on unlock the sign page re-reads `requestId` from `location.search` and re-derives the payload. Keying the purge on `requests` alone would delete the payload of the request the user is about to confirm — the P0 this design exists to avoid.

Conversely, on window reuse `create-open-window` resolves `tabs.update` when navigation _starts_, so a live window can still report the previous request's URL; there the descriptor is what saves it.

- **Fail closed.** If the enumeration rejects, nothing is purged. Purging on no evidence strands a live request, which is worse than the leak.
- **Reads happen after the round trip**, in one synchronous stretch. `unlockVaultSaga` emits `unlockedEvent` shortly after `vaultLoaded`, dapps retry pending signs on it, and `signRequest` has no lock gate — a keep-set computed before the enumeration is stale by the time it lands. This is pinned by a test, and the pin is mutation-verified: hoisting the reads above the `await` fails exactly that test.
- **No new action creator.** `windowRequestResponded` already deletes the payload keyed off the _action_, already no-ops in `windowManagement` for an id that is not `'open'` (so no tombstone, no `MAX_RESPONDED_TOMBSTONES` slot), and is already in the `updateVaultCipher` debounce list — which is how the reclaim reaches the cipher. A new action would have needed that entry added by hand plus a parity-test exclusion.
- **Only the extension's own pages may carry a `requestId`.** Without that filter a dapp that knows its own leaked ids could hold tabs open at `https://…/?requestId=<id>` and block the reclaim indefinitely, defeating the ticket.
- **No URL, `Window` or `Tab` reaches a log.** A `signMessage` approval URL embeds the user's plaintext message as a query param.
- The enumeration is skipped entirely when both maps are empty — the common unlock — so a no-op unlock does not pull every tab URL in the browser into the extension process.

### 3 · A refused write answers the dapp

`sdk-methods` re-reads the map after dispatching and returns the method-correct SDK response **before** `windowRequestOpened`/`openWindow`, so no descriptor and no window exist for a request that cannot render. With the reclaim in place this should be unreachable; the point is that the residual is visible rather than silent.

Read through `getPayload`, never a bare index: `requestId` is dapp-controlled and `'constructor'` clears both entry guards, so an inherited `Object.prototype` member would make the guard conclude the write succeeded. The presence test is `== null`, not truthiness — the deploy path stores a parsed object in a map declared `Record<string, string>`, and a payload of `0` is a legitimate stored value. **Both properties are pinned, and both pins are mutation-verified**: a bare index kills the two prototype-name tests, a truthiness check kills the two falsy-payload tests.

The eip712 branch answers with `signTypedDataResponse`, not `signResponse`. Response routing is by `requestId` alone, so a `Sign:Response` _would_ resolve the promise — with a value that violates `SignTypedDataResult`, whose `signature`, `digest`, `publicKey` and `error` are all required. `buildCancelResponse` settles the same fork the same way.

#### Why the refusal carries a reason where a user cancel does not

The deploy branch returns `{ cancelled: true, message: … }`, the same shape its immediate sibling (`isDeployAlreadySigningWithThisAccount`) already uses. `SignTypedDataResult` has no `message` field, so the eip712 branch puts the reason in `error`, its only free-text slot — the two branches mirror each other rather than diverging.

That does mean a `cancelled: true` now arrives with a non-null `error`, where every other producer pairs it with `error: null`. Deliberate: those others are user cancels, which have no reason to report, and a dapp benefits from telling "the user declined" apart from "the wallet refused and the user never saw it". `errorCode` was the machine-readable alternative, but it is declared in `SignTypedDataResult` and used nowhere in the codebase — no producer, no consumer — so putting the first value in it would invent a convention for a single call site.

### 4 · Survive a cipher written before a payload map existed

Found by review of the three commits above, and the reason they are not on their own.

`sanitizePayloadMap` called `Object.entries` on whatever the cipher held. `eip712ById` entered `VaultState` on 2026-06-17, so every vault created before v2.5.0 decrypts without it — `decryptVault` is a bare `JSON.parse(...) as VaultState` and nothing migrates. The throw lands inside the `vaultLoaded` put, re-enters `unlockVaultSaga` and is swallowed by its catch: `vaultUnlocked()` never runs, the wallet stays locked behind a TypeError banner, and every retry repeats it. Recovery would have meant resetting the vault.

The guard lives inside the helper rather than at the call sites: the cipher reaches this reducer as an unchecked cast, so the compiler can never demand it at a new call site, and both merge paths funnel through here. Applied to `jsonById` too — the same argument covers the next field added.

**This also repairs a latent break that predates the branch.** On `develop` today a legacy cipher leaves `eip712ById` as `undefined`, and both `getPayload` and `storePayload` throw on it. It now lands as `{}`.

The same commit carries two smaller review findings: the payload dispatch and `windowRequestOpened` must stay in one synchronous block — otherwise the reclaim can resume between them, see a payload with no descriptor and no window, and purge a request that is about to open one — now pinned by two tests that queue a microtask at the payload dispatch and assert it has not run by the descriptor; and `redactUrlQuery` had grown a second copy, so it is extracted into a module that imports nothing. It is the redaction that keeps a plaintext `signMessage` out of the logs, which is the last thing that should be allowed to drift.

## Tests

Every new test was confirmed red against the code it pins, and the load-bearing ones were additionally mutation-checked.

- `vaultLoaded`: an entry from each side survives the merge; in-memory wins on a shared id; `__proto__` from the cipher is dropped with both a string and an object value, asserted on what the reducer owns; the four other inherited names survive as own properties.
- Reclaim: a payload whose id appears in an open window's tab URL survives; one covered only by `selectOpenRequests` survives; one covered by neither is purged from both maps; a rejected enumeration purges nothing; a non-extension tab carrying the same `requestId` does **not** keep a payload alive; a tab with no url, an unparseable url and an empty `requestId` are each skipped without aborting; the keep-set is computed after the round trip; the deletion reaches the cipher.
- Acceptance test for the ticket: ten leaked entries, no open windows, no open requests → after `vaultLoaded` both maps are empty and a fresh `deployPayloadReceived` is accepted. Red before the reclaim, green after.
- SDK boundary: at capacity the response is returned and neither `windowRequestOpened` nor `openWindow` is reached, per map; plus the two invariant pins above.

## Verification

- `npm run ci-check` — green (format, lint, tsc, knip, 82 suites / 688 tests). `redux/vault/reducer.ts` stays at 100% across all four metrics; handlers and sagas both above their floors.
- `npm run dependency:circular` — 140 cycles, unchanged from the base.
- `npm run lint` — 167 warnings, 0 errors, unchanged from the base.

## Known gaps, stated rather than hidden

- The reclaim runs at `vaultLoaded`, i.e. at unlock. A request that is still legitimately `'open'` keeps its slot by design, so the terminal paths that never mark `'responded'` can still hold one within a single unlocked session — bounded and non-permanent, and a different bug. Follow-up ticket.
- Three browser behaviours are asserted from source, not from a build: whether Chrome really stops the worker while an approval window sits on the unlock screen; whether Safari populates `tab.url` for its own `safari-web-extension://` pages under `"tabs"` alone; and whether Firefox's `windows.getAll` omits only the URL _hash_ (`requestId` is a search param, so it should be unaffected). Every one of them fails safe: with no window evidence the design degrades to `selectOpenRequests`, which is exact on the two MV2 targets because their background page is persistent and the descriptors are never lost.

[WALLET-1418]: https://make-software.atlassian.net/browse/WALLET-1418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WALLET-1384]: https://make-software.atlassian.net/browse/WALLET-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -36,6 +36,7 @@ import { emitSdkEventToActiveTabsWithOrigin } from '@background/utils';
 import { SiteNotConnectedError, WalletLockedError } from '@content/sdk-errors';
 import { sdkEvent } from '@content/sdk-event';
 import { SdkMethod, sdkMethod } from '@content/sdk-method';
+import { SdkErrorCode } from '@content/sdk-types';
 
 import { encryptAsHexWithCasperPublicKey } from '@libs/crypto';
 
@@ -53,6 +54,20 @@ const APPROVAL_REQUEST_TYPES: ReadonlySet<string> = new Set([
   sdkMethod.signTypedDataRequest.type,
   sdkMethod.decryptMessageRequest.type
 ]);
+
+const CAPACITY_REFUSAL_MESSAGE = 'Too many pending signature requests';
+
+// The dapp half of a capacity refusal is `errorCode`; this is the extension
+// half, so a wallet refusing every signature is not invisible on this side
+// either. Identifiers only — never the deploy, the typed data or the origin.
+// Log-only rather than `sagaError`: this path is dapp-triggerable, and a banner
+// mounted over every app surface would be a page's to spam.
+function reportCapacityRefusal(action: SdkMethod) {
+  console.error(
+    'sdk-methods: pending-payload map at capacity, request refused',
+    { requestId: action.meta.requestId, method: action.type }
+  );
+}
 
 export async function handleSdkMethod(
   action: SdkMethod,
@@ -231,12 +246,15 @@ export async function handleSdkMethod(
         action.meta.requestId
       ) == null
     ) {
+      reportCapacityRefusal(action);
+
       return {
         handled: true,
         response: sdkMethod.signResponse(
           {
             cancelled: true,
-            message: 'Too many pending signature requests'
+            message: CAPACITY_REFUSAL_MESSAGE,
+            errorCode: SdkErrorCode.tooManyPendingRequests
           },
           { requestId: action.meta.requestId }
         )
@@ -327,6 +345,8 @@ export async function handleSdkMethod(
         action.meta.requestId
       ) == null
     ) {
+      reportCapacityRefusal(action);
+
       return {
         handled: true,
         response: sdkMethod.signTypedDataResponse(
@@ -335,7 +355,8 @@ export async function handleSdkMethod(
             signature: null,
             digest: null,
             publicKey: null,
-            error: 'Too many pending signature requests'
+            error: CAPACITY_REFUSAL_MESSAGE,
+            errorCode: SdkErrorCode.tooManyPendingRequests
           },
           { requestId: action.meta.requestId }
         )
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -20,8 +20,11 @@ import {
   eip712PayloadReceived,
   siteDisconnected
 } from '@background/redux/vault/actions';
+import { getPayload } from '@background/redux/vault/payload-map';
 import {
   selectAccountNamesByOriginDict,
+  selectDeploysJsonById,
+  selectEip712JsonById,
   selectIsAccountConnected,
   selectVaultActiveAccount
 } from '@background/redux/vault/selectors';
@@ -217,6 +220,27 @@ export async function handleSdkMethod(
       })
     );
 
+    // At capacity `storePayload` refuses the INCOMING write. Answering here is
+    // what makes that residual visible: without it the window opens on a
+    // payload the page can never read.
+    if (
+      getPayload(
+        selectDeploysJsonById(store.getState()),
+        action.meta.requestId
+      ) == null
+    ) {
+      return {
+        handled: true,
+        response: sdkMethod.signResponse(
+          {
+            cancelled: true,
+            message: 'Too many pending signature requests'
+          },
+          { requestId: action.meta.requestId }
+        )
+      };
+    }
+
     store.dispatch(
       windowRequestOpened({
         requestId: action.meta.requestId,
@@ -292,6 +316,28 @@ export async function handleSdkMethod(
         json: JSON.stringify({ typedData, options })
       })
     );
+
+    // Same refusal as the deploy branch above, on the other map.
+    if (
+      getPayload(
+        selectEip712JsonById(store.getState()),
+        action.meta.requestId
+      ) == null
+    ) {
+      return {
+        handled: true,
+        response: sdkMethod.signTypedDataResponse(
+          {
+            cancelled: true,
+            signature: null,
+            digest: null,
+            publicKey: null,
+            error: 'Too many pending signature requests'
+          },
+          { requestId: action.meta.requestId }
+        )
+      };
+    }
 
     store.dispatch(
       windowRequestOpened({
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -213,6 +213,8 @@ export async function handleSdkMethod(
       };
     }
 
+    // No `await` before `windowRequestOpened` below: `reconcileStalePayloadsSaga`
+    // purges a payload that no descriptor and no window claims.
     store.dispatch(
       deployPayloadReceived({
         id: action.meta.requestId,
@@ -310,6 +312,7 @@ export async function handleSdkMethod(
 
     const { signingPublicKeyHex, typedData, options } = action.payload;
 
+    // Same synchronous block as the deploy branch, for the same reason.
     store.dispatch(
       eip712PayloadReceived({
         id: action.meta.requestId,
--- a/src/background/open-request-windows.test.ts
+++ b/src/background/open-request-windows.test.ts
@@ -1,0 +1,173 @@
+import { windows } from 'webextension-polyfill';
+
+import { collectRequestIdsFromOpenWindows } from './open-request-windows';
+
+jest.mock('webextension-polyfill', () => ({
+  windows: { getAll: jest.fn().mockResolvedValue([]) },
+  runtime: { getURL: jest.fn(() => 'chrome-extension://abcdefghijklmnop/') }
+}));
+
+const mockWindowsGetAll = windows.getAll as jest.Mock;
+
+const approvalWindowUrl = (requestId: string) =>
+  `chrome-extension://abcdefghijklmnop/signature-request.html` +
+  `?requestId=${requestId}&origin=https%3A%2F%2Fdapp.example&tabId=7` +
+  `#/sign-deploy`;
+
+beforeEach(() => {
+  mockWindowsGetAll.mockReset().mockResolvedValue([]);
+  jest.spyOn(console, 'error').mockImplementation(() => undefined);
+});
+
+afterEach(() => {
+  jest.restoreAllMocks();
+});
+
+describe('collectRequestIdsFromOpenWindows', () => {
+  it('asks for populated windows — without tabs there is no URL to read', async () => {
+    await collectRequestIdsFromOpenWindows();
+
+    expect(mockWindowsGetAll).toHaveBeenCalledWith({ populate: true });
+  });
+
+  it('collects the requestId out of an approval window URL, hash and all', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      { id: 1, tabs: [{ url: approvalWindowUrl('req-1') }] }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(
+      new Set(['req-1'])
+    );
+  });
+
+  // The Ledger permission window carries the same requestId as the approval
+  // window it was opened from (`src/hooks/use-ledger.ts`).
+  it('unions ids across windows and across tabs, de-duplicating repeats', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      {
+        id: 1,
+        tabs: [
+          { url: approvalWindowUrl('req-1') },
+          { url: approvalWindowUrl('req-2') }
+        ]
+      },
+      { id: 2, tabs: [{ url: approvalWindowUrl('req-1') }] }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(
+      new Set(['req-1', 'req-2'])
+    );
+  });
+
+  it('falls back to pendingUrl while a tab is still navigating', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      { id: 1, tabs: [{ pendingUrl: approvalWindowUrl('req-pending') }] }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(
+      new Set(['req-pending'])
+    );
+  });
+
+  it('ignores windows and tabs that carry no usable url', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      { id: 1 },
+      { id: 2, tabs: [] },
+      { id: 3, tabs: [{}, { url: '' }] },
+      {
+        id: 4,
+        tabs: [{ url: 'chrome-extension://abcdefghijklmnop/popup.html' }]
+      }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(new Set());
+  });
+
+  it('skips an unparseable url and keeps reading the rest', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      { id: 1, tabs: [{ url: 'not a url' }] },
+      { id: 2, tabs: [{ url: approvalWindowUrl('req-1') }] }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(
+      new Set(['req-1'])
+    );
+  });
+
+  it('ignores an empty requestId param, which keeps no slot alive', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      {
+        id: 1,
+        tabs: [
+          { url: 'chrome-extension://abcdefghijklmnop/popup.html?requestId=' }
+        ]
+      }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(new Set());
+  });
+
+  // `requestId` is dapp-chosen, so without an origin check any page could put a
+  // `?requestId=` of its choosing in its own URL and pin that slot forever.
+  it('ignores a requestId carried by a page that is not one of ours', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      {
+        id: 1,
+        tabs: [{ url: 'https://evil.example/?requestId=req-1' }]
+      }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(new Set());
+  });
+
+  it('reads our own pages while ignoring a foreign tab that spoofs the same id', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      { id: 1, tabs: [{ url: 'https://evil.example/?requestId=spoofed' }] },
+      { id: 2, tabs: [{ url: approvalWindowUrl('req-1') }] }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(
+      new Set(['req-1'])
+    );
+  });
+
+  // `runtime.getURL('')` carries this extension's id, so the prefix test is
+  // per-extension, not merely per-protocol.
+  it('ignores an extension page belonging to a different extension', async () => {
+    mockWindowsGetAll.mockResolvedValue([
+      {
+        id: 1,
+        tabs: [
+          {
+            url: 'chrome-extension://zzzzzzzzzzzzzzzz/signature-request.html?requestId=req-1'
+          }
+        ]
+      }
+    ]);
+
+    expect(await collectRequestIdsFromOpenWindows()).toEqual(new Set());
+  });
+
+  // `null` is "no evidence", not "no window holds a request": the caller must
+  // not purge on it.
+  it('returns null rather than an empty set when the enumeration rejects', async () => {
+    mockWindowsGetAll.mockRejectedValue(new Error('boom'));
+
+    expect(await collectRequestIdsFromOpenWindows()).toBeNull();
+  });
+
+  it('never puts a window URL into the log when the enumeration rejects', async () => {
+    const consoleError = jest.spyOn(console, 'error');
+    mockWindowsGetAll.mockRejectedValue(
+      new Error(
+        'failed for chrome-extension://abcdefghijklmnop/signature-request.html?message=my-secret-message'
+      )
+    );
+
+    await collectRequestIdsFromOpenWindows();
+
+    const logged = JSON.stringify(consoleError.mock.calls);
+    expect(logged).not.toContain('my-secret-message');
+    expect(logged).not.toContain('?');
+  });
+});
--- a/src/background/open-request-windows.test.ts
+++ b/src/background/open-request-windows.test.ts
@@ -96,6 +96,10 @@ describe('collectRequestIdsFromOpenWindows', () => {
 
   // A silently skipped tab purges a live request with nothing pointing at the
   // cause, so the skip is logged — redacted, never the raw url.
+  //
+  // Asserted positively as well as negatively: `new URL`'s TypeError says only
+  // `Invalid URL`, so absence assertions alone would still pass on a line that
+  // names no tab at all and leaves the operator where they were.
   it('logs the skipped url without its query string', async () => {
     const consoleError = jest.spyOn(console, 'error');
     mockWindowsGetAll.mockResolvedValue([
@@ -105,8 +109,12 @@ describe('collectRequestIdsFromOpenWindows', () => {
     expect(await collectRequestIdsFromOpenWindows()).toEqual(new Set());
 
     expect(consoleError).toHaveBeenCalledTimes(1);
+    expect(consoleError).toHaveBeenCalledWith(
+      'collectRequestIdsFromOpenWindows: could not parse a tab url',
+      'http://[::1/',
+      expect.any(String)
+    );
     const logged = JSON.stringify(consoleError.mock.calls);
-    expect(logged).toContain('could not parse a tab url');
     expect(logged).not.toContain('my-secret-message');
     expect(logged).not.toContain('?');
   });
@@ -235,6 +243,11 @@ describe('collectRequestIdsFromOpenWindows', () => {
     expect(await collectRequestIdsFromOpenWindows()).toBeNull();
   });
 
+  // The argument's TYPE is asserted, not only its content. A plain `Error` has
+  // no enumerable own properties, so `JSON.stringify` renders it `{}` and the
+  // two absence assertions below hold whatever the code does — dropping
+  // `redactUrlQuery` and logging the error object would keep them green while
+  // the console started printing the untruncated message.
   it('never puts a window URL into the log when the enumeration rejects', async () => {
     const consoleError = jest.spyOn(console, 'error');
     mockWindowsGetAll.mockRejectedValue(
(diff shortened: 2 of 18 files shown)


Comp0te (commented):
Reviewed the reclaim path this PR adds — the keep-set built from open windows, the `vaultLoaded` merge, the capacity refusal, and the tests around them — against what WALLET-1418 asked for. The comments share one through-line: in each place the guarantee is a little weaker than the code around it claims. The keep-set trusts more URLs than its docstring describes, the merge's ceiling depends on a saga that has three silent early returns, the capacity refusal reaches the dapp as a user rejection, and two tests named for the URL-parse guard never reach it.

src/background/open-request-windows.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const requestIds = new Set<string>();
+  const extensionUrlPrefix = runtime.getURL('');
+
+  for (const browserWindow of allWindows) {
+    for (const tab of browserWindow.tabs ?? []) {
+      // `pendingUrl` (Chrome-only) is set while `url` is still empty.
+      const url = tab.url || tab.pendingUrl;
+
+      if (!url || !url.startsWith(extensionUrlPrefix)) {
```
Comp0te: The keep-set accepts any tab whose URL starts with the extension origin — the path is never inspected. `sdk.bundle.js` is web-accessible on all three targets (`manifest.v3.json:61-66` with `matches: ["<all_urls>"]` and no `use_dynamic_url`; `manifest.v2.json:62`; `manifest.v2.safari.json:58`), and a `requestId` is page-generated (`payload-map.ts:8-9`). So a page that opens a top-level tab at `chrome-extension://<id>/sdk.bundle.js?requestId=<an-id-it-chose>` puts an id of its choosing into the set: `windows.getAll({ populate: true })` at line 19 enumerates ordinary tabs (the `tabs` permission is present on all three manifests), and the returned set seeds `keep` at `vault-sagas.ts:296-315`, where only ids absent from `keep` are treated as orphans at `:319-326`. The extension id isn't a secret on any target — the content script inserts `runtime.getURL('sdk.bundle.js')` into the page DOM at `content/index.ts:191-206`, so the page can read it before the tag is removed.

A dapp knows the `requestId`s it generated, so once leaked payloads have accumulated it can keep all ten `jsonById` slots pinned with background tabs at that URL, and `reconcileStalePayloadsSaga` reclaims nothing on any later unlock.

One link I can't settle from the repo alone is browser navigation policy — whether a page may put a *top-level* tab at a web-accessible-resource URL. Everything on the extension's side of that line holds, and it's tab-level only (an iframe at a WAR URL wouldn't be enumerated). Independent of that question, the docstring above claims a stronger property than line 37 enforces. The approval pages are a closed set — `connect-to-app.html`, `signature-request.html`, `import-account-with-file.html` (`create-open-window.ts:24-44`) — so matching the pathname as well as the origin would make the check say what the docstring says.

src/background/open-request-windows.ts
```diff
@@ -0,0 +1,57 @@
...
+        continue;
+      }
+
+      let requestId: string | null;
+
+      try {
```
Comp0te, line 56: Separately from the origin check above — this `try`/`catch` is unreachable as the code is ordered today, and the two tests named for it never enter it. `open-request-windows.test.ts:86` ("skips an unparseable url and keeps reading the rest") and `vault-sagas.test.ts:727` ("skips a tab whose url cannot be parsed without aborting the run") both feed `{ url: 'not a url' }`, which fails the `startsWith(extensionUrlPrefix)` check at line 37 and `continue`s one branch earlier — so both end up asserting the extension-origin filter a third and fourth time. WHATWG `new URL` accepts everything that clears the prefix check (invalid percent-escapes, lone surrogates, `[bad]` all parse).

`npx jest src/background/open-request-windows.test.ts --coverage --collectCoverageFrom='src/background/open-request-windows.ts'` reports line 47 (the `continue` inside the catch) uncovered with both exercising suites in the run, and `jest.config.js:42-51` leaves this file outside `collectCoverageFrom`, so the repo's coverage gate wouldn't surface the gap either.

Worth fixing because the guard could easily become live: refactoring the origin check to the natural `new URL(url).origin` form does it, and both tests keep passing without ever entering the catch. A tab silently skipped here is a live request the saga treats as an orphan and purges — the WALLET-1418 symptom, with no log line pointing at the cause — and this is the only fallible step in the PR that stays silent (the enumeration failure eight lines up logs via `redactUrlQuery`, and the saga's catch logs and raises a `sagaError`). Either give one test a fixture that starts with the extension prefix and still fails `new URL()`, or drop the try/catch and rename both tests to what they actually pin; if the guard stays, log before `continue` with the `redactUrlQuery` helper already imported here.

src/background/redux/vault/reducer.ts
```diff
@@ -132,12 +149,8 @@ const slice = createSlice({
       accounts,
       activeAccountName,
       secretPhrase,
-      jsonById:
-        Object.keys(state.jsonById).length === 0 ? jsonById : state.jsonById,
-      eip712ById:
-        Object.keys(state.eip712ById).length === 0
-          ? eip712ById
-          : state.eip712ById
+      jsonById: { ...sanitizePayloadMap(jsonById), ...state.jsonById },
```
Comp0te: This merge is the one writer to `jsonById`/`eip712ById` that isn't bounded: `sanitizePayloadMap` (`:105-111`) only filters keys, so the union of the cipher map and the in-memory map — each capped at `MAX_STORED_PAYLOADS` by `storePayload` (`:82-99`, ceiling at `:37`) — can hold up to twice the ceiling.

The comment says that lasts "until `reconcileStalePayloadsSaga` runs", but the saga has three paths that return without reclaiming: both maps empty at entry (`vault-sagas.ts:287-292`), `collectRequestIdsFromOpenWindows` returning `null` (`:300-302`), and its catch (`:342-353`). It's a `takeLatest` on `vaultLoaded` (`:100`) with no retry. Meanwhile `signRequest` has no lock gate — `background/index.ts:218-226` routes every SDK method unconditionally — so in-memory entries can be written while the vault is locked and the cipher already holds leaked ones.

The practical case is the second path: one failed `windows.getAll` on unlock and the map stays at or above the ceiling for the rest of the session, with `sdk-methods.ts:228-244` / `:324-343` refusing whenever `getPayload(...) == null`, which `storePayload:91-96` guarantees for any new id once the map is full. Every dapp then gets "Too many pending signature requests" until the user locks and unlocks again and the enumeration happens to succeed. This PR also removes the previous escape hatch (the replace-if-empty rule at `reducer.ts:135-141` on `develop`), so there's no in-session recovery left when that happens.

The fail-closed design looks right; it's the ceiling that shouldn't depend on the saga succeeding. Bounding the merge itself — keep all in-memory entries, then take cipher entries only up to `MAX_STORED_PAYLOADS` — makes no single path able to leave the map over the ceiling.

src/background/handlers/sdk-methods.ts
```diff
@@ -210,13 +213,36 @@ export async function handleSdkMethod(
...
+    // payload the page can never read.
+    if (
+      getPayload(
+        selectDeploysJsonById(store.getState()),
+        action.meta.requestId
+      ) == null
+    ) {
+      return {
+        handled: true,
+        response: sdkMethod.signResponse(
```
Comp0te, line 253: The capacity refusal goes back to the page as `signResponse({ cancelled: true, message: ... })` here, and as `signTypedDataResponse({ cancelled: true, ..., error: ... })` at `:330-343` — the envelope this codebase otherwise uses for a user decision (`cancel-requests.ts:49-65`, `sign-transaction/index.tsx:296-302`, `sign-eip712/index.tsx:183-193`, and the message/decrypt pages all pair `cancelled: true` with `error: null`). In fairness the shape isn't new here: the pre-existing refusal at `:203-214` ("This deploy already sign by this account") does the same, and `sdk.ts:35-39` declares `cancelled: true; message?: string` — so this follows an in-file convention rather than inventing one.

What's new is that this refusal is machine-generated and the user never saw a window at all. A dapp doing the usual `if (res.cancelled) → the user rejected it` will tell its user they rejected a signature they were never shown, and retry logic keyed on `cancelled` won't retry. Nothing surfaces on the extension side either — both refusal blocks return before `windowRequestOpened`/`openWindow` and dispatch no `sagaError` — so neither side can learn the wallet is at capacity, which is close to the invisible refusal WALLET-1418 set out to remove.

`sdkMethod.signError` / `signTypedDataError` already exist (`content/sdk-method.ts:50, 69`, routed at `content/index.ts:43-48`), so this path has an envelope of its own available. If the wire shape must not change, assigning `errorCode` on this path and documenting `cancelled + errorCode` as the refusal shape would at least give a dapp something to branch on.

src/background/open-request-windows.ts
ost-ptk: Fixed in 979e66ed — the check now matches the pathname as well, against `REQUEST_BEARING_PATHNAMES`: `/signature-request.html`, `/connect-to-app.html`, `/popup.html`.

I did not use `new URL(url).origin`. `chrome-extension:` is a non-special scheme, so WHATWG gives it the opaque origin `"null"` — `new URL('chrome-extension://a/').origin === new URL('chrome-extension://zzzz/').origin` is `true`, and the per-extension-id property would have been lost silently while every test kept passing. Compared on `protocol` + `host` instead, with a line saying why.

Two notes on the set:

`popup.html` is in it even though nothing payload-bearing serves from it today. It is `use-ledger.ts`'s default `domain` and its `requestId` param is optional, so a future payload-bearing flow that combines them compiles green and gets its payload purged mid-signing — the P0 this design exists to prevent. It appears in no manifest's `web_accessible_resources` (checked all three: v3 has `sdk.bundle.js`, both v2 have `logo64.png` + `sdk.bundle.js`, and the string `popup.html` does not occur in any of them), so listing it does not reopen the hole.

`import-account-with-file.html` is out: no call site passes it a `requestId` and no reader looks for one. Only `sign` and `signTypedData` write payloads, and both open their approval window — and their Ledger permission window — on `signature-request.html` (`sign-transaction/index.tsx:314`, `sign-message/index.tsx:187`). I checked that before narrowing, since getting it wrong would purge a signature being confirmed on a device.

On the unresolved link: agreed it does not matter. The change deliberately does not depend on whether a page can put a top-level tab at a WAR URL.

src/background/open-request-windows.ts
ost-ptk: Confirmed and fixed in 979e66ed. The parse now happens before the checks, so `protocol`/`host`/`pathname` are all read off the parsed URL and the `catch` is genuinely on the only fallible step. Both tests keep their names and now mean them — the `'not a url'` fixture reaches the catch instead of falling out at the prefix test one branch earlier.

The silent-skip point was the more useful half. It now logs before `continue`, through the `redactUrlQuery` helper already imported here, so a skipped tab leaves a trace instead of surfacing later as an unexplained purge.

Added a test that the log carries no URL, since that helper exists precisely because a `signMessage` approval URL has the user's plaintext message in its query string.

`open-request-windows.ts` is at 100% on all four metrics measured explicitly (`--collectCoverageFrom`), though as you note the repo gate does not reach this file.

src/background/redux/vault/reducer.ts
ost-ptk: This was the sharpest comment on the PR and I took it as written: the ceiling should not depend on the saga succeeding. Fixed in 1a4ac63c with a `mergePayloadMaps` that bounds the union unconditionally, so none of the saga's three early returns can leave the map over the cap.

Every in-memory entry survives — each arrived in this worker session, so none can be a payload stranded before a restart — and the cipher fills the remaining slots **newest-first**.

Newest-first is the reverse of `storePayload`, and deliberately so. There, "oldest" ranges over a single session in which every entry is a live request seconds old, so the oldest is the one the user has waited on longest — the Ledger case the refusal exists to protect. Across a lock, "oldest" ranges over entries that have outlived their request lifecycle, so age measures unlocks survived unanswered rather than seconds waited: the oldest is a leak, and the newest is the one written just before the lock, which is the only one an open approval window can still be waiting on. Same word, opposite population.

It also restores the in-session recovery this PR had removed with the replace-if-empty rule. Keeping the oldest instead would let a pinned set of ten leaks win every merge forever, and the bound would buy nothing.

Worst case under the rule, stated rather than hidden: a cipher whose newest entry is a genuinely live pre-lock request, plus ten more requests arriving while locked, drops that entry. Reaching it needs a burst outrunning every `windowRequestResponded`, and such a burst has almost certainly already superseded the pre-lock request.

Five of the seven new tests were red before the change (20 and 14 keys instead of 10, `cipher-0` retained, wrong survivor order). Also corrected the comment that told the reader the over-capacity state lasts until the saga runs.

src/background/handlers/sdk-methods.ts
ost-ptk: Took your second route, in 6b9e6325, because the first does not work today.

`signError` and `signTypedDataError` are built with `createSdkAction`, not `createSdkErrorAction` (`sdk-method.ts:51`, `:70` vs the factories at `:14-23`), so they carry no `error: true` — and `fetchFromBackground`'s `waitForResponse` rejects only on `'error' in message && message.error` (`sdk.ts:111`). Switching envelopes today would **resolve** the dapp promise with an `Error` instance as its payload, leaving `cancelled` and `signatureHex` both undefined. That is worse than the current shape, and it is the WALLET-1346 error-masking set rather than this PR's to change.

So the envelope stays and gains a marker: `SdkErrorCode.tooManyPendingRequests`, the first producer of the `errorCode` field `SignTypedDataResult` already declared (it had zero producers and zero consumers). `SignatureResponse`'s cancelled branch gained the same optional field so a TypeScript dapp can read it after narrowing on `cancelled` without a cast — pinned by a compile-level test, verified by deleting the field and watching the suite fail with TS2339 on the exact line a dapp would write. Additive only, so not a breaking wire change.

`sign()`'s `@returns` now states that `cancelled: true` with an `errorCode` is a wallet-side refusal rather than a user decision; it previously said only "`cancelled === true` when rejected", which this path made misleading. The pre-existing "already sign by this account" refusal deliberately carries no code, and a test holds it that way so the two stay distinguishable.

On the extension side being silent — that half mattered as much. It now logs `{ requestId, method }` via `console.error`. I chose that over a `sagaError` banner because this path is dapp-triggerable and the banner renders untranslated over every app surface, so a page could spam it; `sdk-response-to-tab.ts` made the same call for the same class of event. A user-visible surface for it needs copy, i18n and rate limiting, which is its own change.

Comp0te: Agreed on the envelope, and the reason holds up in the code: `signError`/`signTypedDataError` go through `createSdkAction`, not `createSdkErrorAction` (`sdk-method.ts:51`, `:70` vs `:14-23`), so they carry no `error: true`, and `waitForResponse` rejects only on `'error' in message && message.error` (`sdk.ts:112`). Switching today would resolve the dapp promise with an `Error` as its payload and `cancelled` undefined. Route 2 was this comment's own fallback and it is implemented in full — `errorCode` on both refusal paths, the enum declared, readable after narrowing without a cast on both public types, surviving the wire, stated in `@returns`, and pinned by a compile-level test.

Leaving this open rather than resolving it, for one residual — a judgment call rather than a defect in what you did. The envelope is unchanged and `errorCode` had no consumers before this push, so a dapp written against the documented `if (res.cancelled) → the user rejected it` shape still reports a wallet-side capacity refusal as a user rejection, and retry logic keyed on `cancelled` still will not retry. That is the part of the original point that survives. It is now a marked ambiguity instead of an unmarked one, and closing it properly looks like the WALLET-1346 error-masking work you pointed at rather than this PR.

src/background/redux/vault/reducer.ts
```diff
@@ -98,6 +100,45 @@ function storePayload(
...
+// (`signRequest` has no lock gate, so one lands while locked too), so none can
+// be a payload stranded before the last restart. The cipher fills the rest
+// NEWEST first — the reverse of `storePayload`, because age means the opposite
+// across a lock: not the request waited on longest, but the entry that survived
+// the most unlocks unanswered. A live pre-lock request is the newest of them.
+function mergePayloadMaps(
+  cipher: PayloadMap | undefined,
+  inMemory: PayloadMap
+): PayloadMap {
+  const entries = [
```
Comp0te: `Object.entries` is not insertion order, so the newest-first rule this block documents does not hold for every id the wallet accepts.

Integer-like keys are enumerated first, in ascending numeric order, ahead of every string key. `requestId` is dapp-controlled — the only constraint anywhere on the path is `isStorableRequestId`'s `requestId !== '__proto__'` (`windowManagement/request-map.ts:38-40`, repeated at `reducer.ts:89-91`), and the page holds the private port directly (`content/index.ts:177-179` hands `port2` to the page world, picked up at `sdk.ts:21-33`), so a dapp that keys its requests `"1"`, `"2"`, `"42"` rather than using the bundled UUID generator is doing nothing the wallet rejects.

For that dapp the ordering inverts. Cipher at the cap — nine stale UUID-keyed entries plus the live pre-lock request under `requestId: '42'` — and one request arriving while locked: `entries` enumerates as `['42', 'stale-uuid-0', …, 'memory-0']`, and `slice(1)` at `:138` drops `'42'` while keeping all nine stale ones. The comment at `:125` says "A live pre-lock request is the newest of them"; here it is evicted first, and the user who unlocks into a live approval window presses Sign and gets `REQUEST_NO_LONGER_AVAILABLE` (`sign-transaction/index.tsx:229-251`).

This is the worst case you already named in the `mergePayloadMaps` thread, reached at one locked-period request instead of ten — so the "burst outrunning every `windowRequestResponded`" precondition that makes it unlikely does not apply to this class of id. The blast radius is bounded, and worth stating: hoisting only moves that page's *own* entries forward, so a dapp cannot select another origin's payload as the victim.

Nothing pins the ordering either way. Every fixture in the `over MAX_STORED_PAYLOADS` block uses non-integer keys — `cipher-N` / `memory-N` via `mapOf` (`reducer.test.ts:217-223`) and the literal `shared` (`:308-310`).

Storing an explicit ordinal or timestamp alongside each payload and sorting on it would make the rule independent of key shape; normalising ids so integer-like keys cannot occur would too. Either way, a fixture whose cipher mixes an integer-like key written last with older non-integer keys, asserting the integer-keyed one survives, would hold it.

src/background/open-request-windows.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  // `null` is "no evidence", not "no window holds a request": the caller must
+  // not purge on it.
+  it('returns null rather than an empty set when the enumeration rejects', async () => {
+    mockWindowsGetAll.mockRejectedValue(new Error('boom'));
+
+    expect(await collectRequestIdsFromOpenWindows()).toBeNull();
+  });
+
+  it('never puts a window URL into the log when the enumeration rejects', async () => {
```
Comp0te, line 251: This assertion cannot fail. A plain `Error` has no enumerable own properties, so `JSON.stringify(consoleError.mock.calls)` here is `[["collectRequestIdsFromOpenWindows: could not enumerate windows",{}]]` — both `not.toContain('my-secret-message')` and `not.toContain('?')` hold whatever the code does.

Change `redactUrlQuery(error)` at `open-request-windows.ts:34` to `error` — the natural "keep the stack for debugging" edit — and this test stays green while `console.error` starts printing the untruncated message. That helper is the one thing keeping a user's plaintext `signMessage` text out of the log, since the approval URL carries it in its query string (`sdk-methods.ts:296-316`).

The new parse-failure test at `:99` uses the same technique and escapes only by accident: Node's `ERR_INVALID_URL` happens to carry enumerable non-standard `code` and `input`, where a plain `Error` carries none.

Asserting the argument's shape as well as its content closes it — `expect(consoleError).toHaveBeenCalledWith(expect.any(String), expect.any(String))`, the style already used at `sdk-methods.test.ts:551`.

Comp0te (commented):
A second pass over the changes since the last review, focused on the new `mergePayloadMaps` and the reordered URL parse. The bound and the pathname allowlist both do what they set out to do; the two comments below are about guarantees the surrounding code and tests state but do not yet enforce.

Comp0te (approved):

PR #1456, opened 2026-08-11
perf: cut the code each page parses on open (WALLET-1380, WALLET-1381)

Comp0te:
Startup work for the popup and the other page entries. The goal was the code each page has to parse before it can show anything: `casper-js-sdk` is now off every surface that only reads data, and what remains is shared between entries instead of duplicated into each one.

## WALLET-1380

- Replace `react-player` with a native `<video>` element.
- Share vendor chunks across the five page entries, so common code is fetched and parsed once instead of once per entry.

## WALLET-1381

- Lazy-load the popup routes behind a single Suspense boundary.
- Let webpack drop unused barrel re-exports.
- Load `lottie-web` on mount instead of at import. At 165 KB parsed it was the second largest thing in every entry, and nothing any app renders on open uses it — the unlock animation only appears after the password is submitted, and the rest are Ledger states.
- Keep `casper-js-sdk` off the data-reading surfaces. The SDK ships one prebuilt UMD bundle with no ESM build and no `sideEffects` flag, so a single value import costs ~919 KB that no bundler can shake back out, and it was on the startup graph of every page entry and of the content script injected into every dapp page. `casper-wallet-core` splits its repository factory: the eight read repositories are built by `setupDataRepositories()` from `wallet-repositories.ts`, while the two signing repositories move to `signing-repositories.ts`, which only the signature-request surfaces and the background import. The three remaining eager importers move to type-only imports (`constants.ts`), to core's SDK-free derivation (`utils.ts` — `isValidPublicKey`, with `isValidSecretKeyHash` relocated to `@libs/crypto`), and to core's single implementation of `getAccountHashFromPublicKey` (`entities/Account.ts`).
- Build the frontend replica store without the background's sagas. `createMainStoreReplica` used the same `createStore` factory as the background singleton, which installs the saga middleware and runs `rootSaga` — so every frontend page linked and ran the whole background saga graph, including `check-casper2-network-saga`, which imports `HttpHandler` and `RpcClient` and so put the SDK on the startup path of all five entries. Nothing in a replica can trigger a saga: every frontend dispatch goes through `dispatchToMainStore`, and there is no `useDispatch` anywhere in `src/`. The store is also rebuilt on each `popupStateUpdated` broadcast, so every broadcast started a fresh set of watcher tasks and dropped the previous ones. The replica now builds a plain `configureStore` with the same middleware options minus the saga middleware.

Had an action ever reached a replica, those sagas would have run in a page context: `openExportKeysWindow` opens the private-key export window once per open replica, `resetVault` calls `storage.local.clear()`, the vault sagas broadcast SDK events to every dapp tab, and the two timer sagas write the auto-lock deadlines the background resumes from.

## Dependencies

`casper-js-sdk` 5.0.12 → 5.1.0, and `casper-wallet-core` → `d0f77e6` (1.4.0), which pins the same SDK version — `check:casper-sdk-version` enforces that the two agree. 5.1.0 also drops `@scure/bip32`, `@scure/bip39`, `reflect-metadata`, `glob` and a bundled `node-fetch`/`whatwg-url` copy from the SDK's own dependency graph. The wallet is now the only consumer of `@scure/*`, so the note explaining why those are held at v1 is rewritten around what still holds them there: the SDK's `@noble/curves ^1.9.7` and `@noble/hashes ^1.8.0` pins, which the `@scure` v1 line shares.

## Measured

Chrome production build, summing every chunk a page loads on open — `splitChunks` means the entry file alone no longer represents it. Taken when the `casper-js-sdk` removal landed:

| Entry | Before | After |
| --- | --- | --- |
| popup | 2759.1 KB | 993.7 KB |
| connectToApp | 2327.1 KB | 903.9 KB |
| onboarding | 2373.6 KB | 1054.6 KB |
| contentScript | 959.1 KB | 40.0 KB |

`importAccountWithFile` and `signatureRequest` still link the SDK — the first parses secret-key files, the second signs.


ost-ptk (commented):
Reviewed the whole branch commit by commit, plus local Chrome and Firefox production builds.

No blockers. What I verified rather than took on trust:

- **The measurements reproduce.** Summing every chunk each page loads on open: popup 971.2 KB, connect-to-app 881.6 KB, onboarding 1059.2 KB, contentScript 39.9 KB — within ~2% of the table.
- **The SDK really is gone** from `popup`, `onboarding`, `connectToApp`, `contentScript` and `sdk`. Checked by grepping SDK marker symbols (`StoredContractByHash`, `state_get_balance`, `CLTypeOption`, `ModuleBytes`) across every chunk each page actually references, not by chunk name. It survives only in `importAccountWithFile`, `signatureRequest` and `background`, as documented. No async chunk pulls it back in on navigation.
- **The substitutions are equivalent.** Fuzzed 2000 random keys (mixed ed25519/secp256k1, a third upper-cased) through the pre-PR SDK path and the new core-backed `getAccountHashFromPublicKey`: zero divergences, in both the accept/reject sets and the output bytes — and the rejection messages match verbatim, so no caller's `catch` sees a different shape.
- **Both replica-store claims hold.** No `useDispatch`/`useStore`/`bindActionCreators` anywhere in `src/`; every `.dispatch(` is in `src/background/**` or a test. The per-broadcast saga churn was real and is gone. Middleware options match the old factory field for field.
- **Firefox MV2 is fine** — builds byte-for-byte identical to Chrome in size, background/contentScript/sdk stay single files, prod CSP gains nothing, and no page emits an inline script.
- **Every emitted chunk is reachable** in both builds (81/81), once worker chunks and their `Promise.all([...].map(e))` loads are accounted for.
- The merge commit is clean: `git merge-tree` reproduces its tree byte-for-byte.

Three comments below, all optional — two of them are about whether this PR's win survives the next six months rather than about the code being wrong.

webpack.config.js
```diff
@@ -207,6 +227,159 @@ const assertNonceIntegrity = (compiler, compilation) => {
...
+
+// A ConcatenatedModule has no `resource` of its own — its merged modules hang
+// off `.modules`. Missing that would blind the check in production builds,
+// which is where concatenation runs.
+const resourcesOf = module =>
+  module.modules?.length
+    ? module.modules.flatMap(resourcesOf)
+    : [module.resource].filter(Boolean);
+
+const assertSdkFreePageEntries = compilation => {
```
ost-ptk, line 343: **This guard fails open.** `resourcesOf` walks undocumented webpack internals (`ConcatenatedModule.modules`); if a future 5.x renames that, it returns `[]`, `offenders` stays empty, and the check passes forever — silently reverting exactly what this branch bought.

It is also the only one of the three new plugins with no unit test, so nothing in the repo demonstrates it can throw. I confirmed by hand that it does throw today (added `import 'casper-js-sdk';` to `src/apps/popup/index.tsx` → `build:chrome` exits 1 and names the importer), but that proof lives in my terminal, not in CI.

Cheap fix: assert the detector finds at least one `casper-js-sdk` module *somewhere* in the compilation, and throw "SDK detector found nothing — the walker is broken" otherwise. `signatureRequest` and `importAccountWithFile` still link the SDK eagerly, so a positive control is always available.

webpack.config.js
```diff
@@ -207,6 +227,159 @@ const assertNonceIntegrity = (compiler, compilation) => {
...
+// casper-js-sdk is a prebuilt UMD blob with no ESM build, so one value import
+// costs ~900 KB no bundler can shake out — parsed on every popup open
+// (WALLET-1381). Whether it stays out turns on which *names* eagerly-reached
+// modules pull from the @libs/ui/components barrel: `PasswordInputs` or
+// `NoConnectedLedger` re-link it, and nothing else would fail. Hence a tripwire
+// rather than a one-off measurement.
+//
+// Not every page entry: onboarding, importAccountWithFile and signatureRequest
+// still link it eagerly. Add one here once it is cleaned up.
+const SDK_FREE_PAGE_ENTRY_NAMES = ['popup', 'connectToApp'];
```
ost-ptk: This list is already one entry short of what the branch achieved, and the comment above it is stale: `onboarding` no longer links the SDK eagerly.

Two independent checks. Adding `'onboarding'` here and running `build:chrome` passes. And separately, none of the 16 chunks `onboarding.html` loads contains any SDK marker symbol.

Worth claiming now rather than later — onboarding is the surface where the seed phrase is entered, so it is the one you least want silently regressing.

src/background/redux/create-main-store-replica.ts
```diff
@@ -0,0 +1,54 @@
...
+ * so no watcher here could ever fire — and it is rebuilt on every `popupStateUpdated`
+ * broadcast, so each broadcast was starting a fresh set of watcher tasks.
+ *
+ * It is also what dragged `casper-js-sdk` onto every page's startup path, via
+ * `sagas/check-casper2-network-saga.ts`. And had an action ever reached a replica, the
+ * background-only sagas would have run in a page context: `openExportKeysWindow` opens the
+ * private-key export window, `resetVault` calls `storage.local.clear()`, and the vault sagas
+ * broadcast SDK events to dapp tabs — once per open replica.
+ */
+export function createMainStoreReplica<T extends PopupState>(state: T) {
```
ost-ptk, line 24: The invariant this file rests on — that nothing dispatches into a replica — is enforced by nothing: no eslint `no-restricted-imports` on `useDispatch`/`useStore`/`connect`, no runtime guard, no test. The replica's `dispatch` is fully functional.

It does hold today; I checked exhaustively (zero `useDispatch`/`useStore`/`bindActionCreators` in `src/`, every `.dispatch(` in `src/background/**` or a test). And dropping the saga middleware strictly shrinks the blast radius of getting it wrong, since the background-only sagas this file's comment enumerates can no longer run in a page.

But the next `useDispatch` someone adds produces a phantom-state heisenbug: the UI updates, the background never hears it, the next `popupStateUpdated` broadcast silently reverts it. One middleware that throws `'replica is read-only — use dispatchToMainStore'` turns that into a loud failure, makes the invariant executable instead of comment-borne, and gives this new file its first test. Fine as a follow-up if you would rather not widen this branch.

webpack.config.js
Comp0te: Done in 7b8d89fc — `onboarding` is in the list.

Right on both counts, and the comment turned out stale on a third: `PasswordInputs` no longer re-links the SDK either. This branch is what unhooked it — moving `isValidSecretKeyHash` out of `@src/utils` into `@libs/crypto/is-valid-secret-key-hash` took `form-validation-rules.ts` off the SDK, and `PasswordInputs` reaches it through that. `NoConnectedLedger` still links it via `@libs/services/ledger`, so that half stands and I kept it.

Worth recording for whoever reads the guard next: onboarding has no `React.lazy` anywhere, so its whole page tree sits in the initial chunks the check inspects — nothing is hiding behind a dynamic import there.

webpack.config.js
Comp0te: Test added in 25bbe882 — a `describe('AssertSdkFreePageEntries')` next to the existing two, ten cases against a fake compilation: the healthy build, a poisoned initial chunk per guarded entry, the two entries still exempt, importer attribution, the bare-package fallback when the issuer is concatenated, the `ConcatenatedModule` walk, and a missing entry. Mutation-checked rather than trusted green: dropping `onboarding` from the list fails 2 cases, deleting the `.modules` walk fails the concatenation case.

On fails-open the strong form doesn't hold. I simulated the rename — `module.modules` → `module.SOME_RENAMED_FIELD` in `resourcesOf` — together with your `import 'casper-js-sdk';` in `src/apps/popup/index.tsx`, and `build:chrome` still exits 1 with `entry "popup" links casper-js-sdk`. Reason: casper-js-sdk is UMD/CJS, so webpack cannot concatenate it; its own NormalModule keeps `.resource` and gets matched directly. The walk only resolves the *issuer* for the message — with it broken, detection survives and the attribution degrades to a bare `casper-js-sdk`, which is now one of the test cases.

So the walk becomes load-bearing only if casper-js-sdk ships a real ESM build **and** webpack renames the field — two conditions rather than one. The positive control is left out on that basis; the unit test covers the part that was uncontested.

src/background/redux/create-main-store-replica.ts
Comp0te: Agreed, and filed as WALLET-1422 rather than widening this branch.

Re-verified independently before ticketing: zero `useDispatch` / `useStore` / `bindActionCreators` in `src/`, every `.dispatch(` under `src/background/**` or a test.

The ticket carries the middleware plus an eslint `no-restricted-imports` rule on `useDispatch` / `useStore` / `connect` outside the background — the rule catches it before it runs, the middleware catches what the rule misses — and flags that `createMainStoreReplica` has moved out of `get-main-store.ts`, since WALLET-1385's line references point at the old location and the two tickets reshape the same function from opposite directions.

ost-ptk (approved):
Approving. Re-verified both new commits locally rather than reading them.

**7b8d89fc — onboarding in the guard.** A clean `build:chrome` on 25bbe882 passes with `onboarding` guarded, so the claim is now enforced by webpack's own module graph instead of by my marker grep. No `React.lazy` anywhere in `src/apps/onboarding`, so the initial chunks the check inspects really are the whole page tree.

One correction to the mechanism in your comment, which does not change the outcome: `form-validation-rules.ts` is not off the SDK. It still statically imports `isValidSecretKeyHash`, which value-imports `PrivateKey` and `KeyAlgorithm` from `casper-js-sdk`. What keeps it out of onboarding is the `sideEffects` list from 38704042 plus the finer module granularity — the chain is still there, and webpack drops it because nothing onboarding imports reaches that specifier. That is an argument for the tripwire rather than against it: the absence rests on tree-shaking a live import chain, and the guard is the only thing that would notice if it stopped being shaken out.

**25bbe882 — the test.** Ten cases, green. Both mutations you cite reproduce here: dropping `onboarding` from `SDK_FREE_PAGE_ENTRY_NAMES` fails exactly two (`rejects casper-js-sdk in the initial chunks of onboarding`, `rejects an entry the build did not emit at all`), and collapsing `resourcesOf` to `[module.resource].filter(Boolean)` fails only `finds it inside a ConcatenatedModule`.

**Fails-open — you are right, and I withdraw it.** Ran your simulation rather than taking it: `.modules` renamed to `.SOME_RENAMED_FIELD` in `resourcesOf`, plus `import 'casper-js-sdk';` at the top of `src/apps/popup/index.tsx`. `build:chrome` exits 1 with `entry "popup" links casper-js-sdk`. The attribution did not even degrade in that case — the issuer is the entry's own NormalModule, so the message still named `src/apps/popup/index.tsx -> casper-js-sdk`. Detection survives because the UMD SDK cannot be concatenated, exactly as you described, so the walk is message-only until the SDK ships ESM.

The replica thread is filed as WALLET-1422 and reads well — the "re-check the audit against whatever branch is current" caveat and the WALLET-1385 collision note are both the right things to carry over.

All six CI checks green; prettier and eslint clean on both changed files.

PR #1462, opened 2026-08-13
WALLET-1416 — close the windows a Ledger flow owns

ost-ptk:
## WALLET-1416 — close the windows a Ledger flow owns

https://make-software.atlassian.net/browse/WALLET-1416

> **Base note.** This work was cut from `release/2.7.0`, which was merged into `develop` and
> deleted while it was in progress, so the PR targets `develop`. `develop` has since been merged
> into the branch twice and the tree is conflict-free.

### The bug

On a dapp `sign` / `signMessage` with a Ledger account, when the device needs a WebHID permission
prompt the wallet opens a second window. The flow's cleanup asked the browser for every popup
window in the profile and removed all of them — other dapps' approval windows (each removal
reaching the cancel-on-close path and cancelling a live request), the secret-key export window,
import-account-with-file windows, and popup windows belonging to ordinary web pages. It never
closed its own permission window, because that one is created `type: 'normal'` while the filter
asked for `'popup'`.

### The model: ownership is proved, never inferred

`state.ledger.windowId` is **one global scalar** with no per-request keying
(`redux/ledger/reducer.ts:5-19`). It says a Ledger permission window exists somewhere in the
profile; it says nothing about whose. Everything below follows from refusing to read it as an
answer to "is this window mine".

A caller can always produce that answer honestly, in one of two ways: it **opened** the window, or
it **is** the window. `useLedger` derives `ownPermissionWindowId` from exactly those two facts —
a ref set at creation (`use-ledger.ts:267`) and `windows.getCurrent()` for a page rendering inside
the permission window (`use-ledger.ts:209`) — and cross-checks the result against the slot
(`use-ledger.ts:236-240`), so a window it opened and then lost stops counting once the background
clears the stale id. A takeover reads as `null` — "no permission window of mine" — never as
someone else's. The hook no longer returns the raw slot at all, so no page can branch on it.

### What is in this PR

1. **The sweep is gone** (`use-ledger.ts`), replaced by a background command that closes only the
   caller's own permission window ∪ `requests[requestId].windowIds`. That is the CTA/abandon path.
2. **The background handler verifies before it removes** (`close-ledger-flow-windows.ts`). It never
   reads `state.ledger.windowId` for targeting; it takes `permissionWindowId` from the payload
   (`:32`), subtracts every window another open request still claims (`:60-69` — the same
   subtraction the response-path sibling does, and for the same reason: a shared window is that
   request's only display), and dispatches `ledgerStateCleared()` only when the slice still names
   the id **and** that id survived the subtraction (`:74-80`), so a flow that took the slot over
   keeps the `deploy`/`transaction` it is signing.
3. **The message binds its `requestId` to the sender** (`redux-actions.ts:275`). `isTrustedUiSender`
   admits every wallet page and stops there, and this branch decides a request's lifecycle without
   consulting `method` — `connect`, `switchAccount` and `decryptMessage` were treated like a Ledger
   sign. Every legitimate dispatcher's page URL already carries the id, including the permission
   window's (`use-ledger` builds that URL from the same params), so the comparison against
   `new URL(sender.url).searchParams.get('requestId')` needs no special case.
4. **The response path closes the flow.** The earlier fix fired from a Ledger-event subscriber, and
   every subscriber in the codebase sits behind `debounceTime(300)` (`ledger.ts:44-45`), so on a
   successful sign the command arrived ~300 ms after the descriptor had already collapsed to
   `{ status: 'responded' }` and dropped `windowIds`. `close-windows-on-response.ts` snapshots the
   window set in the same synchronous block that marks the request responded, delivers the
   response, then re-checks ownership against `selectOpenRequests` and removes what is still ours.
5. **The approval window's controls stop cancelling a live signature.** 'Got it' and the header
   back-arrow shared one handler that tore down the whole flow; in the approval window that made it
   the last removal, so the dapp was answered `{cancelled:true}` 250 ms later while the user was
   confirming on the device. With a permission window of **this flow** live, both controls now
   dismiss only the approval window — `candidates` requires `windowIds.length === 1`
   (`cancel-requests.ts:178`), so the request survives and the signature still answers the dapp.
6. **The decision is a value, not a call.** `decideLedgerFlowControl` returns
   `'end-flow' | 'dismiss-this-window' | 'return-to-main'` and each page switches on it. The
   effects-object it replaced had three `() => void` members, so every permutation of its bodies
   type-checked — and the swap that binds `dismissThisWindow` to the permission window turns the
   user's only exit into a silent no-op (`closeCurrentWindow` removes a window only when
   `windows.getCurrent().type === 'popup'`, while the permission window is `type: 'normal'`) with
   tsc, the suite and CI all green.
7. **The error CTA restores page state on every path**, as it did before this work.
   `closeCurrentWindow` rejects, *and* resolves having done nothing on a non-popup window; either
   one used to leave the user on a dead error screen with no route back to the transaction details.
8. **The raw-JSON back arrow is no longer a flow control.** It rendered for `RowDataContent` too,
   so leaving that view ran the Ledger teardown — reachable with no Ledger error involved at all.
   It now sets page state and nothing else.
9. **A stale `ledger.windowId` is cleared by the background** on `windows.onRemoved`. That slot is
   a mutex; without this, (5) would trade a wrong cancel for a session-long Ledger lockout. The two
   are not separable.

### What is deliberately not in this PR

- Making the approval window _reflect_ the permission window's progress. It still renders "Please
  provide permission to connect your Ledger device" for the whole flow. Follow-up ticket.
- MV3 service-worker restart mid-flow: the request map and the ledger slice are memory-only, so
  after a restart nothing is closed and both windows stay on screen. The signature still reaches
  the dapp and nothing is cancelled. Accepted and documented; follow-up ticket.
- If the user dismisses the approval window and then simply leaves the permission window open,
  nothing answers the dapp until they act in that window. The permission window is the focused one
  throughout the device phase and every way of closing it answers the dapp, but the wait is
  unbounded under user inaction.

### What is NOT proven

- **There is no Ledger e2e coverage at all** (`grep -rli ledger e2e-tests` returns nothing), and
  the popup e2e suite runs with `MOCK_STATE=true`. Nothing end-to-end exercises this flow.
- **`src/hooks` and `src/apps` are outside the coverage gate** (`jest.config.js`
  `collectCoverageFrom` covers only `src/background/redux/**/reducer.ts`,
  `src/background/handlers/**/*.ts` and `src/background/redux/sagas/**/*.ts`). No test imports
  either signature-request page or `use-ledger.ts`, and there is still no React-hook harness in the
  tree. `decideLedgerFlowControl` has its own unit tests and, returning a value rather than calling
  effects, has no permutations left to get wrong — but the page wiring around it is verified by
  nothing beyond `tsc`.
- **`windows.getCurrent()` resolves asynchronously**, so for a few ms after mount a page that IS the
  permission window has no `ownPermissionWindowId` and its teardown control is a no-op. It logs a
  `console.warn` rather than returning silently. Nothing drives this window, and the interval is far
  shorter than any user interaction, but it is unmeasured.
- That the surviving permission window keeps signing after its sibling document is removed. That is
  hardware behaviour; QA item 11 below is the only instrument.
- That the fix works on Firefox or Safari. The permission window is reachable only through a
  transport probe those targets do not pass today, so this path is effectively Chrome-only.
- Real `windows.remove` semantics on a dead id, on any target. The unit tests prove control flow
  against `jest.fn()`s and nothing about the browser.
- **That `windows.onRemoved` fires for a window the background removed itself** (premise E6). It is
  a browser API contract, not provable from this tree. This PR makes it load-bearing: it is the only
  recovery for a stale `ledger.windowId`, so if it did not hold, the mutex would stay set and no
  later Ledger flow could open a permission window for the rest of the session. QA item 13 is the
  only instrument.
- **Whether any target browser recycles numeric window ids.** The ownership check compares ids for
  equality, so a recycled id could make a caller's proof read true for a window it does not own.
  Bounded — the caller would have to have opened, or be rendering in, a window that has since been
  destroyed and its id reissued — but unmeasured.
- Ordering conclusions were derived from control flow, not measured.

**Known, unchanged by this PR:** closing the wallet's windows does not clear a pending prompt on
the Ledger device — it is resolved only on the device or by the transport closing. That was already
true when the permission window was orphaned instead of closed.

### QA status — Chrome + real Ledger

**The hardware run below predates the ownership rework** (commits up to `6875327b`). The rework
changes which window each control acts on, so every item that involves a permission window needs to
be run again against the current head. Treat the list as unrun.

The earlier run passed: 1 (the ticket), 2 (`signMessage`), 5 (a second dapp's request survives,
verified with the HID permission revoked so a permission window was genuinely in play), 6
(export-keys window and an ordinary web popup survive — the regression check), 9 (disconnect
mid-flow), 11 and 12 (both approval-window controls), 13 (no lockout), 14 (the permission window's
control is still the teardown), 15 (no permission window, no behaviour change). Items 3, 4, 7, 8
and 10 were never run.

### QA checklist — Chrome with a real Ledger device

1. Dapp `sign` with a Ledger account, device **not yet permitted** → permission window opens →
   confirm on device. **Both windows close and the dapp receives the signature.** (The ticket.)
2. Same for `signMessage`.
3. Reject on the device, then press "Got it" **in the permission window** → both windows close and
   the dapp receives `{cancelled:true}` after ~250 ms. Repeat and press "Got it" **in the approval
   window** instead: only that window closes, the dapp is still pending, and the permission
   window's own control finishes the job.
4. Dapp sign with the device **already permitted** (no permission window) → the approval window
   closes itself; nothing else is touched.
5. **Precondition — revoke the extension's HID permission first** (`chrome://settings/content/hidDevices`),
   and confirm a permission window actually opens. WebHID permission persists once granted, so
   after any earlier item on this list the flow runs with NO permission window, the request then
   has a single window, and reusing it for a second request cancels it — correct pre-existing
   supersede behaviour, not this fix. With both windows open, raise a **second** dapp request from
   another tab (it reuses the shared approval window, which then shows the second request), then
   confirm the first signature on the device → the second dapp's approval window must survive and
   stay usable.
6. With a Ledger sign pending, open the secret-key export window and an ordinary
   `window.open`-style popup → confirm on the device → **neither may close.** (The WALLET-1416
   regression check.)
7. Internal `transfer` with Ledger via `sign-with-ledger-in-new-window` → Success → Close → only
   that window closes.
8. `import-account-from-ledger` → import → Close → only that window closes.
9. With a Ledger sign pending and the permission window open, disconnect the signing account from
   the dapp origin in the extension popup so the approval window's footer flips to the
   Connect/Cancel branch, then press Cancel. **The dapp receives `{cancelled:true}` and BOTH
   windows close** — before this work the permission window was orphaned and `ledger.windowId`
   stayed set, blocking every later Ledger flow. Then reconnect the account (it was just
   disconnected) and start a fresh Ledger sign: a permission window must still open.
10. **Observation only, no pass/fail.** With the permission window open, focus the approval window:
    no Sign button, a single 'Got it' footer button, and the text still says "Please provide
    permission…" even after permission was granted. Screenshot it for the progress-channel
    follow-up.
11. **R8 minimum — pass/fail.** Grant permission and reach the device-confirmation screen. Without
    touching the device, switch to the approval window and press 'Got it'. **Only the approval
    window closes; the permission window stays and the device prompt is still live; confirming on
    the device still delivers the signature to the dapp.** On the branch point `b9170bf8` (now
    inside `develop`) both windows close and the dapp gets `{cancelled:true}` — run it there first
    to see the difference.
12. **Same, with the header back-arrow instead of 'Got it'** — identical expectation.
13. **Lockout check.** Dismiss the approval window as in 11, then close the permission window with
    the OS control (the dapp should receive `{cancelled:true}`). **Then start a fresh dapp Ledger
    sign: a permission window must still open.**
14. **The permission window's controls are still the teardown.** With both windows open at the
    device prompt, press the back-arrow **in the permission window**: both windows close and the
    dapp receives `{cancelled:true}`.
15. **No permission window, no behaviour change.** Trigger a device error (device locked) with the
    device already permitted, press 'Got it' → the approval window returns to the transaction
    details with a working Cancel and Sign; nothing closes.
16. **New — the raw-JSON view is not a flow control.** On a Ledger dapp sign, open "Show raw JSON"
    and press the header back-arrow. **The page returns to the transaction details and the dapp
    request stays pending.** Before this rework that arrow ran the Ledger teardown whenever any
    permission window was tracked, including another flow's.

Firefox and Safari QA of this fix is close to worthless (the permission window is not reachable
there through a supported transport), but a smoke test that non-Ledger dapp flows are unaffected on
both is still worth running.

### Coverage

`close-ledger-flow-windows.test.ts` pins the removal set, the ownership refusals (`a second flow
holding the global slot`) and the subtraction of a window a second open request still claims.
`redux-actions.test.ts` pins the sender binding in three directions: a page naming a request it
does not display, a page omitting the id its own URL carries, and an absent `sender.url`.
`close-ledger-flow-windows.integration.test.ts` runs the abandon path over a real store through
`handleReduxAction` → `handleWindowRemoved` → the grace timer, and pins that the abandoned request
is answered exactly once with `{cancelled:true}` and that a second dapp's approval window survives
with its descriptor still open — the assertion the mocked unit suite cannot make.

Cited rather than duplicated: `cancel-requests-displaced-by.test.ts:67-83` and
`cancel-open-requests-on-close.test.ts:62-69` already pin the invariant items 3 and 11 depend on —
two `windowIds` means a removal detaches but never cancels. Both pass unchanged.


Comp0te (commented):
Read as the Ledger window-ownership change: the new `close-ledger-flow-windows` handler, its `redux-actions` branch, the `ledger-flow-controls` decision function and the two page call sites, against the sibling `close-windows-on-response` written alongside it.

The through-line is `state.ledger.windowId` — one global scalar slot that both the UI decision and the background teardown treat as "this flow's permission window". Where a second Ledger flow holds the slot, the branch at `ledger-flow-controls.ts:22` and the removal set at `close-ledger-flow-windows.ts:20` both act on someone else's window. The sibling handler in this same PR already carries the ownership checks the new one is missing, which makes that asymmetry the quickest thing to act on. The rest: the dismiss path lost a recovery the old code guaranteed, the new message branch doesn't bind its `requestId` to the sender, and the decisions this PR makes are covered by no check in the repo.

src/apps/signature-request/ledger-flow-controls.ts
```diff
@@ -0,0 +1,28 @@
...
+  permissionWindowId: number | null,
+  effects: LedgerFlowControlEffects
+): void {
+  if (isPermissionWindow) {
+    effects.returnToMain();
+    effects.endLedgerFlow();
+    return;
+  }
+
+  if (permissionWindowId != null) {
```
Comp0te: The value this branch tests is the global `ledger.windowId` slot, not a per-flow id. Both pages destructure it out of the same `useLedger` call (`sign-transaction/index.tsx:305-310`, `sign-message/index.tsx:179-184`), and that is `useSelector(selectLedgerNewWindowId)` at `use-ledger.ts:79`; the reducer holds one scalar with no per-request keying (`redux/ledger/reducer.ts:5-19`). So the branch can't tell "this request has a permission window open" from "some other flow's is open" — which the comment at :9-10 already says.

The cost lands on the dapp. With a foreign permission window tracked (a popup transfer, an account import), this returns at :22-25 before `effects.returnToMain()` at :27 and the approval window closes. That window was the request's only display, so `cancelRequestsDisplacedBy` hits `windowIds.length === 1` (`cancel-requests.ts:177-179`) and after `CANCEL_GRACE_MS` (:24) the dapp gets `signResponse({ cancelled: true })` (:50) — reached via `windows.onRemoved` → `handleWindowRemoved` (`background/index.ts:187-192`, `window-removed.ts:31`) → `cancel-open-requests-on-close.ts:24-29`. That's QA item 15 ("no permission window, no behaviour change") going the other way.

It's also reachable with no Ledger error involved at all: the header back-arrow renders for `RowDataContent` too (`sign-transaction/index.tsx:472-477`), so backing out of the raw-JSON view cancels the dapp request whenever any Ledger permission window is tracked. Before this PR that arrow only reset page state.

`useLedger` already holds the id it created (`use-ledger.ts:212-237`) — keeping it in a ref and returning it alongside `windowId` would give this branch a per-flow value to test.

src/background/handlers/close-ledger-flow-windows.ts
```diff
@@ -0,0 +1,63 @@
...
+ * unhandled rejection in a service worker is invisible.
+ */
+export async function handleCloseLedgerFlowWindows(
+  store: MainStore,
+  requestId: string | undefined
+): Promise<void> {
+  const state = store.getState();
+  const targets = new Set<number>();
+
+  const permissionWindowId = state.ledger.windowId;
```
Comp0te: The slot goes into `targets` unconditionally at :20-23 and :46 clears the whole slice, with no test that the named request owns that window and no subtraction of windows other open requests still claim. The sibling you wrote in this same PR does both: `isLedgerFlow` at `close-windows-on-response.ts:36-38` (enforced at :53), `claimedByOthers` at :55-62.

`ledger.windowId` can be released while its window is still open — `LedgerDisconnectedFooter`'s Connect CTA dispatches `ledgerStateCleared()` (`ledger-disconnected-footer.tsx:67,91`, inside the `!ledgerTransport` branch, so from the originating window), and `renderLedgerFooter` shows that footer for `LedgerAskPermission` (`ledger-footer.tsx:26-30`); the takeover route is what `use-ledger.ts:272-277` describes. Once a second `useLedger` takes the slot, the first flow's teardown removes the second flow's permission window mid-device-confirmation, and `ledgerStateCleared` (`redux/ledger/reducer.ts:20` → `initialState`) wipes the `deploy`/`transaction` that `sign-with-ledger-in-new-window/index.tsx:31-32,41-51` is signing.

The UI guard at `use-ledger.ts:301` (`if (!windowId) return`) doesn't help — it reads the same selector as :79, so it establishes the slot is live, never that it's yours.

The signature `(store, requestId: string | undefined)` has nowhere to put an owned window id, so this needs the type to change: add the slot to `targets` only when it appears in the named request's `windowIds` (or when no `requestId` was supplied), subtract what `selectOpenRequests` still claims, and clear the slice only when it still names a window being removed. Nothing pins the current behaviour either — the closest case, `close-ledger-flow-windows.test.ts:85-91`, has no scenario where another open request claims the tracked window.

src/background/handlers/redux-actions.ts
```diff
@@ -234,6 +236,34 @@ export async function handleReduxAction(
     return { handled: true, response: undefined };
   }
 
+  // Intercepted rather than forwarded: there is no reducer case for it, and the
+  // window set it closes is derived from `windowManagement.requests`, which no
+  // replica can see. Gated on `sender` for the same reason as the branch above —
+  // closing an approval window reaches `cancelOpenRequestsForClosedWindow`, so
+  // this decides a request's lifecycle.
+  if (closeLedgerFlowWindows.match(action)) {
+    if (!isTrustedUiSender(sender)) {
```
Comp0te, line 259: `payload.requestId` is read at :253 and passed to `handleCloseLedgerFlowWindows` at :259 with nothing binding it to the sender. `isTrustedUiSender` (`private-state.ts:34-40`) checks `sender.id === runtime.id` plus an extension-URL prefix, so it admits every wallet page and stops there. The handler then closes whatever that id names: `close-ledger-flow-windows.ts:25-33` looks it up in `windowManagement.requests` and removes `windowIds` for any `status === 'open'` descriptor without consulting `method` — `connect`, `switchAccount` and `decryptMessage` are treated the same as a Ledger sign. Removing an approval window then decides that request's lifecycle through `handleWindowRemoved` → `cancelOpenRequestsForClosedWindow` → `cancelRequestsDisplacedBy`.

To be clear about severity: nothing untrusted is in the admitted set today. Extension pages aren't web-accessible (`manifest.v3.json:61-66`, `manifest.v2.json:62` expose only `sdk.bundle.js` and `logo64.png`), and the sole dispatcher (`use-ledger.ts:303-307`) passes its own URL's `requestId`. So this is a request for the same hardening the branch above already carries rather than a reachable cross-origin cancel — `windowRequestWindowAttached` at :210-237 is gated exactly this way, and its comment at :217-218 gives the reason: the sender gate is "the layer that does not depend on that allowlist staying right".

Every legitimate dispatcher's page URL already carries the id, so comparing it against `new URL(sender.url).searchParams.get('requestId')` costs a line — `sdk-response-to-tab.ts:27-36` already recovers identity from `sender.url` that way.

src/apps/signature-request/pages/sign-transaction/index.tsx
```diff
@@ -323,10 +325,18 @@ export function SignTransactionPage() {
     }
   });
 
-  const onErrorCtaPressed = () => {
-    setSigningPageState(SigningPageState.MainContent);
-    closeNewLedgerWindowsAndClearState();
-  };
+  const onErrorCtaPressed = () =>
+    runLedgerFlowControl(isLedgerNewWindow, ledgerPermissionWindowId, {
+      returnToMain: () => setSigningPageState(SigningPageState.MainContent),
```
Comp0te: The dismiss path is now `closeCurrentWindow().catch(console.error)` and nothing else (:331-337; `ledger-flow-controls.ts:22-25` returns before `returnToMain`). Before this PR the CTA reset page state synchronously first, on every path — `onErrorCtaPressed` was `setSigningPageState(SigningPageState.MainContent); closeNewLedgerWindowsAndClearState();` — so the user left the ledger error screen whether or not the window actually closed. That guarantee is gone.

`closeCurrentWindow` can genuinely reject: `close-current-window.ts:5-20` re-throws from its own catch, so both `windows.getCurrent()` and `windows.remove()` failures propagate — presumably why the `.catch` is there at all. When it fires, the CTA produces a console line and the user is left on a dead screen with no route back to the transaction details.

There's a quieter second path to the same place: when `currentWindow.type !== 'popup'` (:13) the promise *resolves* having done nothing — no close, no state reset, not even a log.

Running `returnToMain` as the fallback restores the property the change dropped. Same shape at `sign-message/index.tsx:205-208`.

src/apps/signature-request/pages/sign-message/index.tsx
```diff
@@ -197,10 +199,15 @@ export function SignMessagePage() {
     }
   });
 
-  const onErrorCtaPressed = () => {
-    setShowLedgerConfirm(false);
-    closeNewLedgerWindowsAndClearState();
-  };
+  const onErrorCtaPressed = () =>
+    runLedgerFlowControl(isLedgerNewWindow, ledgerPermissionWindowId, {
```
Comp0te: Four decisions this PR makes that no check in the repository covers — grouped because they share one fix shape.

This wiring is the clearest instance: all three members of `LedgerFlowControlEffects` are `() => void`, so every permutation of this object literal type-checks. Swap the bodies bound to `returnToMain` and `dismissThisWindow` and the permission branch calls `closeCurrentWindow()`, which only removes a window when `windows.getCurrent().type === 'popup'` (`close-current-window.ts:13`) — while `openNewSeparateWindow` creates the permission window `type: 'normal'` (`create-open-window.ts:189,194`). The user's only exit from a failed Ledger flow becomes a silent no-op button, and tsc, the suite and CI all stay green. `ledger-flow-controls.test.ts` drives only `jest.fn()`s and never renders either page; the same wiring is at `sign-transaction/index.tsx:329`.

The other three:

- **`use-ledger.ts:303`** — the single dispatch site of `closeLedgerFlowWindows`, whose `requestId` is what makes the handler's approval-window half run. `closeLedgerFlowWindows({})` type-checks (the payload is `{ requestId?: string }`) and leaves the suite green, including `redux-actions.test.ts:215` and `close-ledger-flow-windows.test.ts:85` — which pin both shapes as valid and would then simply describe the wrong flow. At runtime the dapp approval window stops being closed, half of what the ticket asks for, with nothing failing.
- **`use-ledger.ts:300-308`** — `closeNewLedgerWindowsAndClearState` returns early on a falsy `windowId` (:301), and nothing exercises it in that state, so "a permission window whose slot was released cannot end its flow" is unpinned. There is no `use-ledger` test file in the tree at all.
- **`close-ledger-flow-windows.test.ts:151`** — the suite pins which ids are removed, `ledgerStateCleared` and console output, but never that the abandoned request still gets answered; it mocks `webextension-polyfill` down to `windows.remove` and never drives `handleWindowRemoved`. A "tidy the descriptor up" change inside the handler that tombstones it would build no cancel candidate, and the dapp's promise would never settle. `close-windows-on-response.test.ts:571-604` has exactly this integration for the response path.

Underneath the first three: `src/apps/**` and `src/hooks/**` are outside `jest.config.js` `collectCoverageFrom`, and no test imports either page or `use-ledger.ts`. The repo's own precedent is `src/hooks/register-ledger-permission-window.ts`, extracted from this same hook because "the repo has no React-hook harness" — the same move fits, or have `runLedgerFlowControl` return a `'return-to-main' | 'dismiss-this-window' | 'end-flow'` decision the page switches on, which has no permutations. For the last one, a real-store case in the style of `close-windows-on-response.test.ts` (`makeRealStore` + `openWith`).

src/apps/signature-request/ledger-flow-controls.ts
ost-ptk: Confirmed, and fixed in `c3e9640a`.

The branch now tests a per-flow value. `useLedger` derives `ownPermissionWindowId` two ways — a ref set to the id of the window **this instance opened** (`use-ledger.ts`, at the `ledgerNewWindowIdChanged` dispatch), or `windows.getCurrent()` for a page that **is** the permission window — and returns that instead of the raw slot. It is still derived from `state.ledger.windowId`, so a window this instance opened and then lost stops counting once the background clears the stale id; a takeover reads as `null` ("no permission window of mine"), not as someone else's.

The hook no longer exports `windowId` at all, so there is no raw slot left for a page to branch on.

On the second path you found — the back arrow rendering for `RowDataContent` — that one was wrong independently of ownership: leaving the raw-JSON view is a page-state move, not a flow control. The header now routes it to its own `onBackFromRawData` and only the Ledger screen's arrow reaches `onErrorCtaPressed`.

src/background/handlers/close-ledger-flow-windows.ts
ost-ptk: Confirmed — including the takeover route through `LedgerDisconnectedFooter`'s Connect CTA. Fixed in `c3e9640a`, though not quite the way you sketched it.

Your suggestion was to add the slot to `targets` only when it appears in the named request's `windowIds`, or when no `requestId` was supplied. The second half is the problem: the internal flows (`import-account-from-ledger`, `sign-with-ledger-in-new-window`) never supply one, and those are exactly the case your takeover scenario describes — so that rule would keep trusting the slot precisely where it cannot be checked.

So ownership is proved by the caller instead. `closeLedgerFlowWindows` now carries `permissionWindowId`, which the dispatcher can always produce honestly: either it opened that window, or it is that window (`windows.getCurrent()`). The handler:

- no longer reads `state.ledger.windowId` for targeting at all;
- subtracts `claimedByOthers` — the union of `selectOpenRequests` windowIds excluding the named request, the same subtraction the response-path sibling does and for the same reason;
- dispatches `ledgerStateCleared()` only when the slice still names `permissionWindowId` **and** that id survived the subtraction, so a flow that took the slot over keeps the `deploy`/`transaction` it is signing.

One consequence worth naming: the caller's window is still closed when the slot has moved on, since the caller owns it either way — an orphan gets cleaned up rather than left on screen.

Pinned by two new cases under `a second flow holding the global slot` in `close-ledger-flow-windows.test.ts`, plus `leaves a window a second open request still claims, even when this flow lists it` — the gap you noted around the old `:85-91`.

src/background/handlers/redux-actions.ts
ost-ptk: Agreed, and taken as the hardening request it is — the branch now does exactly what you costed:

```ts
if ((payload.requestId ?? null) !== recoverRequestId(sender.url)) {
  return { handled: true };
}
```

`recoverRequestId` is the same `new URL(sender.url).searchParams.get(...)` recovery `sdk-response-to-tab.ts` does for the origin. It works for every legitimate dispatcher without a special case: the approval window's URL carries `?requestId=`, and so does the permission window's, because `use-ledger` builds that URL out of the same `askPermissionUrlData.params`. The internal flows send no `requestId` and their page URL carries none, so both sides read `null` and agree.

Three cases added to `redux-actions.test.ts`: a trusted page naming a request it does not display, a trusted page omitting the id its own URL carries, and an unparseable/absent `sender.url`. All three drop.

I did not extend the same binding to `permissionWindowId` — window ids are not in the URL, so there is nothing to bind it to. It stays trusted-by-construction from a gated sender, which is the same standing as `windowId` on `windowRequestWindowAttached`.

src/apps/signature-request/pages/sign-transaction/index.tsx
ost-ptk: Right on both paths, including the quiet one — `closeCurrentWindow` resolving without doing anything when `currentWindow.type !== 'popup'` is the worse of the two, since it does not even leave a console line.

Fixed by restoring the property rather than adding a fallback to one branch: `onErrorCtaPressed` now resets the page state **unconditionally**, before it acts on the decision, which is what the pre-PR `onErrorCtaPressed` did on every path. So a rejection, a no-op resolve, or a successful close all leave the user on the transaction details if the window is still there.

```ts
const decision = decideLedgerFlowControl(isLedgerNewWindow, ownPermissionWindowId);
setSigningPageState(SigningPageState.MainContent);
if (decision === 'end-flow') { closeNewLedgerWindowsAndClearState(); return; }
if (decision === 'dismiss-this-window') { closeCurrentWindow().catch(...); }
```

Same shape at `sign-message/index.tsx` with `setShowLedgerConfirm(false)`.

src/apps/signature-request/pages/sign-message/index.tsx
ost-ptk: Took the second of your two suggestions — the decision return — since it removes the permutation space rather than testing it.

`runLedgerFlowControl(isPermissionWindow, id, effects)` is now `decideLedgerFlowControl(isPermissionWindow, ownPermissionWindowId): 'end-flow' | 'dismiss-this-window' | 'return-to-main'`, and each page switches on the result. There is no object literal left to permute, and the swap you described stops being expressible.

On the other three:

- **`use-ledger.ts:303`** — `closeLedgerFlowWindows({})` no longer type-checks its way past the problem, because the payload is now the ownership proof: the guard is `if (ownPermissionWindowId == null) return`, and the dispatch passes both fields. A missing `requestId` is still a valid shape for the internal flows, but the background now rejects a `requestId` that disagrees with `sender.url`, so "the wrong flow described" is caught at the boundary rather than only by a reader.
- **`use-ledger.ts:300-308`** — still no React-hook harness, so the falsy-id early return is still not driven by a test. What it does now is `console.warn` instead of returning silently, so the case you named — a permission window whose slot was released cannot end its flow — leaves evidence rather than a dead button. It is also reachable for a few ms on mount before `windows.getCurrent()` resolves, which the warning names explicitly.
- **`close-ledger-flow-windows.test.ts:151`** — added `close-ledger-flow-windows.integration.test.ts`, built the way you pointed at (`makeRealStore` + `openWith`, real reducers, driving `handleReduxAction` then `handleWindowRemoved` then the grace timer). It pins that the abandoned request is answered exactly once with `{cancelled: true}`, and that a second dapp's approval window survives the abandon and keeps its own descriptor open. That is the assertion the mocked unit suite could not make.

`src/apps/**` and `src/hooks/**` are still outside `collectCoverageFrom`, and neither page is rendered by any test — that part of your note stands.

Comp0te: Thanks — taking the four in turn, since they came in as one.

**`close-ledger-flow-windows.test.ts:151` — closed.** `close-ledger-flow-windows.integration.test.ts` is the assertion the mocked suite could not make, and it holds under mutation: dispatch `windowRequestResponded({ requestId })` inside `handleCloseLedgerFlowWindows` and the integration file fails at :99 with `Expected number of calls: 1, Received number of calls: 0`. The same edit against the previous head left all 23 tests in both files green. That gap is pinned.

**`ledger-flow-controls.ts` decision — mostly closed.** The permutation space is genuinely gone: `decideLedgerFlowControl` returning a union means a mistyped decision is a compile error. What is not closed is the surface the note was filed against — the page wiring. `sign-transaction/index.tsx:339-348` and `sign-message/index.tsx:213-222` are still hand-bound with nothing checking them, and neither consumes the union exhaustively: both are two `if`s with `'return-to-main'` as an implicit fall-through, so moving `closeCurrentWindow()` into the `'end-flow'` branch type-checks and ships green.

**`use-ledger.ts` dispatch — the type argument does not hold.** `closeLedgerFlowWindows({})` still compiles: `ledger/actions.ts:30-33` declares both payload fields optional, so the payload is not yet the ownership proof the type would have to make it. What that mutation now produces is different but not better — the message is dropped at `redux-actions.ts:275-277` and nothing is closed at all. The boundary failing closed is a real improvement over the previous head; the seam is still absent.

**`ledger-flow-controls.test.ts` case — half.** As a pure return-value assertion at :10-12 it no longer reads as end-to-end coverage, and the `console.warn` at :347-353 does leave evidence where the old code returned silently. But nothing in the tree drives `closeNewLedgerWindowsAndClearState`, so its reachable surface is still unchecked — and it grew: `ownPermissionWindowId` is null both when the slot was released and, for every page that *is* the permission window, until `windows.getCurrent()` resolves.

Agreed on the last line of your reply, and it is the through-line here: `src/apps/**` and `src/hooks/**` outside `collectCoverageFrom` is what leaves all three of the remaining halves unenforced.

src/hooks/use-ledger.ts
```diff
@@ -201,6 +201,45 @@ export const useLedger = ({
...
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  // Null until the slot names a window this instance owns, so a takeover between
+  // the two flows reads as "no permission window of mine" rather than as someone
+  // else's. Still derived from the slot, so a window this instance opened and
+  // then lost stops counting once the background clears the stale id.
+  const ownPermissionWindowId =
```
Comp0te: `ownPermissionWindowId` (:236-241) recognises two owners: the instance that called `openNewSeparateWindow` (`openedPermissionWindowIdRef`, assigned only at :267) and the instance rendering inside the window (`hostWindowId`, from the one-shot `windows.getCurrent()` at :211-230 — empty deps, and the `.catch` at :221-225 only logs). Both are per-document, so an instance that is neither computes `null` while the slot is still set, and `closeNewLedgerWindowsAndClearState` takes the early return at :347-353: one `console.warn`, no dispatch.

Repro: popup → Import account from Ledger → the permission window opens and takes focus, so the popup document is torn down → reopen the popup and navigate back to the Ledger screen (`navigation-menu/index.tsx:191-204` → `app-router.tsx:387-388`) → press the X in `LedgerConnectionView`. Nothing happens but the console line. That is a regression in this control: base `use-ledger.ts:285-292` gated on `if (windowId)` — the same global slot, so any instance passed — and closed the window and dispatched `ledgerStateCleared()` from the hook itself.

Two things bound it, and both are worth stating because they are why this is not a stuck wallet: the permission window's own X still works (`hostWindowId === windowId` there), and `window-removed.ts:22-25`, new in this PR, clears the slice however that window goes away — so the `!windowId` gate at :246-251 unblocks the next flow as soon as it closes. `LedgerDisconnectedFooter`'s Connect CTA on the same screen also clears it (:66-68, :90-92). The cost is a silently dead control on the popup side, not an unrecoverable state.

The fix the codebase already suggests is a third witness that survives a remount: `isLedgerNewWindow` is derived from the URL (`sign-transaction/index.tsx:93`) and `import-account-from-ledger/index.tsx:13-16` reads the same param, so an opener id carried in the URL or persisted in the slice would make the re-mounted popup a legitimate owner again.

One open question rather than a claim: can `windows.getCurrent()` reject in an extension page? If it can, `hostWindowId` stays null for that window's whole life and the comment at :348-350 ("briefly, on mount") would not hold — the X, the `end-flow` branch at `sign-transaction/index.tsx:339-341` and the auto-close at :363-372 would all be dead there, and the `.catch` at :221-225 has no retry.

Comp0te (commented):
Re-read the increment against the round-1 notes. Four of the five threads are closed by `c3e9640a` — the ownership derivation, the handler's targeting and its `claimedByOthers` subtraction, the sender binding on `requestId`, and the unconditional page-state reset all check out at the lines you named. The one new note below is the other side of that same ownership question: the two witnesses `ownPermissionWindowId` is derived from are both per-document, so an instance that is neither the opener nor the window itself reads as a non-owner.

src/hooks/use-ledger.ts
ost-ptk: Confirmed, including the repro — popup → Import account from Ledger → the popup document goes away when the permission window takes focus → reopen → the X does nothing but log. Fixed in `0da4daea`.

Both witnesses were per-document, which is the whole defect: neither of them can describe a document that no longer exists. The third one rides in the slice instead, next to the id it qualifies — `ledgerNewWindowIdChanged` now carries `{ windowId, openerWindowId }` (`redux/ledger/reducer.ts:21-25`), where `openerWindowId` is the window the opening instance was rendering in. A remounted popup re-derives that same id from its own `windows.getCurrent()`, so it is an owner again.

The decision moved out of the hook into `src/hooks/ledger-window-ownership.ts` — `resolveOwnPermissionWindowId({ slotWindowId, openerWindowId, openedWindowId, hostWindowId })`, eight cases in `ledger-window-ownership.test.ts`. That is the seam your last note asked for twice: the hook has no harness, but the rule does.

One narrowing worth naming rather than leaving to be found: the opener witness is per *window*, not per instance, so a different instance rendering in the same browser window also reads as an owner. Base gated on `if (windowId)` and so admitted every instance anywhere; this is narrower than that, not wider — but it is not "only the flow that started it" either.

On your open question — can `windows.getCurrent()` reject in an extension page. It resolves for any rendered document: no permission is involved and the call is answered by the window containing the caller. The failure mode is a caller with no containing window (a service worker with no open windows), which is not this. So the "briefly, on mount" comment holds for the popup side.

What I am not claiming is that it cannot fail. If it did, the page that IS the permission window would still be stuck, because `hostWindowId` is its only witness — it did not open itself, and its `openerWindowId` names the popup. What bounds that is the same thing that bounds the case you found: the window's own titlebar close still works, and `window-removed.ts:22-25` clears the slice however it goes away, so the next flow is not blocked. I left it there rather than adding a retry with no reachable trigger to test it against.

src/apps/signature-request/pages/sign-message/index.tsx
ost-ptk: Both halves you left open are closed in `0da4daea`, and you were right that the type argument did not hold.

**The page wiring.** Two `if`s with `'return-to-main'` as an implicit fall-through is exactly the surface the note was filed against, and the union did nothing about it. Both sites are now an exhaustive `switch` with an `assertNever` default (`sign-transaction/index.tsx:342-358`, `sign-message/index.tsx:216-229`, `ledger-flow-controls.ts:26-30`). Moving `closeCurrentWindow()` into the `'end-flow'` branch is still expressible — nothing types away a wrong body — but a new member of the union is now a compile error at each call site instead of silently taking the fall-through, which is the half that was actually unguarded.

**`closeLedgerFlowWindows({})`.** Agreed, and fixed at the declaration rather than around it: `permissionWindowId` is required in the payload type (`ledger/actions.ts`), required in `CloseLedgerFlowWindowsTarget`, and the background drops a message that lacks one with a warning (`redux-actions.ts:281-286`) — the id is the ownership proof, so a payload without it names nothing the handler may close. `{}` is now a compile error, and the runtime guard is the layer that does not depend on the type staying right. The old "payload-less is routed with undefined" test now pins the drop, plus a case for a payload naming a request but no window.

**`closeNewLedgerWindowsAndClearState`.** This is the one your other note turned out to be about, and it is answered there: the ownership rule is now `resolveOwnPermissionWindowId` in `src/hooks/ledger-window-ownership.ts` with its own test file, so the reachable surface you named — including the case where `ownPermissionWindowId` is null while the slot is set — is pinned as a rule even though the hook still is not rendered by anything.

**The integration test.** Thanks for running the mutation; I read that as closed too.

And yes on the through-line: `src/apps/**` and `src/hooks/**` are still outside `collectCoverageFrom`, and neither page is rendered by any test. Extracting the rule is a way around that for one decision at a time, not a fix for it.

Comp0te (approved):

PR #1464, opened 2026-08-13
fix: a dropped dispatch or a failed window open is no longer a silent no-op

ost-ptk:
Closes [WALLET-1390](https://make-software.atlassian.net/browse/WALLET-1390).

Completes WALLET-1364 P1 #2/#3. That work made the _in-saga_ failure loud; two adjacent halves of the same journey stayed silent, so "the user clicks and nothing happens" was still reachable after it shipped.

## Why this needed a new surface rather than a `sagaError`

A `sagaError` cannot report a transport failure: it lives in the background store, and the background store is precisely what is unreachable when `runtime.sendMessage` rejects — the dispatch that would report the error is the dispatch that failed.

The replica cannot hold one either. `createMainStoreReplica(state)` runs in the _render body_ of every app's `Tree` (`popup/index.tsx:50` and four siblings), so a new store is built on every broadcast and any locally dispatched action is discarded on the next one.

So the errors live in a small module store, `ui-error-channel.ts`, which the already-mounted `SagaErrorBanner` reads through `useSyncExternalStore`. One surface, two inputs — not a second banner. It carries a `kind` rather than a message, because its callers are plain functions with no access to `t()`; the component owns the copy.

## Transport half

`dispatchToMainStore` now resolves `true` / `false` and still **never rejects** — ~105 call sites do not catch, so rejecting would produce an unhandled rejection at each of them. The log stays universal (type + cause, never the action); the _banner_ is filtered by an allow-list.

`SURFACED_DISPATCH_ACTIONS` has one written-down criterion: the dispatch IS the sole source of the visible result. Seven actions qualify — the export-keys window (the case from the ticket), `lockVault` (the wallet stays unlocked, a security outcome rather than an inconvenience), the Ledger window attach, `resetVault`, `accountsImported` / `accountImported`, and `dismissSagaError` (without it, pressing × on a background row while the transport is down does nothing and explains nothing). A banner on all ~105 sites — theme changes and the like — would train the user to ignore the banner.

The list is built from the creators' `.type` (a rename is a compile error) and pinned by a test, so growing or shrinking it is a deliberate edit. It adds no import cycle — `redux/utils.ts` already imports `redux-action.ts`, which pulls every slice's actions — verified with `dependency:circular` (140 before, 140 after) and `import/no-cycle`.

## Nine call sites that reported failure as success

Chaining `.then` was only half of it; the paths that dispatch and continue without chaining had the same bug.

| Site                              | On a dropped dispatch, before                                                                                                                       |
| --------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| `reset-vault`                     | `closeWindowByReloadExtension()` reloaded the extension, presenting an unperformed vault reset as complete and taking the banner down with the page |
| `onboarding/welcome`              | advanced to create-password on a vault that was never reset                                                                                         |
| `connected-ledger`                | navigated Home with nothing imported                                                                                                                |
| `import-account-with-file-upload` | navigated to its success screen for an import that never reached the vault                                                                          |
| `import-account-from-torus`       | set its success step, same                                                                                                                          |
| `onboarding/create-secret-phrase` | back button navigated to create-password with the generated phrase still alive — the user would set a password over a phrase they never recorded    |
| `layout/error/window-page`        | tore the window down, then dropped the user into onboarding believing the vault was reset                                                           |

All now branch on the verdict. `onboarding/reset-wallet` needs no guard: it dispatches and navigates nowhere, so the banner is the whole fix there.

The error page also loses its `try/catch`: all three calls inside it were unawaited, so it never caught anything.

## UI half

The two identical `.catch(e => console.error(e))` handlers around `openWindow` are gone. Reporting moved _into_ `useWindowManager`, so the returned `openWindow` never rejects and a third call site cannot reintroduce the swallow. It logs `error.name`, not the rejection: the URL `windows.create` is given embeds `OpenWindowProps.searchParams`, where a sign-message plaintext can ride on other window paths, and semgrep's `cw-logging-secrets` pattern would not catch that.

## The copy says only what is true

One string serves five unrelated actions, so it claims neither that nothing changed nor that retrying helps. Both would be false somewhere: `handleReduxAction` dispatches `resetVault` into the store and only then awaits `enableOnboardingFlow`, so a rejection can arrive with the vault already wiped; and `use-ledger` sets `triggeredRef.current` immediately after its dispatch, so that effect cannot re-run. It reads: _The wallet didn't respond. Your last action may not have been applied._

Each kind renders a literal `<Trans>` rather than a `t()` lookup keyed on `kind` — a dynamic key is invisible to i18next-parser, so the strings would never reach a catalog. No extraction run here (`lang/` is far behind); they ship with their English fallback.

## Two import changes, and why they are here

`SagaErrorBanner` takes its layout primitives from `@libs/layout/containers` and its `SvgIcon` / `Typography` directly, and `typography.tsx` takes `Skeleton` from its sibling instead of from the `@libs/ui/components` barrel — a barrel importing itself back, which is what `containers.ts` warns about in its own header comment ("cause huge problems with webpack bundle and lead to blank popups").

That was also what made the component untestable: the barrels transitively load `webextension-polyfill`, which throws on import outside an extension, and `mac-scrollbar`, which ships only an ESM entry with no `main` that jest cannot resolve. It is why no barrel-importing component in this repo has had a test. Breaking the chain was cheaper and more useful than three mocks plus a `jest.config.js` change — and since it touches a component used app-wide, it is verified against a production `build:chrome`.

## Tests

- `ui-error-channel` — dedupe by key, snapshot-reference stability (a fresh array each call would re-render forever), no allocation on a deduped repeat, subscribe/unsubscribe, re-report after dismiss, server snapshot.
- `SURFACED_DISPATCH_ACTIONS` — pin, checked by mutation: dropping an entry fails the test.
- `dispatchToMainStore` — `true` on success, `false` instead of a rejection, log shape, allow-listed action surfaces, non-allow-listed action logs only.
- `SagaErrorBanner` — rendered with `renderToStaticMarkup` for each channel separately and both at once, asserting the action type never reaches the markup.

`ci-check` green: 87 suites, 763 tests, knip clean.

## The dispatch→banner seam is covered by e2e

`e2e-tests/popup/error-surface/dropped-dispatch.spec.ts` drives the real extension: it breaks the page's transport to the background, clicks the menu item, and asserts the banner. Four cases — banner on a dropped export-keys dispatch (and that no window opens), no banner for a non-surfaced action, dismiss-then-reappear, and the failed window open. 4/4 green in ~13s.

Three things that had to be established by running it, and are recorded in the file:

- **The break has to be a rejected promise, not a synchronous throw.** Current Chrome exposes `browser` natively, so `webextension-polyfill` re-exports it instead of wrapping `chrome`, and the app calls the patched function directly. A `throw` escapes synchronously out of `dispatchToMainStore` and trips the ErrorBoundary — the popup renders "Something went wrong" and no banner is involved at all.
- **Stopping the service worker does not reproduce this.** In MV3 an incoming message is itself the event that cold-starts a stopped worker, and while the vault is unlocked the keep-alive alarm wakes it within 30s regardless.
- **`unlockVault` returns before the unlock finishes** (it waits only for network idle; the flow completes through further dispatches from the page), so the transport must not be broken until the unlocked UI is up.

## One finding worth a decision, not fixed here

While a banner row is visible it **covers the header**, including the menu button. Measured: the banner is `position: fixed; top: 0` at a tooltip z-index, occupying 1280×77 over a menu icon at y=24..48, and `elementFromPoint` at the icon's centre returns the banner. So the user cannot reopen the menu to retry until they dismiss the row — the × is right there, so it is one extra click rather than a dead end.

This is pre-existing `SagaErrorBanner` layout, not something this PR introduces, but this PR is what makes the banner routine rather than rare. Fixing it means either pushing page content down by the banner height or moving the banner out of the header's strip — a visual change worth a deliberate call rather than a drive-by. It is also why the dedupe case lives in the unit test instead of e2e: a second menu click cannot land while the row is up.

## Deliberately left out

- **`loginRetryLockoutTimeSet`** (`use-lock-wallet-when-no-more-retries.ts:26`) is the dispatch that actually enforces the brute-force lockout, and it is not surfaced. A banner would tell the truth and change nothing: the effect's deps cannot change, so there is no retry path, and the real fix is not letting the lockout depend on a droppable dispatch. Worth its own ticket.
- **Layering**: `redux/utils.ts` now imports from `@libs/ui/...`, so the channel would ride into the service-worker bundle. Not reachable today — no background module calls `dispatchToMainStore` — but it inverts the background→UI direction. Fixing it properly means injecting the reporter, which is more machinery than the problem currently justifies.
- No change to the `app-events` slice or `SagaErrorSource`: local errors deliberately do not survive a reload.

[WALLET-1390]: https://make-software.atlassian.net/browse/WALLET-1390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
+++ b/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
@@ -94,11 +94,18 @@ popup.describe('Popup UI: import account with file', () => {
         .getByPlaceholder('Account name', { exact: true })
         .fill(ACCOUNT_NAMES.importedPemAccountName);
 
-      // Only the guarded dispatch: `checkAccountNameIsTaken` rides the same
-      // transport from submit-time validation, so breaking all of it stops
-      // `handleSubmit` before `onSubmit` runs and the page never gets as far as
-      // the action under test. The literal is pinned in
-      // `surfaced-dispatch-actions.test.ts`.
+      // Only the guarded dispatch, which narrows what this case covers to the
+      // late-failure window: the transport drops after the name last validated,
+      // so Import is pressable and the guard is what stops the success screen.
+      //
+      // A whole-transport break is a different, still-silent story that no
+      // guard here can reach — `checkAccountNameIsTaken` is a bare
+      // `runtime.sendMessage` inside the yup resolver, so it rejects
+      // `handleSubmit` before `onSubmit` runs, and `checkSecretKeyExist`
+      // escapes the file reader as raw untranslated text. Both are legacy sends
+      // outside `dispatchToMainStore`; fixing them is its own change.
+      //
+      // The literal is pinned in `surfaced-dispatch-actions.test.ts`.
       await breakTransport(importAccountPage, ['vault/accountImported']);
 
       await importAccountPage.getByRole('button', { name: 'Import' }).click();
--- a/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
+++ b/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
@@ -5,6 +5,7 @@ import {
   secretKeyPathForCER,
   secretKeyPathForPEM
 } from '../../constants';
+import { DISPATCH_FAILED, breakTransport } from '../../error-surface';
 import { popup, popupExpect } from '../../fixtures';
 
 popup.describe('Popup UI: import account with file', () => {
@@ -62,6 +63,54 @@ popup.describe('Popup UI: import account with file', () => {
       ).toBeVisible();
     }
   );
+  popup(
+    'should not claim success when the import dispatch is dropped',
+    async ({ unlockVault, context, popupPage }) => {
+      await unlockVault();
+
+      await popupPage.getByTestId('menu-open-icon').click();
+
+      const [importAccountPage] = await Promise.all([
+        context.waitForEvent('page'),
+        popupPage.getByText('Import account').click()
+      ]);
+
+      const fileChooserPromise = importAccountPage.waitForEvent('filechooser');
+
+      await popupExpect(
+        importAccountPage.getByRole('heading', {
+          name: 'Import account from secret key file'
+        })
+      ).toBeVisible();
+
+      await importAccountPage
+        .getByRole('button', { name: 'Upload your file' })
+        .click();
+
+      const fileChooser = await fileChooserPromise;
+      await fileChooser.setFiles(secretKeyPathForPEM);
+
+      await importAccountPage
+        .getByPlaceholder('Account name', { exact: true })
+        .fill(ACCOUNT_NAMES.importedPemAccountName);
+
+      // Broken only now: the page dispatches while it loads, and breaking it
+      // earlier would fail the flow before the button under test is reached.
+      await breakTransport(importAccountPage);
+
+      await importAccountPage.getByRole('button', { name: 'Import' }).click();
+
+      await popupExpect(
+        importAccountPage.getByText(DISPATCH_FAILED)
+      ).toBeVisible();
+      await popupExpect(
+        importAccountPage.getByRole('heading', {
+          name: 'Your account was successfully imported'
+        })
+      ).toHaveCount(0);
+    }
+  );
+
   popup(
     'should import account via .cer file',
     async ({ unlockVault, context, popupPage }) => {
--- a/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
+++ b/e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
@@ -94,9 +94,12 @@ popup.describe('Popup UI: import account with file', () => {
         .getByPlaceholder('Account name', { exact: true })
         .fill(ACCOUNT_NAMES.importedPemAccountName);
 
-      // Broken only now: the page dispatches while it loads, and breaking it
-      // earlier would fail the flow before the button under test is reached.
-      await breakTransport(importAccountPage);
+      // Only the guarded dispatch: `checkAccountNameIsTaken` rides the same
+      // transport from submit-time validation, so breaking all of it stops
+      // `handleSubmit` before `onSubmit` runs and the page never gets as far as
+      // the action under test. The literal is pinned in
+      // `surfaced-dispatch-actions.test.ts`.
+      await breakTransport(importAccountPage, ['vault/accountImported']);
 
       await importAccountPage.getByRole('button', { name: 'Import' }).click();
 
--- a/src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
+++ b/src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
@@ -36,6 +36,7 @@ export function ConfirmSecretPhrasePage({
 
   const [isConfirmationSuccess, setIsConfirmationSuccess] = useState(false);
   const [isFormValid, setIsFormValid] = useState(false);
+  const [isSubmitting, setIsSubmitting] = useState(false);
 
   if (phrase == null) {
     return <Navigate to={RouterPath.Any} />;
@@ -47,10 +48,19 @@ export function ConfirmSecretPhrasePage({
         // Only navigate once the vault actually exists: keys and encryption key
         // are already in place here, so the success page renders regardless of
         // whether this dispatch arrived.
+        //
+        // `isSubmitting` closes the window that awaiting the verdict opens:
+        // nothing else disables the button, `initVaultSaga` runs to completion
+        // inside one delivery, and `accountAdded` appends without a dedupe — so
+        // a second click would leave the new wallet with two identical
+        // "Account 1" entries on one key pair.
+        setIsSubmitting(true);
         dispatchToMainStore(initVault({ secretPhrase: phrase })).then(
           dispatched => {
             if (dispatched) {
               navigate(RouterPath.ConfirmSecretPhraseSuccess);
+            } else {
+              setIsSubmitting(false);
             }
           }
         );
@@ -93,7 +103,10 @@ export function ConfirmSecretPhrasePage({
       )}
       renderFooter={() => (
         <TabFooterContainer style={{ marginTop: '28px' }}>
-          <Button disabled={!isFormValid} onClick={handleSubmit}>
+          <Button
+            disabled={!isFormValid || isSubmitting}
+            onClick={handleSubmit}
+          >
             <Trans t={t}>Confirm</Trans>
           </Button>
         </TabFooterContainer>
--- a/src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
+++ b/src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
@@ -44,8 +44,16 @@ export function ConfirmSecretPhrasePage({
   function handleSubmit() {
     try {
       if (isConfirmationSuccess && phrase) {
-        dispatchToMainStore(initVault({ secretPhrase: phrase }));
-        navigate(RouterPath.ConfirmSecretPhraseSuccess);
+        // Only navigate once the vault actually exists: keys and encryption key
+        // are already in place here, so the success page renders regardless of
+        // whether this dispatch arrived.
+        dispatchToMainStore(initVault({ secretPhrase: phrase })).then(
+          dispatched => {
+            if (dispatched) {
+              navigate(RouterPath.ConfirmSecretPhraseSuccess);
+            }
+          }
+        );
       } else {
         throw Error(ErrorMessages.secretPhrase.INVALID_SECRET_PHRASE.message);
       }
--- a/src/background/redux/surfaced-dispatch-actions.test.ts
+++ b/src/background/redux/surfaced-dispatch-actions.test.ts
@@ -1,0 +1,26 @@
+import { ledgerNewWindowIdChanged } from './ledger/actions';
+import { lockVault, openExportKeysWindow, resetVault } from './sagas/actions';
+import { SURFACED_DISPATCH_ACTIONS } from './surfaced-dispatch-actions';
+import { accountsImported } from './vault/actions';
+
+describe('SURFACED_DISPATCH_ACTIONS', () => {
+  // A pin, not a tautology: the list decides which dropped dispatches the user
+  // is told about, and the criterion for adding one is a judgement call. Growing
+  // or shrinking it has to be a deliberate edit, not a side effect.
+  it('contains exactly the five actions whose dropped dispatch is otherwise invisible', () => {
+    expect([...SURFACED_DISPATCH_ACTIONS].sort()).toEqual(
+      [
+        openExportKeysWindow.type,
+        lockVault.type,
+        ledgerNewWindowIdChanged.type,
+        resetVault.type,
+        accountsImported.type
+      ].sort()
+    );
+  });
+
+  it('is keyed on the creators, so a rename cannot silently drop an entry', () => {
+    expect(SURFACED_DISPATCH_ACTIONS.has(openExportKeysWindow.type)).toBe(true);
+    expect(SURFACED_DISPATCH_ACTIONS.has(lockVault.type)).toBe(true);
+  });
+});
--- a/src/background/redux/surfaced-dispatch-actions.test.ts
+++ b/src/background/redux/surfaced-dispatch-actions.test.ts
@@ -5,6 +5,7 @@ import {
   initVault,
   lockVault,
   openExportKeysWindow,
+  recoverVault,
   resetVault
 } from './sagas/actions';
 import { SURFACED_DISPATCH_ACTIONS } from './surfaced-dispatch-actions';
@@ -25,6 +26,7 @@ describe('SURFACED_DISPATCH_ACTIONS', () => {
         resetVault.type,
         initKeys.type,
         initVault.type,
+        recoverVault.type,
         accountsImported.type,
         accountImported.type,
         dismissSagaError.type
@@ -45,6 +47,7 @@ describe('SURFACED_DISPATCH_ACTIONS', () => {
         'RESET_VAULT_SAGA',
         'INIT_KEYS_SAGA',
         'INIT_VAULT_SAGA',
+        'RECOVER_VAULT_SAGA',
         'vault/accountsImported',
         'vault/accountImported',
         'appEvents/dismissSagaError'
--- a/src/background/redux/surfaced-dispatch-actions.test.ts
+++ b/src/background/redux/surfaced-dispatch-actions.test.ts
@@ -1,20 +1,23 @@
+import { dismissSagaError } from './app-events/actions';
 import { ledgerNewWindowIdChanged } from './ledger/actions';
 import { lockVault, openExportKeysWindow, resetVault } from './sagas/actions';
 import { SURFACED_DISPATCH_ACTIONS } from './surfaced-dispatch-actions';
-import { accountsImported } from './vault/actions';
+import { accountImported, accountsImported } from './vault/actions';
 
 describe('SURFACED_DISPATCH_ACTIONS', () => {
   // A pin, not a tautology: the list decides which dropped dispatches the user
   // is told about, and the criterion for adding one is a judgement call. Growing
   // or shrinking it has to be a deliberate edit, not a side effect.
-  it('contains exactly the five actions whose dropped dispatch is otherwise invisible', () => {
+  it('contains exactly the seven actions whose dropped dispatch is otherwise invisible', () => {
     expect([...SURFACED_DISPATCH_ACTIONS].sort()).toEqual(
       [
         openExportKeysWindow.type,
         lockVault.type,
         ledgerNewWindowIdChanged.type,
         resetVault.type,
-        accountsImported.type
+        accountsImported.type,
+        accountImported.type,
+        dismissSagaError.type
       ].sort()
     );
   });
--- a/src/background/redux/surfaced-dispatch-actions.test.ts
+++ b/src/background/redux/surfaced-dispatch-actions.test.ts
@@ -1,20 +1,30 @@
 import { dismissSagaError } from './app-events/actions';
 import { ledgerNewWindowIdChanged } from './ledger/actions';
-import { lockVault, openExportKeysWindow, resetVault } from './sagas/actions';
+import {
+  initKeys,
+  initVault,
+  lockVault,
+  openExportKeysWindow,
+  resetVault
+} from './sagas/actions';
 import { SURFACED_DISPATCH_ACTIONS } from './surfaced-dispatch-actions';
 import { accountImported, accountsImported } from './vault/actions';
+import { windowRequestWindowAttached } from './windowManagement/actions';
 
 describe('SURFACED_DISPATCH_ACTIONS', () => {
   // A pin, not a tautology: the list decides which dropped dispatches the user
   // is told about, and the criterion for adding one is a judgement call. Growing
   // or shrinking it has to be a deliberate edit, not a side effect.
-  it('contains exactly the seven actions whose dropped dispatch is otherwise invisible', () => {
+  it('contains exactly the actions whose dropped dispatch is otherwise invisible', () => {
     expect([...SURFACED_DISPATCH_ACTIONS].sort()).toEqual(
       [
         openExportKeysWindow.type,
         lockVault.type,
+        windowRequestWindowAttached.type,
         ledgerNewWindowIdChanged.type,
         resetVault.type,
+        initKeys.type,
+        initVault.type,
         accountsImported.type,
         accountImported.type,
         dismissSagaError.type
@@ -22,8 +32,23 @@ describe('SURFACED_DISPATCH_ACTIONS', () => {
     );
   });
 
-  it('is keyed on the creators, so a rename cannot silently drop an entry', () => {
-    expect(SURFACED_DISPATCH_ACTIONS.has(openExportKeysWindow.type)).toBe(true);
-    expect(SURFACED_DISPATCH_ACTIONS.has(lockVault.type)).toBe(true);
+  // The pin above cannot see a rename: both sides read the same creators, so a
+  // renamed slice reducer moves them together. These literals are the only thing
+  // in the file that does not come from a creator.
+  it('pins the type strings themselves, which the creator-derived pin cannot', () => {
+    expect([...SURFACED_DISPATCH_ACTIONS].sort()).toEqual(
+      [
+        'OPEN_EXPORT_KEYS_WINDOW_SAGA',
+        'LOCK_VAULT_SAGA',
+        'windowManagement/windowRequestWindowAttached',
+        'ledger/ledgerNewWindowIdChanged',
+        'RESET_VAULT_SAGA',
+        'INIT_KEYS_SAGA',
+        'INIT_VAULT_SAGA',
+        'vault/accountsImported',
+        'vault/accountImported',
+        'appEvents/dismissSagaError'
+      ].sort()
+    );
   });
 });
--- a/src/background/redux/surfaced-dispatch-actions.ts
+++ b/src/background/redux/surfaced-dispatch-actions.ts
@@ -1,7 +1,14 @@
 import { dismissSagaError } from './app-events/actions';
 import { ledgerNewWindowIdChanged } from './ledger/actions';
-import { lockVault, openExportKeysWindow, resetVault } from './sagas/actions';
+import {
+  initKeys,
+  initVault,
+  lockVault,
+  openExportKeysWindow,
+  resetVault
+} from './sagas/actions';
 import { accountImported, accountsImported } from './vault/actions';
+import { windowRequestWindowAttached } from './windowManagement/actions';
 
 // Which dropped dispatches the user is told about.
 //
@@ -19,11 +26,21 @@ export const SURFACED_DISPATCH_ACTIONS: ReadonlySet<string> = new Set([
   // The user pressed Lock and the wallet stayed unlocked — a security outcome,
   // not an inconvenience.
   lockVault.type,
-  // The Ledger window never attaches to the request, which then hangs until the
-  // dapp's own timeout (see attach-window-to-request.ts).
+  // The Ledger permission window never attaches to the request, which then hangs
+  // until the dapp's own timeout (see attach-window-to-request.ts).
+  windowRequestWindowAttached.type,
+  // Its sibling, and a separate send: without it `windowId` stays null, the
+  // close tracker detaches and the permission window is never closed when the
+  // signature completes. `triggeredRef` is set regardless of either outcome, so
+  // neither has a retry path.
   ledgerNewWindowIdChanged.type,
   // Without this the reload presents an unperformed reset as done.
   resetVault.type,
+  // Onboarding's two writes. Without `initKeys` the password screen keeps
+  // rendering and nothing happens; without `initVault` the success screen
+  // renders over a vault that was never created.
+  initKeys.type,
+  initVault.type,
   // Without these the navigation presents an unperformed import as done —
   // `accountsImported` from the Ledger flow, `accountImported` from the
   // secret-key-file and Torus flows.
--- a/src/background/redux/surfaced-dispatch-actions.ts
+++ b/src/background/redux/surfaced-dispatch-actions.ts
@@ -5,6 +5,7 @@ import {
   initVault,
   lockVault,
   openExportKeysWindow,
+  recoverVault,
   resetVault
 } from './sagas/actions';
 import { accountImported, accountsImported } from './vault/actions';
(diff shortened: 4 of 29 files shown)


src/background/redux/surfaced-dispatch-actions.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// Built from the creators' `.type` so a rename is a compile error, and pinned by
+// `surfaced-dispatch-actions.test.ts` so a change here is deliberate.
+export const SURFACED_DISPATCH_ACTIONS: ReadonlySet<string> = new Set([
+  // The menu closes and no export window ever appears.
+  openExportKeysWindow.type,
+  // The user pressed Lock and the wallet stayed unlocked — a security outcome,
+  // not an inconvenience.
+  lockVault.type,
+  // The Ledger window never attaches to the request, which then hangs until the
```
Comp0te: `ledgerNewWindowIdChanged` is in the set justified by "The Ledger window never attaches to the request, which then hangs until the dapp's own timeout" — but that action only writes `windowId` into the ledger UI slice (`src/background/redux/ledger/reducer.ts:16-19`), and nothing else in the tree reads it to attach anything. The action that performs the attach is `windowRequestWindowAttached`, dispatched two lines later at `src/hooks/register-ledger-permission-window.ts:42` (from `use-ledger.ts:230`), and it isn't in the set — so the hang this comment describes is still entirely silent.

The two are independent `runtime.sendMessage` calls, and `triggeredRef` is set at `use-ledger.ts:236` regardless of either outcome, so there's no retry path if only the second is dropped.

Worth keeping the entry, though — dropping `ledgerNewWindowIdChanged` alone does cause a real failure, just not this one: `windowId` stays null, the close tracker detaches, and the permission window is never closed on signature completion (`use-ledger.ts:279-311`). So the fix is to add `windowRequestWindowAttached.type` and retarget the comment, rather than to swap them. Both mean updating the exhaustive pin at `surfaced-dispatch-actions.test.ts:12` and its "exactly the seven actions" name.

src/background/redux/utils.ts
```diff
@@ -3,25 +3,42 @@ import { runtime } from 'webextension-polyfill';
...
-    // indistinguishable from the next. Log the type and the error, NEVER the
-    // action: payloads carry vault and signature material.
-    // nosemgrep: cw-logging-secrets
-    console.error('Dispatch to Main Store failed: ' + action.type, error);
-  });
+// Resolves `true` when the background acknowledged the action, `false` when the
+// send failed. NEVER rejects: ~105 call sites do not catch, so a rejection here
+// would become an unhandled rejection at each of them.
+export function dispatchToMainStore(action: ReduxAction): Promise<boolean> {
+  return runtime
```
Comp0te: The "NEVER rejects" contract at `:17-19` is enforced by the `.catch()` at `:42`, which covers a rejected promise but not a synchronous throw out of `runtime.sendMessage` itself — the call at `:20` is unwrapped, so a sync throw escapes `dispatchToMainStore` before any handler runs. That's the shape Chrome uses for `Extension context invalidated`, and this PR's own e2e file records the consequence at `e2e-tests/popup/error-surface/dropped-dispatch.spec.ts:11-16`: it "escapes synchronously out of `dispatchToMainStore` and trips the ErrorBoundary … the whole popup renders 'Something went wrong'".

The `await`ing sites get an unhandled rejection instead of `false`. Four app entries also dispatch from the render body — `popup/index.tsx:56,58`, `connect-to-app/index.tsx:55,57`, `import-account-with-file/index.tsx:55,57`, `signature-request/index.tsx:57,59` — and those sit above the `ErrorBoundary`, which is a child of the returned tree (`popup/index.tsx:74`) with nothing wrapping `root.render(<Tree />)` at `:88`. Narrower than it sounds: all four are inside `if (themeMode === undefined …)`, so it's a first-run path rather than every render, and `onboarding/index.tsx` has no render-body dispatch at all.

`return Promise.resolve().then(() => runtime.sendMessage(action)).then(...).catch(...)` routes the synchronous path into the same handler.

src/background/redux/surfaced-dispatch-actions.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+// The criterion for membership — and the only one: the dispatch IS the sole
+// source of the visible result. If it never reaches the background, the user
+// sees nothing happen and no other feedback path exists. Everything else stays
+// log-only, because a banner on all ~105 dispatch sites (theme changes, banner
+// dismissals) would train the user to ignore the banner.
+//
+// Built from the creators' `.type` so a rename is a compile error, and pinned by
+// `surfaced-dispatch-actions.test.ts` so a change here is deliberate.
+export const SURFACED_DISPATCH_ACTIONS: ReadonlySet<string> = new Set([
```
Comp0te, line 24: Separately from the Ledger entry above: two actions meet the criterion stated at `:8-12` and are absent. `initVault` is dispatched with the return value ignored at `src/apps/onboarding/pages/confirm-secret-phrase/index.tsx:47`, with `navigate(RouterPath.ConfirmSecretPhraseSuccess)` unconditional at `:48`; `initKeys` has the same shape at `create-vault-password/index.tsx:61` with `saveIsLoggedIn(true)` at `:62`. Both are in `FORWARDED_ACTION_TYPES` (`redux-actions.ts:123-124`), so a dropped send is log-only.

The router doesn't catch either. With `initKeys` dropped, `keysDoesExist` stays false and `NoVaultRoutes` keeps rendering (`onboarding/app-router.tsx:43-56`) — the user sees nothing happen on the password screen. With `initVault` dropped, `keysDoesExist && encryptionKeyDoesExist` are already true, so `ConfirmSecretPhraseSuccessPage` renders normally over a vault that was never created.

`SagaErrorBanner` is already mounted in the onboarding Tree at `onboarding/index.tsx:44`, so adding both types would surface immediately. If they're excluded deliberately, worth saying so in the comment alongside the others.

src/libs/ui/components/saga-error-banner/ui-error-channel.ts
```diff
@@ -0,0 +1,63 @@
...
+const listeners = new Set<() => void>();
+
+function emit() {
+  listeners.forEach(listener => listener());
+}
+
+// Carries a `kind`, not a message: the callers (`dispatchToMainStore`, the
+// window-manager hook) are plain functions with no access to `t()`. The banner
+// owns the copy.
+export function reportUiError(kind: UiErrorKind, detail: string) {
```
Comp0te, line 25: Nothing clears a row when the same action later succeeds. `dismissUiError` at `:36-45` is the only removal path, and its only production caller is the banner's × at `saga-error-banner.tsx:121-123`. The success paths don't touch the channel: `utils.ts:23` is a bare `.then(() => true)`, and `use-window-manager.ts:26-40` reports only from the `catch`.

The guards make this reachable by design — on failure the UI doesn't advance, so the user retries, and the success screen is exactly where the stale row is still showing. All three import success paths stay in the same document (`import-account-with-file-upload/index.tsx:33`, `import-account-from-torus/index.tsx:48`, `connected-ledger.tsx:133-134`) and the banner is mounted above the router, so the module-level row outlives the route change. At `position: fixed; top: 0` and tooltip z-index (`saga-error-banner.tsx:28-41`) that puts "The wallet didn't respond. Your last action may not have been applied." over the screen saying the import worked. The `dispatch-failed:<TYPE>` dedupe at `:28-30` means it won't even be refreshed, just left.

A `clearUiError(kind, detail)` called from `dispatchToMainStore`'s success path and the `openWindow` success path would close it.

src/libs/layout/reset-vault/index.tsx
```diff
@@ -27,8 +27,12 @@ export const ResetVaultPage = ({ popupLayout }: ResetVaultPageProps) => {
   const { t } = useTranslation();
 
   function handleResetVault() {
-    dispatchToMainStore(resetVault()).then(() => {
-      closeWindowByReloadExtension();
+    dispatchToMainStore(resetVault()).then(dispatched => {
```
Comp0te, line 30: The seven newly gated call sites have no test that reaches their failure branch. Deleting `if (dispatched)` at `:33-35` — restoring the unconditional `closeWindowByReloadExtension()` that was here at the merge base — leaves the whole suite green: `wallet.spec.ts:28-59` drives this page but asserts only that the onboarding page appears, and the new spec never touches reset. A symbol search across every `*.test.ts(x)` / `*.spec.ts(x)` returns zero hits for `ResetVaultPage`, `WelcomePage`, `CreateSecretPhrasePage`, `WindowErrorPage`, `ImportAccountWithFileUploadPage`, `ImportAccountFromTorusPage` and `ConnectedLedger`.

The regression that would let through is the extension reloading into onboarding and telling the user their wallet was reset while the vault cipher is still on disk.

The two `navigation-menu` sites are fine — they have no in-component branch, and their new behaviour is already covered at `utils.test.ts:65-74` and `dropped-dispatch.spec.ts:113-124`.

Reusing `breakTransport` from `dropped-dispatch.spec.ts` in `wallet.spec.ts` before the final "Reset wallet" click — asserting the banner is visible and the confirmation heading is still on screen — would pin it, and the equivalent on `importAccountPage` in `import-account-with-file.spec.ts`.

src/hooks/use-window-manager.ts
```diff
@@ -10,9 +15,30 @@ import { createOpenWindow } from '@background/create-open-window';
...
+    const open = createOpenWindow();
+
+    return async (props: OpenWindowProps) => {
+      try {
+        await open(props);
+      } catch (error) {
+        // Name only, not the rejection: `searchParams` is embedded in the URL
+        // `windows.create` was given, and a sign-message plaintext can ride
+        // there. `use-ledger.ts` logs `error.name` for the same reason.
+        console.error(
```
Comp0te: The `(error as Error)?.name` reduction here is pinned by nothing. There's no `use-window-manager.test.ts` (`src/hooks` holds only `register-ledger-permission-window.test.ts`, `ledger-window-close-listener.test.ts`, `use-private-state.test.ts`), a tree-wide grep for `useWindowManager` returns four hits and none is in a test, and the new e2e case at `dropped-dispatch.spec.ts:113-124` asserts `getByText(WINDOW_FAILED)` without reading the console. `create-open-window.test.ts` covers `createOpenWindow`, not this wrapper. `utils.test.ts:54-63` isn't the equivalent either — it pins that the *action* is never logged while the full `Error` is (`expect.any(Error)`), which is a weaker property than name-only.

`OpenWindowProps.searchParams` (`create-open-window.ts:60-64`) reaches `windows.create` through `getUrlByWindowApp` at `:135`/`:140`. Both current callers — `account-list.tsx:121-124` and `navigation-menu/index.tsx:175-178` — pass `{ windowApp, isNewWindow }` only, so today the redaction is defensive; nothing would flag a regression back to logging the rejection or `props` when a third caller does pass `searchParams`.

The repo's own pattern for logic trapped in a hook (`run-keys-download.ts`, `register-ledger-permission-window.ts`) applies: extract `createReportingOpenWindow(open)` and unit-test that a rejecting `open` logs a bare name string and calls `reportUiError('window-open-failed', windowApp)`.

src/background/redux/surfaced-dispatch-actions.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        ledgerNewWindowIdChanged.type,
+        resetVault.type,
+        accountsImported.type,
+        accountImported.type,
+        dismissSagaError.type
+      ].sort()
+    );
+  });
+
+  it('is keyed on the creators, so a rename cannot silently drop an entry', () => {
```
Comp0te: This doesn't exercise a rename. A rename is a compile-time concern and both tests import the same creators, so they move together — there's no rename under which this fails and `:12` passes.

Against the exhaustive pin it's also nearly redundant: if `:12` passes, the set is exactly the seven creator-derived types and both `.has()` assertions follow. The one case where it does add something is narrow — someone drops `lockVault.type` from the set *and* updates the expected literal at `:13-21` to match, which is the natural response to a failing pin; then `:12` passes while `:26-27` fails. Worth keeping for that, but it's a different property from the one the name claims.

Either rename it to what it checks, or assert against values derived independently of the creators so a rename genuinely can't drag both sides along.

Comp0te (commented):
Reviewed the error-surfacing work across the dispatch guard, the UI error channel, the surfaced-action allow-list and the new tests. The comments cluster on two things: the allow-list and its comment describe coverage that isn't quite what the set contains, and several of the new guarantees — the never-rejects contract, the banner's lifecycle, the guards themselves — have a gap or nothing holding them in place.

src/background/redux/utils.ts
ost-ptk: Fixed in 4c61bc91, with exactly the `Promise.resolve().then(() => runtime.sendMessage(action))` shape you suggested — both paths now land in the same handler.

Pinned by a new case in `utils.test.ts`: a `sendMessage` whose implementation throws synchronously resolves `false` and still reaches the banner, where before it escaped.

The e2e helper's comment also drops its claim that a synchronous throw trips the ErrorBoundary. That was true when it was written and is what made the rejected-promise choice necessary; it isn't true anymore, so the comment now says only that a rejection is the shape a sleeping worker actually produces (685d54b9).

src/background/redux/surfaced-dispatch-actions.ts
ost-ptk: Added `windowRequestWindowAttached.type` and kept `ledgerNewWindowIdChanged`, as you suggested rather than swapping them — 71cc9928.

Both comments retargeted: the attach entry now describes the hang it actually causes, and the `windowId` entry describes what you found — `windowId` stays null, the close tracker detaches, and the permission window is never closed when the signature completes. The note that `triggeredRef` is set regardless of either outcome, so neither has a retry path, sits with the pair.

The exhaustive pin moved with them, and its name lost "seven" — the set is ten now.

src/background/redux/surfaced-dispatch-actions.ts
ost-ptk: Both added — 71cc9928. Not a deliberate exclusion; they were missed.

Surfacing alone wasn't enough at either site, so the two navigations are now gated on the verdict too: `navigate(ConfirmSecretPhraseSuccess)` and `saveIsLoggedIn(true)` were firing on a dropped send.

`create-vault-password` needed one more change to make the banner actionable. The button was disabled on `isSubmitSuccessful`, which becomes true on a dropped dispatch as well — so the user would get the banner on the one screen whose only button had just been disabled for good. Swapped for an explicit `isSubmitted` set only on a true verdict, with `isSubmitting` covering the in-flight case.

src/libs/ui/components/saga-error-banner/ui-error-channel.ts
ost-ptk: Added `clearUiError(kind, detail)` in 29b84dab, called from `dispatchToMainStore`'s success path and the `openWindow` success path as you described.

It returns before `emit()` when no row matches the key, so the ~105 dispatches that never failed neither allocate nor re-render. That's pinned rather than assumed: the test asserts `getUiErrorsSnapshot()` is reference-identical across a no-op clear and that no subscriber fires, then that a real clear notifies exactly once.

src/background/redux/surfaced-dispatch-actions.test.ts
ost-ptk: Agreed — it exercised nothing a rename could break. Replaced rather than renamed (71cc9928), taking the second of your two options.

The test now asserts the same set against the type strings as literals — the only thing in the file that doesn't come from a creator — so a renamed slice reducer fails it while the creator-derived pin sails through. The narrow case you identified, dropping an entry *and* updating the expected literal, stays caught by the first pin.

src/hooks/use-window-manager.ts
ost-ptk: Extracted as `createReportingOpenWindow(open)` in ff58feba, following `register-ledger-permission-window.ts` as you suggested; `useWindowManager` is now the `useMemo` around it.

`create-reporting-open-window.test.ts` pins the redaction against the regression you named. The leak case builds the rejection out of a `searchParams` plaintext and asserts the exact `console.error` arguments — `'openWindow failed'`, the `windowApp`, `'TypeError'` — plus that the secret appears nowhere in the recorded calls. Widening back to the rejection or to `props` fails it. Also covered: a non-`Error` rejection logs `undefined` rather than throwing on `.name`, and a successful open logs nothing and clears the row (per the `ui-error-channel` thread).

src/libs/layout/reset-vault/index.tsx
ost-ptk: Covered in 685d54b9, along the lines you suggested. `breakTransport` and the copy constants moved to `e2e-tests/error-surface.ts` so the specs share one definition instead of `dropped-dispatch.spec.ts` owning them privately.

Two cases, on the flows that already exist:

- `wallet.spec.ts` — transport broken after the checkbox, then the final "Reset wallet" click; asserts the banner and that the "Are you sure you want to reset your wallet?" heading is still on screen, which is precisely what the unconditional `closeWindowByReloadExtension()` destroyed.
- `import-account-with-file.spec.ts` — asserts the banner and `toHaveCount(0)` on the success heading. The break lands only after the file and the account name are in: the page dispatches while it loads, so breaking earlier fails the flow before the Import button is reached.

Both are running in CI on this push — I haven't run them locally, so I can't claim to have confirmed the mutation check (deleting `if (dispatched)`) end to end.

The other five gated sites — welcome, create-secret-phrase, window-page, torus, connected-ledger — are still unpinned; no existing spec drives them, and each would need its flow built from scratch.

ost-ptk: Following up on my earlier reply — the import case failed in CI, and the fix is 12d332e5.

The cause was in the test, not the guard. `breakTransport` broke the whole transport, and the import flow puts two legacy sends on it that never go through `dispatchToMainStore`: `checkAccountNameIsTaken`, which the yup resolver re-runs during submit-time validation, and `checkSecretKeyExist`, inside the file reader. The first rejection stopped `handleSubmit` before `onSubmit`, so the guarded dispatch was never reached and no banner was ever due. The captured snapshot shows it — the form still on screen, neither the success screen nor the failure route.

`breakTransport` now takes an optional list of action types and passes everything else through to the real `sendMessage`; the import case aims at `vault/accountImported`. The default is unchanged, so the reset case and `dropped-dispatch.spec.ts` still break the whole transport.

Also correcting my earlier reply, where I said I hadn't confirmed the mutation check: I have now. With `if (dispatched)` removed from the reset page and the `if (await dispatchToMainStore(...))` removed from the import page, both cases fail on their own assertions — the reset one on the confirmation heading, the import one on `toHaveCount(0)` — and both pass with the guards restored. Full popup suite locally: 57 passed, 1 skipped, with `contacts.spec.ts` flaky on an unrelated case.

The other five gated sites remain unpinned, as above.

Comp0te: The two cases land, and the mutation run is the part I was after — both failing with the guard removed and passing with it restored is the pin.

On the remaining five: two of the sites this push newly gated aren't in that bucket, and their flows already exist.

`confirm-secret-phrase/index.tsx:50-56` and `create-vault-password/index.tsx:65-68` are both reached today — `fixtures.ts:223-256` walks the confirmation screen, `:144-167` the password one — and the `onlyTypes` filter you just added is what makes them expressible: `breakTransport(page, ['INIT_VAULT_SAGA'])` and `(['INIT_KEYS_SAGA'])` (`sagas/actions.ts:18-21`). Welcome, create-secret-phrase, window-page, torus and connected-ledger I agree with; nothing drives them.

</details>

Comp0te (commented):
Re-read at `12d332e5`. The round-1 points are all closed, and the fixes are the ones I'd have written. What's left is the same shape in three new places: a continuation gated on the dispatch verdict but with nothing guarding the second click, the third onboarding write left out of the surfaced set, and a spec narrowed past the failure it was written for.

src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
```diff
@@ -44,8 +44,16 @@ export function ConfirmSecretPhrasePage({
   function handleSubmit() {
     try {
       if (isConfirmationSuccess && phrase) {
-        dispatchToMainStore(initVault({ secretPhrase: phrase }));
-        navigate(RouterPath.ConfirmSecretPhraseSuccess);
+        // Only navigate once the vault actually exists: keys and encryption key
+        // are already in place here, so the success page renders regardless of
+        // whether this dispatch arrived.
+        dispatchToMainStore(initVault({ secretPhrase: phrase })).then(
```
Comp0te, line 58: `:96` is still `disabled={!isFormValid}`, and `handleSubmit` re-reads only `isConfirmationSuccess` and `phrase` — neither changed by the first click. A second click inside this round trip dispatches `initVault` again, leaving the new wallet with two identical "Account 1" entries sharing one key pair.

`create-vault-password:127` gained `isSubmitting || isSubmitted` in this same push; mirroring it here closes the window.

`initVault` is in `FORWARDED_ACTION_TYPES` (`redux-actions.ts:140`) and reaches `store.dispatch` at `:309-311` with no dedupe. `initVaultSaga` (`onboarding-sagas.ts:116-141`) is synchronous throughout — `deriveKeyPair` is a plain sync call, `disableOnboardingFlow()` is called not yielded — so the task ends inside the first delivery and `takeLatest` has nothing to cancel. `accountAdded` (`vault/reducer.ts:423-431`) appends with no publicKey or name dedupe.

Nothing else shuts the window: `app-router.tsx:40-56` branches only on `keysDoesExist && encryptionKeyDoesExist`, already true here; `phrase` is plain `useState` (`use-onboarding-form-state.ts:10-25`).

At the base `navigate` ran in the click handler — one scheduler tick, not a background round trip.

</details>

src/background/redux/surfaced-dispatch-actions.ts
```diff
@@ -0,0 +1,53 @@
...
+  // signature completes. `triggeredRef` is set regardless of either outcome, so
+  // neither has a retry path.
+  ledgerNewWindowIdChanged.type,
+  // Without this the reload presents an unperformed reset as done.
+  resetVault.type,
+  // Onboarding's two writes. Without `initKeys` the password screen keeps
+  // rendering and nothing happens; without `initVault` the success screen
+  // renders over a vault that was never created.
+  initKeys.type,
+  initVault.type,
```
Comp0te, line 45: The comment at `:39-41` says "Onboarding's two writes", but there are three. `recoverVault` is dispatched at `select-accounts-to-recover/index.tsx:111-116` as a verdict-blind `.finally(closeActiveTab)`, in neither the set nor the pins. A dropped send closes the onboarding tab as though it worked: no accounts, no vault, no message.

Adding the type and gating the close would match the sites you fixed.

`closeActiveTab` (`onboarding/utils/close-active-tab.ts:3-16`) removes the onboarding tab itself, so the type alone is not enough — `SagaErrorBanner` is mounted in that tree (`onboarding/index.tsx:44`) and dies with it. On a dropped send `dispatchToMainStore` resolves `false`, `recoverVaultSaga` (`onboarding-sagas.ts:143-161`) never runs.

`recoverVault` has one dispatch site repo-wide, and neither pin lists `RECOVER_VAULT_SAGA` (`surfaced-dispatch-actions.test.ts:19-33`, `:38-53`). `recover-secret-phrase-flow.spec.ts:41-45` clicks this button but asserts only success-path state.

The same shape at `reset-vault/index.tsx:30` went from a verdict-blind `.then` at the base to the gated form at head; this one was missed.

</details>

e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
```diff
@@ -62,6 +63,57 @@ popup.describe('Popup UI: import account with file', () => {
...
+      await importAccountPage
+        .getByPlaceholder('Account name', { exact: true })
+        .fill(ACCOUNT_NAMES.importedPemAccountName);
+
+      // Only the guarded dispatch: `checkAccountNameIsTaken` rides the same
+      // transport from submit-time validation, so breaking all of it stops
+      // `handleSubmit` before `onSubmit` runs and the page never gets as far as
+      // the action under test. The literal is pinned in
+      // `surfaced-dispatch-actions.test.ts`.
+      await breakTransport(importAccountPage, ['vault/accountImported']);
```
Comp0te, line 109: The narrowing is well explained, but it aims the spec off its target: `checkAccountNameIsTaken` is a bare `runtime.sendMessage` inside the yup resolver, so a whole-transport break rejects `handleSubmit` — the guard at `index.tsx:32` never runs, no banner due.

Probably a follow-up on the legacy sends, not this PR — but `:97-101` should say the spec covers only the late-failure window.

yup 1.7.1 routes a non-`ValidationError` from a test to `panic`, which is `validate`'s `reject`; `@hookform/resolvers` 5.4.0 rethrows anything without `.inner`. `reportUiError` has two call sites repo-wide (`redux/utils.ts:50`, `create-reporting-open-window.ts:34`), neither on this path.

The form is `mode: 'onChange'` (`:80`) with `disabled={!isValid || !isDirty}` (`:108`), so there are two silent states, split by when the transport drops:

- before or during typing — change-time validation rejects, `isValid` never turns true, Import never becomes pressable;
- after the field last validated — Import is pressable but inert, guard bypassed.

`checkSecretKeyExist` escapes the same way, with `use-secret-key-file-reader.ts:78-84` routing it to `onFailure` as raw untranslated text.

</details>

src/apps/onboarding/pages/confirm-secret-phrase/index.tsx
ost-ptk: Confirmed and fixed in 2243b9d3 — and it is a regression this PR introduced, not a pre-existing one: awaiting the verdict is what replaced the synchronous `navigate` with a round trip.

I checked the downstream claim rather than taking the shape on trust, and each link holds. `initVaultSaga` runs to completion inside one delivery, so `takeLatest` has nothing to cancel. `accountAdded` appends with no dedupe on name or public key. And `disableOnboardingFlow` does not save it either — it only repoints the browser action and removes a storage key, so the second message still reaches the handler and dispatches. Two identical "Account 1" entries on one key pair, as you described.

`isSubmitting` now mirrors `create-vault-password:127`, releasing on a false verdict so the banner still has a button to point at.

src/background/redux/surfaced-dispatch-actions.ts
ost-ptk: You're right, and it was a miss rather than a decision — fixed in e2196cdc. The comment now says three writes.

Both halves, since the type alone would have been useless for exactly the reason you gave: the close is gated on the verdict, and `RECOVER_VAULT_SAGA` is in the set and both pins.

The gate brought a second problem with it, the same one from the `confirm-secret-phrase` thread: keeping the tab open on failure opens a round-trip window on a button nothing disables, and `accountsAdded` appends without a dedupe, so a second click would double every recovered account. `isSubmitting` covers it there too.

One limit worth stating plainly: the new e2e case for this cannot pin the gate. `closeActiveTab` is inert under Playwright — the success-path specs `page.goto` the popup on the same `page` right after a recover that closes the tab in a real browser, which would throw on a closed target. So the case pins that the banner reaches this tree, and reverting the gate to `.finally` leaves it green. That's recorded in the spec.

e2e-tests/popup/import-account-with-file/import-account-with-file.spec.ts
ost-ptk: Added in a2db8de0 — `:97-101` now says the case covers only the late-failure window, and names both legacy sends and why no guard on this page can reach them.

Agreed on the follow-up. Your split into two silent states matches what I found: with the transport down before or during typing, change-time validation rejects and Import never becomes pressable; after the field last validated, Import is pressable but inert. `checkSecretKeyExist` is the same shape one step later, surfacing as raw untranslated text through `onFailure`.

Neither is reachable from `dispatchToMainStore`, so fixing them means giving those two legacy sends a verdict of their own — its own change, as you say.

src/libs/layout/reset-vault/index.tsx
ost-ptk: You're right that those two aren't in the unpinnable bucket — both now have a case, plus a third for `recoverVault`, in `e2e-tests/onboarding-flow/dropped-dispatch-flow.spec.ts` (a2db8de0). The remaining five stand as agreed.

Worth reporting how that went, because the mutation run is what picked the assertions and two of my three first attempts pinned nothing.

- `initKeys`: asserting the banner and the URL passes with the guard removed — `keysDoesExist` is false either way, so `NoVaultRoutes` keeps the screen up on its own. The guard's observable effect is that `isSubmitted` does not latch on a write that never landed, so the case asserts the button is still enabled. Fails on mutation now.
- `initVault`: same release of its in-flight gate. Fails on mutation.
- `recoverVault`: my first version passed under mutation by winning the race against an async `closeActiveTab`. Tightening it didn't help, because `closeActiveTab` is inert under Playwright — the success-path specs `page.goto` the popup on the same `page` right after a recover that closes the tab in a real browser. So this one pins only that the banner reaches that tree; the gate itself is held by review, and the spec says so rather than implying coverage it doesn't have.

Verified: onboarding suite 9/9 with `--workers=1`. At 3+ workers `confirm-secret-phrase-flow.spec.ts` goes flaky on the word picker and passes on retry — pre-existing and unrelated to the gate (`isSubmitting` is provably false there, since it is the first click that fails to land), and `--repeat-each=4` on that spec alone is 8/8.

Comp0te (approved):

PR #1466, opened 2026-08-13
fix: WALLET-1364 — close the P3 residual

ost-ptk:
Closes the six residual P3 defects recorded in WALLET-1364. #1465 — which introduced the
`runKeysDownload` extraction this branch builds on — is merged, so this now sits directly on
`develop`.

Jira: https://make-software.atlassian.net/browse/WALLET-1364

## What each change removes

| #   | Defect                                                        | What the user experienced                                                                                                                                                                                                                                                                                            |
| --- | ------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 10  | Account mutations waited out the re-encrypt debounce          | A browser quit or MV3 worker kill inside the 500 ms window dropped the last vault mutation. When that was an **imported** account, the secret key was gone — it exists nowhere else. The four account-mutating actions now persist immediately via `takeEvery`; everything re-derivable or cosmetic stays debounced. |
| 11  | A failed password change looked exactly like a successful one | `navigate(Home)` sat outside both worker callbacks and ran unconditionally. The user landed on Home believing their password had changed when it had not. Navigation now happens only on a successful worker result; a failure shows a field error and stays put.                                                    |
| 12  | Unlock worker errors said nothing                             | Both error paths stopped the spinner with no message — indistinguishable from a wrong password, which the same screen reports properly a few lines above.                                                                                                                                                            |
| 13  | A partial key archive was presented as success                | Accounts whose `createAsymmetricKeys` returned `secretKey: null` were silently omitted from the zip while Success was shown. Now the export fails loudly, **before** anything is written to disk, and logs a count only.                                                                                             |
| 14  | Password hash compared with `===`                             | Short-circuits at the first differing character. Now a constant-time comparison over the decoded bytes.                                                                                                                                                                                                              |
| 16  | Non-Chrome `style-src` had no `'self'`                        | A source list of only `'unsafe-inline'` matches no external stylesheet, and every app template links `/assets/fonts/fonts.css`. Fixed in `webpack.config.js` and in Safari's hand-maintained runtime CSP in `src/utils.ts`, which had drifted the same way.                                                          |

Item 15 (QR sync uses AES-CBC with no MAC) is **deliberately not fixed** — see below.

## What the review round changed

Three findings, all confirmed against the code before acting on them. The first was the important
one: it meant items 11 and 12 were not actually delivered.

**`onerror` could not fire for the failures it was added to report.** Every worker in this repo is
written `onmessage = async` with no `try`/`catch`, so anything that throws inside — `encodePassword`,
`deriveEncryptionKey`, `decryptVault`'s GCM tag check, `encryptVault`'s missing-vault guard —
becomes a rejected promise, and a rejection inside a worker raises no `error` event on the parent
`Worker`. Only a script *load* failure reaches `onerror`. In `change-password` that was worse than
the bug it replaced: `isSubmitting` was cleared only inside `onerror`, so a rejected worker left the
button disabled on `Loading` forever, with no message and no navigation. Before this PR the user at
least left the page.

Failures now travel as a message. `src/background/workers/types.ts` holds the contract —
`{ error: true }` with no payload (the error object stays in the worker's console rather than
reaching the UI), plus `WorkerResult<T>` and an `isWorkerError` guard. Each worker wraps its body;
each page branches on the guard. `onerror` is kept for the script-load case it does cover.

**The defect was wider than this PR's diff.** `generate-sync-wallet-qr-data-worker.ts` has the same
shape and the same dead `onerror`, and it can throw for real — `scryptAsync`, `PublicKey.fromHex` on
a corrupt imported key. Worse, `generateQRCode` was `async` but resolved as soon as the worker was
constructed, so the `.catch` that `PasswordProtectionPage` already wraps `onClick` in was
unreachable and the page span forever. It now returns a promise that resolves when the QR data
exists and rejects on failure. **The cipher is untouched** — this is delivery of the failure, not the
format; item 15 below still stands.

**`password-protection-page` had to be pulled in.** It consumes the same verify-password worker, and
without handling `{ error: true }` a worker failure would have destructured `isPasswordCorrect:
undefined` and read as a *wrong password* — dispatching `loginRetryCountIncremented()`, burning a
login attempt and eventually locking the wallet. That would have been a new bug, not a leftover. Its
own `onerror` is no longer silent either, which closes a parity gap this description previously left
open.

**The change-password worker outlived its page.** It was never terminated and there was no unmount
cleanup, while the back link stays live throughout — only the submit button is disabled. Pressing
back during the worker's ~1-2 s of scrypt still committed the rotation from a stale closure, so the
user's *next* unlock with the old password would fail. React-router does not stop the `navigate` that
follows either: in the installed 7.18.2, `activeRef.current = true` is set in a layout effect with no
cleanup, so the post-unmount guard passes. The worker is now held in a ref and terminated on unmount,
and both callbacks check a mounted flag. The back link stays enabled — cancelling mid-scrypt now
genuinely cancels. `unlock-vault` was also spawning two workers per submit and terminating neither;
it now disposes them.

**The constant-time comparison was unwireable without turning anything red.** `constantTimeEqualHex`
was tested thoroughly in isolation, but nothing observed that `verifyPasswordAgainstHash` routed
through it — its three tests assert only correct → truthy and wrong → falsy, which plain `===`
satisfies identically. Reverting the call site left the suite green. An upper-case hash is the one
input where the two differ, and that assertion is now in the suite.

## Two things an earlier review found

**A watch account used to break the whole export.** Watch accounts are stored with `secretKey: ''`
and no `hardware` flag, so they passed the export list's `!account.hardware` filter. Item 13's
fail-loudly rule then meant a user with three real accounts and one watch account got **no archive
at all**. That is worse than the bug being fixed, so the export list now excludes accounts with
nothing to export (`selectVaultAccountsAvailableForExport`), and the hard failure is reserved for a
genuinely unexpected null — data corruption, which should still never be shown as success.

**Change-password now waits for the worker, so it needed a submitting state.** Navigation correctly
moved inside the success path, which leaves the page on screen for the worker's two scrypt passes
(N=2¹⁸, roughly 1-2 s). Without a guard, two taps spawn two ~256 MB scrypt workers in the popup.
It now mirrors the `isSubmitting` pattern already used by `password-protection-page`.

## Limitations, stated rather than buried

- **The constant-time property is not test-enforceable here.** Mutating the loop to `break` on the
  first mismatch does not change the output — `|=` is monotonic — only the timing. What the tests now
  prove is that the comparison is _correct_ and that the call site routes through it; the in-code
  comment is still the only thing standing between a future edit and a reintroduced short-circuit.
- **The React paths have no automated coverage.** This project's jest runs
  `testEnvironment: 'node'` with no jsdom, so `change-password`, `unlock-vault`,
  `password-protection-page` and `wallet-qr-code` cannot be unit-tested. Code reading was the only
  verification. Items to click through are listed below. The workers themselves _are_ covered —
  see `src/background/workers/workers.test.ts`, which drives all four through a stubbed
  `globalThis.onmessage`/`postMessage`.
- `change-password` disables the submit button while the worker runs, but unlike its sibling
  `password-protection-page` it does not also guard Enter re-submission or make the field
  read-only. The button guard covers the double-tap vector; the parity gap is noted rather than
  closed here.
- The string `Something went wrong. Please try again.` has no catalogue entry, so the eight
  non-English locales fall back to English. `locale:extract_pot` was deliberately not run — `lang/`
  is hundreds of msgids behind and regenerating it would swamp this diff. A shorter
  `Something went wrong` already exists in four places; consolidating is a reasonable follow-up.

## Why item 15 is not here

`generate-sync-wallet-qr-data-worker.ts` still uses AES-CBC with no MAC. The format is a
**cross-application contract**: `casper-wallet-mobile/src/utils/aes/index.ts` decrypts it with
`aes-256-cbc` and identical scrypt parameters (N = 2¹⁸, r = 8, p = 1, dkLen = 32), and its
`react-native-aes-crypto` exposes CBC only. Changing the extension alone breaks wallet sync until a
matching mobile release ships. The residual risk is also narrow — the ciphertext travels from the
user's own screen to their own camera. If picked up, it needs a versioned envelope landed on mobile
first. The review round touched this worker's *error delivery* only; the ciphertext it produces is
byte-for-byte what it was.

## Please verify before merging

The CSP change needs a real browser, which is the one thing this branch cannot settle itself:

1. `npm run build:firefox`, load `build/firefox` via `about:debugging`, open the popup and the
   onboarding page, and check the console for a blocked-stylesheet violation on `fonts.css` —
   before and after. Same for the Safari build.
2. Change password: confirm it lands on Home only after the worker finishes, and that the button is
   disabled while it runs. Then start a change and press **back** during the scrypt pass — the old
   password must still unlock afterwards.
3. Unlock with a wrong password — the message must still be the wrong-password one, not
   `Something went wrong`.
4. Sync wallet QR — confirm the happy path still produces a scannable code, since its page function
   was promisified.
5. Export keys with a watch account present, using "select all" — the watch account should not be
   offered at all.

`npm run ci-check` is green: 94 suites, 903 tests.

### DIFF
--- a/src/apps/popup/pages/change-password/index.tsx
+++ b/src/apps/popup/pages/change-password/index.tsx
@@ -49,7 +49,8 @@ export const ChangePasswordPage = () => {
     register,
     handleSubmit,
     formState: { isDirty, errors },
-    control
+    control,
+    setError
   } = useCreatePasswordForm();
 
   const password = useWatch({
@@ -101,13 +102,16 @@ export const ChangePasswordPage = () => {
           vaultCipher: newVaultCipher
         })
       );
+
+      navigate(RouterPath.Home);
     };
 
     worker.onerror = error => {
       console.error(error);
+      setError('password', {
+        message: t('Something went wrong. Please try again.')
+      });
     };
-
-    navigate(RouterPath.Home);
   };
 
   if (!isPasswordConfirmed) {
--- a/src/apps/popup/pages/change-password/index.tsx
+++ b/src/apps/popup/pages/change-password/index.tsx
@@ -39,6 +39,7 @@ interface CreatePasswordWorkerMessageEvent extends MessageEvent {
 export const ChangePasswordPage = () => {
   const [isPasswordConfirmed, setIsPasswordConfirmed] =
     useState<boolean>(false);
+  const [isSubmitting, setIsSubmitting] = useState(false);
 
   const { t } = useTranslation();
   const navigate = useTypedNavigate();
@@ -63,13 +64,16 @@ export const ChangePasswordPage = () => {
   }, []);
 
   const isSubmitButtonDisabled = calculateSubmitButtonDisabled({
-    isDirty
+    isDirty,
+    isSubmitting
   });
 
   const onSubmit = (data: CreatePasswordFormValues) => {
     const worker = new Worker(
       new URL('@background/workers/create-password-worker.ts', import.meta.url)
     );
+
+    setIsSubmitting(true);
 
     worker.postMessage({
       password: data.password,
@@ -111,6 +115,7 @@ export const ChangePasswordPage = () => {
       setError('password', {
         message: t('Something went wrong. Please try again.')
       });
+      setIsSubmitting(false);
     };
   };
 
@@ -146,7 +151,7 @@ export const ChangePasswordPage = () => {
       renderFooter={() => (
         <FooterButtonsContainer>
           <Button disabled={isSubmitButtonDisabled}>
-            <Trans t={t}>Continue</Trans>
+            {isSubmitting ? t('Loading') : <Trans t={t}>Continue</Trans>}
           </Button>
         </FooterButtonsContainer>
       )}
--- a/src/apps/popup/pages/change-password/index.tsx
+++ b/src/apps/popup/pages/change-password/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useWatch } from 'react-hook-form';
 import { Trans, useTranslation } from 'react-i18next';
 import { useSelector } from 'react-redux';
@@ -12,6 +12,7 @@ import { encryptionKeyHashCreated } from '@background/redux/session/actions';
 import { dispatchToMainStore } from '@background/redux/utils';
 import { vaultCipherCreated } from '@background/redux/vault-cipher/actions';
 import { selectVault } from '@background/redux/vault/selectors';
+import { WorkerResult, isWorkerError } from '@background/workers/types';
 
 import {
   FooterButtonsContainer,
@@ -27,13 +28,13 @@ import {
 import { calculateSubmitButtonDisabled } from '@libs/ui/forms/get-submit-button-state-from-validation';
 
 interface CreatePasswordWorkerMessageEvent extends MessageEvent {
-  data: {
+  data: WorkerResult<{
     passwordHash: string;
     passwordSaltHash: string;
     newEncryptionKeyHash: string;
     keyDerivationSaltHash: string;
     newVaultCipher: string;
-  };
+  }>;
 }
 
 export const ChangePasswordPage = () => {
@@ -45,6 +46,9 @@ export const ChangePasswordPage = () => {
   const navigate = useTypedNavigate();
 
   const vault = useSelector(selectVault);
+
+  const workerRef = useRef<Worker | null>(null);
+  const isMountedRef = useRef(true);
 
   const {
     register,
@@ -58,6 +62,18 @@ export const ChangePasswordPage = () => {
     control,
     name: 'password'
   });
+
+  // the back link stays live while scrypt runs; without this the rotation would
+  // still commit from a stale closure and the old password would stop working
+  useEffect(() => {
+    isMountedRef.current = true;
+
+    return () => {
+      isMountedRef.current = false;
+      workerRef.current?.terminate();
+      workerRef.current = null;
+    };
+  }, []);
 
   const setPasswordConfirmed = useCallback(() => {
     setIsPasswordConfirmed(true);
@@ -73,6 +89,7 @@ export const ChangePasswordPage = () => {
       new URL('@background/workers/create-password-worker.ts', import.meta.url)
     );
 
+    workerRef.current = worker;
     setIsSubmitting(true);
 
     worker.postMessage({
@@ -80,7 +97,29 @@ export const ChangePasswordPage = () => {
       vault
     });
 
+    const disposeWorker = () => {
+      worker.terminate();
+      workerRef.current = null;
+    };
+
+    const handleFailure = () => {
+      disposeWorker();
+      setError('password', {
+        message: t('Something went wrong. Please try again.')
+      });
+      setIsSubmitting(false);
+    };
+
     worker.onmessage = (event: CreatePasswordWorkerMessageEvent) => {
+      if (!isMountedRef.current) {
+        return;
+      }
+
+      if (isWorkerError(event.data)) {
+        handleFailure();
+        return;
+      }
+
       const {
         passwordHash,
         passwordSaltHash,
@@ -88,6 +127,8 @@ export const ChangePasswordPage = () => {
         keyDerivationSaltHash,
         newVaultCipher
       } = event.data;
+
+      disposeWorker();
 
       dispatchToMainStore(
         keysUpdated({
@@ -110,12 +151,16 @@ export const ChangePasswordPage = () => {
       navigate(RouterPath.Home);
     };
 
+    // only reached by a script load failure — a rejection inside the worker
+    // arrives through onmessage instead
     worker.onerror = error => {
       console.error(error);
-      setError('password', {
-        message: t('Something went wrong. Please try again.')
-      });
-      setIsSubmitting(false);
+
+      if (!isMountedRef.current) {
+        return;
+      }
+
+      handleFailure();
     };
   };
 
--- a/src/libs/crypto/hashing.ts
+++ b/src/libs/crypto/hashing.ts
@@ -39,6 +39,12 @@ export function constantTimeEqualHex(a: string, b: string): boolean {
   const left = convertHexToBytes(a);
   const right = convertHexToBytes(b);
 
+  // Buffer.from(hex) truncates at a dangling nibble or the first non-hex
+  // character, so a short decode means the input was not canonical hex.
+  if (left.length !== a.length / 2 || right.length !== b.length / 2) {
+    return false;
+  }
+
   let diff = 0;
   // must scan every byte — returning early on the first mismatch reintroduces the timing leak
   for (let i = 0; i < left.length; i++) {
--- a/src/libs/crypto/hashing.ts
+++ b/src/libs/crypto/hashing.ts
@@ -31,6 +31,23 @@ export async function encodePassword(
     });
 }
 
+export function constantTimeEqualHex(a: string, b: string): boolean {
+  if (a.length !== b.length) {
+    return false;
+  }
+
+  const left = convertHexToBytes(a);
+  const right = convertHexToBytes(b);
+
+  let diff = 0;
+  // must scan every byte — returning early on the first mismatch reintroduces the timing leak
+  for (let i = 0; i < left.length; i++) {
+    diff |= left[i] ^ right[i];
+  }
+
+  return diff === 0;
+}
+
 export async function verifyPasswordAgainstHash(
   passwordHash: string,
   passwordSaltHash: string,
@@ -38,7 +55,7 @@ export async function verifyPasswordAgainstHash(
 ): Promise<boolean> {
   const digest = await encodePassword(password || '', passwordSaltHash);
 
-  return passwordHash === digest;
+  return constantTimeEqualHex(passwordHash, digest);
 }
 
 export async function deriveEncryptionKey(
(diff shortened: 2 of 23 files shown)


src/apps/popup/pages/change-password/index.tsx
```diff
@@ -101,13 +106,17 @@ export const ChangePasswordPage = () => {
           vaultCipher: newVaultCipher
         })
       );
+
+      navigate(RouterPath.Home);
```
Comp0te, line 151: Separate from the `onerror` point below — this one is about the success path.

The `Worker` created in `onSubmit` is never terminated and there's no unmount cleanup (no `useEffect` in this file, and no `terminate()` call anywhere in `src`), while the back link stays live throughout: `HeaderSubmenuBarNavLink linkType="back"` renders unconditionally at line 137 and has no disabled path (`header-submenu-bar-nav-link.tsx:84-96`) — only the submit button is disabled.

So during the worker's ~1-2 s of scrypt the user can press back, and the rotation lands anyway: the three `dispatchToMainStore` calls at lines 92-108 commit the new password from a closure that never consults mount state. Their next unlock with the old password then fails.

`navigate(RouterPath.Home)` on this line follows them from that same stale closure, and react-router's post-unmount guard doesn't stop it: in `react-router@7.18.2`, `useNavigateUnstable` sets `activeRef.current = true` in a layout effect **with no cleanup**, so it stays `true` after unmount, the `if (!activeRef.current) return` check passes, and the callback goes on to `navigator.push` against the still-live `HashRouter` history. The user gets pulled to Home from wherever they'd navigated to.

Suggested fix: hold the worker in a ref, `terminate()` it on unmount, and guard the callbacks with a mounted/abort flag. If cancellation isn't wanted at all, disabling the back link while `isSubmitting` is the smaller change.

src/apps/popup/pages/change-password/index.tsx
```diff
@@ -101,13 +106,17 @@ export const ChangePasswordPage = () => {
...
     };
 
     worker.onerror = error => {
       console.error(error);
+      setError('password', {
```
Comp0te: Separate from the worker-lifetime point above — this is about whether this handler can run at all.

All three password workers are written `onmessage = async` with no `try`/`catch`: `create-password-worker.ts:18-39`, `unlock-vault-worker.ts:24-53`, `verify-password-worker.ts:11-22`. Every call that can fail — `encodePassword`, `deriveEncryptionKey`, `decryptVault`'s GCM tag check, `encryptVault`'s "Vault State doesn't exist" — sits inside that async body, so a throw becomes a **rejected promise**, and a rejection inside a worker raises no `error` event on the parent `Worker`. Only a script *load* failure (missing chunk, CSP) reaches `onerror`. There's no `onmessageerror` handler anywhere in `src` either.

To see it, in the popup's devtools console:

```js
const w = new Worker(URL.createObjectURL(new Blob(
  ["onmessage = async () => { throw new Error('boom') }"], {type: 'text/javascript'})));
w.onerror   = () => console.log('onerror');    // never logs
w.onmessage = () => console.log('onmessage');  // never logs
w.postMessage(1);
```

Same result whether the async handler throws synchronously or awaits a rejection; only a non-`async` `onmessage` that throws fires `onerror`.

The cost is worse here than in unlock-vault. `isSubmitting` is cleared in exactly one place — line 118, inside this handler — so when the worker rejects the button stays disabled rendering `t('Loading')` indefinitely, no field error appears, and the `navigate(RouterPath.Home)` you moved into `onmessage` never runs. Before this PR the user at least left the page. In `unlock-vault` the behaviour is genuinely unchanged (`setIsLoading(false)` was already there and equally never ran) — what's new there is only that the fix reads as delivered.

Suggested fix: wrap each worker's body in `try`/`catch` and `postMessage({ error: … })` on failure, then handle that message in the page the way `onerror` is handled now.

src/libs/crypto/hashing.ts
```diff
@@ -31,14 +31,37 @@ export async function encodePassword(
...
+
 export async function verifyPasswordAgainstHash(
   passwordHash: string,
   passwordSaltHash: string,
   password: string | undefined
 ): Promise<boolean> {
   const digest = await encodePassword(password || '', passwordSaltHash);
 
-  return passwordHash === digest;
+  return constantTimeEqualHex(passwordHash, digest);
```
Comp0te, line 64: Different from the limitation you flagged in the description: you noted that mutating the loop to `break` early can't be caught because `|=` is monotonic — that's right. But the *call site* is unprotected in a way a test can catch.

`constantTimeEqualHex` is tested thoroughly in isolation (`hashing.test.ts:69-102` — the canonical-hex guard, the length guard and the non-hex cases all discriminate). Nothing observes that `verifyPasswordAgainstHash` routes through it: its three tests assert only correct-password → truthy and wrong-password → falsy, which plain `===` satisfies identically.

Revert this line to `return passwordHash === digest;` and run `npx jest src/libs/crypto/hashing.test.ts` — 1 suite, 11 tests, all green. So the helper can end up exported, tested and unwired while the short-circuit this PR was filed to close is back.

One assertion closes it, on the one input where the two differ:

```ts
expect(verifyPasswordAgainstHash(
  FIXED_PASSWORD_HASH.toUpperCase(), FIXED_PASSWORD_SALT, FIXED_PASSWORD_TEXT
)).toBe(true);
```

It also pins the hex case-insensitivity this introduces as a side effect — `'AB'` now equals `'ab'` where `===` said false. Harmless today since the wallet only emits lowercase, but currently unstated.

Comp0te (commented):
Reviewed the six residual fixes against the flows they touch. The comments cluster on the two items that moved work into worker callbacks: the new `onerror` handlers can't fire for the failures they were added to report, and the change-password page now outlives its worker without owning it. One more is about the constant-time comparison being unwired-able at its call site without turning anything red. The CSP and export changes read as sound, and the item-15 rationale is convincing as written.

src/apps/popup/pages/change-password/index.tsx
ost-ptk: Confirmed, and it turned out to be wider than this PR. Fixed in bcd2289f and 1fb6cf59.

All four workers in the tree are `onmessage = async` with no try/catch, so failures become rejected promises and never raise an `error` event on the parent — the QR worker (`generate-sync-wallet-qr-data-worker.ts`) has the same shape and the same dead `onerror` at `wallet-qr-code/index.tsx:67`, outside this PR's diff.

Failures now travel as a message. `src/background/workers/types.ts` holds the contract: `{ error: true }` with no payload — the error object stays in the worker's console rather than reaching the UI — plus `WorkerResult<T>` and an `isWorkerError` guard. Every worker wraps its body; every page branches on the guard. `onerror` is kept for the script-load case it does cover, now with a comment saying so.

Two consequences worth flagging:

**`password-protection-page` had to learn the message too.** It consumes the same verify-password worker, and without handling `{ error: true }` a worker failure would have destructured `isPasswordCorrect: undefined` and read as a wrong password — dispatching `loginRetryCountIncremented()`, burning a login attempt and eventually locking the wallet. That would have been a new bug, not a leftover. Its own `onerror` is no longer silent either, which closes the parity gap the description had left open.

**The QR page hung rather than stuck.** `generateQRCode` was `async` but resolved as soon as the worker was constructed, so the `.catch` that `PasswordProtectionPage` already wraps `onClick` in was unreachable and `setIsLoading(false)` lived only in the dead `onerror`. It now returns a promise that resolves when the QR data exists and rejects on failure, which routes into that existing catch — and the catch sets a field message instead of only logging.

Covered by `src/background/workers/workers.test.ts`: all four workers driven through a stubbed `globalThis.onmessage`/`postMessage`, asserting the rejection comes back as `{ error: true }`. Removing a `try`/`catch` turns that worker's case red — checked per worker, not assumed.

src/apps/popup/pages/change-password/index.tsx
ost-ptk: Confirmed on every step — no `terminate()` anywhere in `src`, no unmount cleanup, and the `'back'` branch of `header-submenu-bar-nav-link.tsx` is a plain `NavLink` calling `navigate(-1)` with no disabled path. I also checked the react-router claim in the installed 7.18.2: `activeRef.current = true` is set in a layout effect with no cleanup, so the `if (!activeRef.current) return` guard passes after unmount and the navigate goes through to the live `HashRouter` history.

Fixed in bcd2289f with the first option rather than the smaller one: the worker is held in a ref and terminated in an unmount cleanup, and both callbacks return early on a mounted flag. Terminating discards the queued message, so the rotation cannot land from the stale closure; the mounted flag is the belt to that braces.

The back link stays enabled — cancelling mid-scrypt now genuinely cancels instead of committing.

Same file, unrelated to unmount: the worker is also terminated once it has delivered, and `unlock-vault` got the same treatment. It was spawning two workers per submit and terminating neither, including the unlock worker that a wrong password never posts to.

No automated coverage for either page — jest here is `testEnvironment: 'node'` with no jsdom. Verified by reading. Worth a manual pass: start a password change, press back during the ~1-2 s of scrypt, then confirm the old password still unlocks.

src/libs/crypto/hashing.ts
ost-ptk: Verified the way you described it: reverting line 64 to `return passwordHash === digest;` left `npx jest src/libs/crypto/hashing.test.ts` at 11/11 green. The helper really was unwireable without anything turning red.

Added your assertion in 7e74959f, as the case it actually pins:

```
it('routes through the constant-time comparison, not ===: an upper-case hash still verifies', ...)
```

Re-ran the same mutation with it in place — 1 failed, 11 passed. So the call site is now held.

It also pins the hex case-insensitivity as you noted, which was previously unstated anywhere. The timing property itself stays unenforceable here for the reason in the description — mutating the loop to `break` early does not change the output, only the timing — so the in-code comment is still the only guard on that loop.

Comp0te (approved):

PR #1473, opened 2026-08-17
fix(redux): make the popup broadcast boundary one-directional and explicit (WALLET-1385)

ost-ptk:
Part 1 of 2 for [WALLET-1385](https://make-software.atlassian.net/browse/WALLET-1385).

The background service worker owns the wallet's only real redux store and broadcasts a `PopupState` to every extension page, which rebuilds a read-only replica from it. This PR makes that boundary **one-directional and explicitly defined**. It does not yet change what is broadcast — narrowing the decrypted vault out of it is PR 2, and this is the groundwork it needs.

## 1. The password change moves into the background, key material and all

`change-password` read the **replica's** vault, encrypted it in a page worker and dispatched `vaultCipherCreated` back, so a page-held copy of the vault became the wallet's only stored cipher. Once the broadcast stops carrying secrets, that same code would encrypt a sanitized vault over the real one and lose the secret phrase and every private key.

The page now sends one `changePassword({ currentPassword, password })` action and nothing else; `changePasswordSaga` does the whole re-key against the background's own state. `create-password-worker.ts` is deleted with its last caller.

**The caller no longer chooses the encryption key.** `changePassword` rides the sender-ungated forwarded path (`redux-actions.ts`, generic branch), so while the payload carried `newEncryptionKeyHash` an extension page could name the key the background re-encrypts the vault under, then read the result back through `fetchPrivateState` — which admits any sender under `runtime.getURL('')` and returns `vaultCipher` — and decrypt it offline. The saga now derives the salts and both hashes itself, from a plaintext password, in the same sequence `initKeysSage` already uses for onboarding.

**And it verifies the current password itself.** Page-side verification alone does not gate that path: `password-protection-page` checks against the `passwordHash` it reads from that same `fetchPrivateState`, so a page that can read the hash can replay the check. The plaintext current password the saga now demands is not derivable from anything an extension page can read. `scryptAsync` yields to the event loop between blocks, so the three derivations do not wedge the worker.

The saga **fails closed** at every step:

- the payload is shape-guarded — and the new password checked non-empty — **inside** the `try`. `{ type: 'CHANGE_PASSWORD_SAGA' }` with no payload reaches `store.dispatch` unchanged on the forwarded path, and a destructure above the `try` would throw past every catch here into `rootSaga`'s boundary-less `all([...])` — cancelling every watcher in the tree, auto-lock and vault persistence included, while `keep-alive.ts` pins the worker alive. It is a `sagaError` instead;
- it refuses when the vault is locked, before burning three scrypt derivations. `lockVaultSaga` has emptied the vault by then, and `encryptionKeyHashCreated` does not clear `isLocked`, so without the guard the new key would be stored and an **empty** vault persisted over the real cipher;
- it verifies before deriving, so a wrong current password costs nothing and changes nothing;
- it encrypts **before** storing anything, and re-checks the lock after that await. Storing first and encrypting after would leave new key material over an old-key cipher whenever `encryptVault` throws — a state the next unlock rejects under both the old and the new password;
- it retracts its own previous errors first (`dismissSagaErrorsBySource`), so a failed attempt's banner cannot outlive a successful retry. `appEvents` is append-only and `SagaErrorBanner` is mounted route-independently.

`keysUpdated`, `encryptionKeyHashCreated` and `vaultCipherCreated` are no longer forwardable from a UI page. `handleReduxAction` forwards allow-listed actions without a sender-page check, so while they stayed listed any compromised extension page could overwrite the stored cipher with arbitrary bytes.

## 2. The broadcast has one definition

`POPUP_SLICES` in the new `popup-state.ts` is the single list of what leaves the background; `PopupState` is derived from it, and `selectPopupState` builds an object literal per slice, so a new slice or a new field reaches replicas only by a deliberate edit. `redux/types.d.ts` is deleted and its nine importers moved.

Every override is built field-by-field rather than by spread — including `session`, through a `popupSession` helper mirroring `popupVault`. A spread both satisfies and widens the override type, so a newly added `SessionState` field would be required by the target type and supplied by the spread: it compiles clean and ships to every page. Probed on this repo's own `tsc --strict`: with one new `SessionState` field, the helper form errors `TS2322` and the spread form does not. `popup-state.test.ts` pins the exact key set of `session` alongside `keys` and `windowManagement`.

The selector's output is field-for-field what it was before; `get-main-store.test.ts` pins it and needed no changes.

## 3. The replica is built from an allowlist, not a spread

`createMainStoreReplica` took `<T extends PopupState>` and spread `...state` into `preloadedState`. That bound **widens** rather than narrows — a full `RootState` satisfies it — so a wider payload would have carried `vaultCipher` into every page with no diagnostic at all. `preloadedState` is now copied slice by slice from `POPUP_SLICES`, and a test feeding an over-wide payload proves an unlisted slice cannot reach a replica. (Reverting the implementation to `...state` makes that test fail — checked.)

The allowlist is over slices, not fields: what a **listed** slice carries is `selectPopupState`'s business, and the comment says only that. A third test asserts a listed slice actually arrives, so the copy itself is guarded — replacing its body with `acc[key] = undefined as never` fails it, where the two assertions-by-absence stayed green.

`payloadSeqById` leaves the broadcast with it: no UI reads it, and the replica restores an empty map, which is truthful for a store that tracks no payload write order.

## Known residuals

- **The three puts are not atomic.** Each persists separately, so a service-worker death between them can still leave key material inconsistent with the cipher. The window is storage-write ordering rather than the crypto, and it is far narrower than the page round-trip it replaces, but only a single multi-slice action would close it.
- **A lock in flight races a password change.** If `lockVaultSaga`'s flush (encrypting under the old key) and the new cipher resolve out of order, keys=new can end up over an old-key cipher. Both encrypts are ~2ms and start order strongly predicts resolve order; the pre-existing equivalent window was wider.
- **`unlockVault` is the surviving instance of the caller-chosen-key shape.** It rides the same ungated forwarded path with an `encryptionKeyHash` in its payload. Out of scope here; worth its own ticket.

## Verification

`npm run ci-check` clean (94 suites / 888 tests), `build:chrome` and `build:firefox` both succeed, and the full popup e2e suite passes (52 passed, 1 skipped) — including `settings/change-password.spec.ts`, whose happy path changes the password, opens the secret phrase with the **new** password, locks and unlocks again. That runs the real scrypt path end to end, and is exactly the vault-integrity scenario this PR exists to protect.

[WALLET-1385]: https://make-software.atlassian.net/browse/WALLET-1385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/background/redux/sagas/vault-sagas.ts
```diff
@@ -191,6 +193,87 @@ export function* lockVaultSaga() {
...
+    yield put(
+      sagaError({
+        source: 'changePasswordSaga',
+        message: 'Password was not changed: the wallet locked. Try again.'
+      })
+    );
+    return;
+  }
+
+  const {
```
Comp0te: The destructure here sits above the `try` that opens on line 230, so a `changePassword` action arriving with no `payload` key throws before any of this saga's own error handling can see it. Every other saga in the file — `initKeysSage`, `unlockVaultSaga`, `createAccountSaga` — destructures inside its `try`.

That matters because of what the comment on line 232 already notes: this action rides the sender-ungated forwarded path (`redux-actions.ts:124`, generic branch at :238-241), so a message of just `{ type: 'CHANGE_PASSWORD_SAGA' }` reaches `store.dispatch` unchanged. The resulting `TypeError` escapes into `root-saga.ts`'s bare `all([...])`, which has no error boundary, and `sagaMiddleware.run(rootSaga)` is called with no `onError` — so it cancels every watcher in the tree, not only this one.

To reproduce: against an unlocked session, dispatch `{ type: changePassword.type }` and observe `TypeError: Cannot destructure property 'passwordHash' of 'action.payload' as it is undefined` escaping uncaught with no `sagaError` put; then dispatch `lockVault` and observe it is never handled (`task.isRunning() === false`). Auto-lock (`timeoutCounterSaga`), the debounced `updateVaultCipher` persistence and `lockVaultSaga` all stop with it, while `keep-alive.ts` pins the service worker alive — so the wallet stays unlocked, stops auto-locking and stops persisting vault edits, with no banner, until the worker is restarted by hand.

Moving the destructure and the `isLocked` pre-check inside the `try`, with a shape guard on the payload, turns this into a `sagaError` instead of an escaping throw.

src/background/handlers/redux-actions.ts
```diff
@@ -124,9 +121,9 @@ export const FORWARDED_ACTION_TYPES: ReadonlySet<string> = new Set(
     initVault,
     recoverVault,
     createAccount,
+    changePassword,
```
Comp0te, line 140: `changePassword` joins `FORWARDED_ACTION_TYPES` here, and its payload is the new key material: the sender chooses `passwordHash`, `passwordSaltHash`, `keyDerivationSaltHash` and `newEncryptionKeyHash` (`sagas/actions.ts:29-34`), `changePasswordSaga` passes `newEncryptionKeyHash` straight into `encryptVault` (`vault-sagas.ts:238-242`), and the current password is never verified in the background — that check lives page-side only, in `password-protection-page/index.tsx:102-149`.

To be clear about the direction of travel: this diff is a net improvement here. It removes `keysUpdated`, `encryptionKeyHashCreated` and `vaultCipherCreated` from the forwarded list, where the base version dispatched that trio directly from the page and handed the plaintext vault to a page worker. The parity test's claim about that removal is accurate and this is not a regression.

The narrower point is that the replacement is new code on the same ungated path, and it re-instantiates in one action roughly what the trio granted in three. The generic branch at :238-241 applies no `isTrustedUiSender` gate — the `windowRequestWindowAttached` branch directly above it (:203-230) does — so an extension-page context can dispatch it, then read the re-keyed cipher back through `fetchPrivateState` (`private-state.ts:34-39` admits any sender under `runtime.getURL('')`, and `selectPrivateState` returns `vaultCipher` plus the hashes) and decrypt offline. A partially-filled payload is its own case: `keysUpdated` takes `Partial<Omit<KeysState,'keysDoesExist'>>` spread over state (`keys/reducer.ts:24-28`), so the saga re-emitting `undefined` hashes flips `keysDoesExist` to `false` via `withDerivedFlag` — the wallet then presents as having no password while its cipher has been rewritten.

For fairness on reach: this is not exposed to a web page. The content relay pins `SDK_REQUEST_TYPES` (`content/index.ts:24-32`), and a content script fails `isTrustedUiSender` regardless.

Since the series is closing this class of door, is it worth deriving the key material in the background here — as `initKeysSage` already does — so the caller cannot choose the encryption key, and verifying `passwordHash` inside the saga before re-keying? Gating the forwarded branch alone would not be enough, since a compromised extension page passes that gate. `unlockVault` (:118) is the stronger surviving instance of the same shape, if it is not already on the follow-up list.

src/background/redux/popup-state.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export const selectPopupState = (state: RootState): PopupState => ({
+  keys: {
+    passwordHash: null,
+    passwordSaltHash: null,
+    keyDerivationSaltHash: null,
+    keysDoesExist: state.keys.keysDoesExist
+  },
+  session: { ...state.session, encryptionKeyHash: null },
```
Comp0te: This is the one override in the file built by spread. `keys` above (lines 60-65) is an explicit four-field literal and `vault` goes through the named `popupVault` helper — both fail closed when their slice type grows. `session` does not: because `PopupSliceOverrides['session']` is `Omit<SessionState,'encryptionKeyHash'> & { encryptionKeyHash: null }`, a newly added `SessionState` field is both *required* by the target type and *supplied* by the spread, so it compiles clean and ships.

Checked against this repo's own `tsc --strict` on a probe replicating all three override forms with one new field each: `session` compiles clean, `keys` errors TS2322 ("Property 'newKeysSecret' is missing"), `popupVault` errors TS2741. So the next field added to `SessionState` — the slice that by definition holds session-secret material, and the one this branch is actively adding to — reaches every open popup, onboarding page and dapp-approval window with no compile error and no review signal.

`popup-state.test.ts` does not catch it either: it uses the exact-key idiom for `keys` (`toEqual`) and `windowManagement` (`Object.keys`), but not for `session`.

A four-line `popupSession(session: SessionState): PopupSliceOverrides['session']` mirroring `popupVault`, plus `expect(Object.keys(payload.session).sort()).toEqual([...])` matching the `windowManagement` idiom already in the test file, would close it.

src/background/redux/create-main-store-replica.ts
```diff
@@ -21,7 +21,18 @@ import { PopupState } from './types';
  * private-key export window, `resetVault` calls `storage.local.clear()`, and the vault sagas
  * broadcast SDK events to dapp tabs — once per open replica.
  */
-export function createMainStoreReplica<T extends PopupState>(state: T) {
+export function createMainStoreReplica(state: PopupState) {
+  // The allowlist is over slices, not fields: each listed slice is copied
+  // wholesale, but a payload wider than PopupState would otherwise carry
+  // vaultCipher and the real password hashes into a page.
+  const slices = POPUP_SLICES.reduce(
```
Comp0te, line 29: The trailing clause of the comment just above — "and the real password hashes" — claims more than this reduce delivers. `vaultCipher` it does stop: that is a top-level `RootState` slice absent from `POPUP_SLICES`, so it is dropped here. The password hashes are fields *inside* the `keys` slice, and `keys` is `POPUP_SLICES[0]`, so `acc[key] = state[key]` copies them through whole; `preloadedState` below overrides only `windowManagement` and `vault`.

The comment's first clause already says this ("over slices, not fields"), so it is only the trailing clause that overreaches — but it is the half a reader will carry away, and it names exactly the material this ticket is about.

Concretely, `createMainStoreReplica({ ...selectPopupState(fullState), keys: { passwordHash: 'REAL', passwordSaltHash: 'REAL', keyDerivationSaltHash: 'REAL', keysDoesExist: true } } as unknown as PopupState)` yields a replica whose `getState().keys.passwordHash === 'REAL'`, and the same holds for an injected `session.isLocked: false` or an attacker-chosen `vault.accounts`. It does need that cast — `PopupState` pins those fields to `null` at the type level — so type-correct callers cannot reach it. What makes the cast reachable in practice is `use-subscribe-to-redux.ts:20-28`, which casts `message: unknown` to `BackgroundEvent`, matches on the action type alone, and calls `setPopupState(event.payload)` with no sender check.

Narrowing the comment to the slice-level claim it can support settles it. If the test should hold the line too: the new over-wide-payload test adds only a *top-level* `vaultCipher` key, which is the case the allowlist does handle, so a case that pollutes a **listed** slice and asserts the replica's `keys` comes back nulled would cover the other half.

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -191,6 +193,87 @@ export function* lockVaultSaga() {
   }
 }
 
+// The page derives the new password material off-thread — not because scrypt
+// can't run in an MV3 service worker (it's plain JS and would run there fine),
+// but to keep several seconds of scrypt off the worker's single thread. Either
+// way it never sees the vault: re-encryption happens here, against the
+// background's own state.
+export function* changePasswordSaga(action: ReturnType<typeof changePassword>) {
```
Comp0te: This saga never dispatches `dismissSagaErrorsBySource('changePasswordSaga')` before attempting the change, so a failed attempt's banner survives a successful retry.

The `appEvents` reducer is append-only by design and documents this exact case (`app-events/reducer.ts:48-54` — a producer that can be retriggered by the user leaves its previous failures pinned; "A producer dispatches this before re-attempting"), and `export-keys-window-saga.ts:172-178` follows the convention for its own retriable flow, with the same rationale in its comment.

Nothing else clears it. `selectSagaErrors` returns `state.appEvents.errors` unfiltered and the banner maps over all of them with no source or recency filter; `resetAppEventsDismission` is dispatched from exactly one site (`onboarding-sagas.ts:67`, the full vault-reset path). In particular, the unlock the user must perform between a locked-wallet failure and a retry does not clear it. And `<SagaErrorBanner />` is mounted at `apps/popup/index.tsx:72`, outside `<AppRouter />`, so it renders on every route.

That makes the stale banner the common path rather than an exotic one, because the failure this saga is written to produce — the wallet locking mid-scrypt, lines 209-217 and 248-256 — is precisely the retriable one. (Worth noting `change-password/index.tsx:90` navigates to Home synchronously at submit, before the worker returns, so the user is already on Home when either outcome lands.)

`yield put(dismissSagaErrorsBySource('changePasswordSaga'))` at the top, before the lock check, mirroring `export-keys-window-saga.ts:178`.

src/background/redux/create-main-store-replica.test.ts
```diff
@@ -0,0 +1,31 @@
+import { createMainStoreReplica } from '@background/redux/create-main-store-replica';
+import { PopupState, selectPopupState } from '@background/redux/popup-state';
+import rootReducer from '@background/redux/root-reducer';
+import { RootState } from '@background/redux/store-types';
+
+const fullState = rootReducer(undefined, { type: '@@INIT' }) as RootState;
+
+describe('createMainStoreReplica', () => {
+  it('ignores slices outside POPUP_SLICES even when the payload carries them', () => {
```
Comp0te, line 15: Neither test in this file observes a listed slice actually arriving in the replica, so the allowlist copy itself is unguarded.

This test is assertion-by-absence throughout (`.not.toBe('SECRET-CIPHER')` plus the `JSON.stringify(...).not.toContain(...)` check). The second test's three assertions all read `windowManagement.requests`, `windowManagement.exportKeysWindowId` and `vault.payloadSeqById` — every one of them produced by the explicit `preloadedState` overrides at `create-main-store-replica.ts:44-54`, which are built from the `state` argument directly and never touch the `slices` object the reduce produces.

Confirmed by mutation: replacing the reduce body with `acc[key] = undefined as never` and running `npx jest src/background/redux/create-main-store-replica.test.ts` gives "Tests: 2 passed, 2 total". So a replica built entirely from reducer defaults on every extension page — no accounts to select from, default theme, empty contacts, `session.isLocked` reset — reaches the CI gate green, and is caught only by the e2e suite, which `CLAUDE.md` documents as outside the default gate. The five construction sites in `src/apps/{popup,onboarding,signature-request,connect-to-app,import-account-with-file}/index.tsx` all render from it.

Seeding `fullState` with a distinguishable value in a pass-through slice (one `contacts` entry, or `settings.themeMode`), passing `selectPopupState(thatState)`, and asserting `replica.getState().contacts` equals it would add one positive assertion alongside the two negative ones.

Comp0te (commented):
Read through the broadcast-boundary rework and the change-password move into the background. The direction is right, and the new `popup-state.ts` seam is a real improvement on the `...state` spread it replaces — most of what follows is about the edges of the contract it establishes rather than the contract itself: one override that fails open where its siblings fail closed, a comment that claims field-level filtering the allowlist does not do, and a unit test that stays green if the allowlist stops copying anything. Two more on the saga: a destructure sitting above its `try`, and a missing error dismissal on a user-retriable flow.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in 1303d940. The destructure and the `isLocked` pre-check both moved inside the `try`, behind an `isChangePasswordPayload` shape guard, so `{ type: 'CHANGE_PASSWORD_SAGA' }` throws where this saga's own catch sees it and becomes a `sagaError` instead of escaping into `rootSaga`'s `all([...])`.

The guard also rejects an empty new password — the form enforces a minimum length, but the action does not arrive from the form on this path, and `password: ''` would have re-keyed the vault under `scrypt('')`.

Test: "reports a malformed payload instead of throwing past its own catch" asserts the `sagaError` and that `encryptVault` was never reached, plus "refuses an empty new password".

src/background/redux/popup-state.ts
ost-ptk: Fixed in 3235d3c2. `session` now goes through a `popupSession(session: SessionState): PopupSliceOverrides['session']` helper, built field by field like `popupVault`.

Reproduced your probe against this repo's `tsc` before and after: with one new `SessionState` field, the helper form errors `TS2322` and the spread form compiles clean. `popup-state.test.ts` also pins `Object.keys(payload.session).sort()`, matching the `windowManagement` idiom already in the file.

src/background/redux/create-main-store-replica.ts
ost-ptk: Fixed in 3235d3c2 — narrowed the comment rather than widening the code, which is the resolution you land on. It now reads that each listed slice is copied wholesale, so what a listed slice carries is `selectPopupState`'s business, and that what the reduce stops is an unlisted top-level slice, `vaultCipher` above all.

Did not add the "pollutes a listed slice, asserts `keys` comes back nulled" case, because that would assert a property the reduce does not have — `keys` is nulled by `selectPopupState`, and `popup-state.test.ts` already pins that. Making the replica null it too would be a second, independent place to keep in sync with the slice's shape. What the replica does own is the slice-level line, and that is now the only thing the comment claims and the only thing the tests assert.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in 1303d940. `yield put(dismissSagaErrorsBySource('changePasswordSaga'))` is now the saga's first statement, ahead of the lock check, mirroring `export-keys-window-saga.ts:178`.

Test: "retracts what a previous attempt reported before re-attempting".

src/background/redux/create-main-store-replica.test.ts
ost-ptk: Fixed in 3235d3c2. Added "copies the listed slices through to the replica": it seeds `fullState` with one `contacts` entry, passes it through `selectPopupState`, and asserts the replica's `contacts.contacts` equals it.

Confirmed with your mutation — `acc[key] = undefined as never` now gives 1 failed / 2 passed where it previously gave 2 passed / 2 total.

src/background/handlers/redux-actions.ts
ost-ptk: Taken, and done in 1303d940 — both halves.

**The caller no longer names the key.** `changePassword`'s payload is now `{ currentPassword, password }`. The saga derives `passwordSaltHash`, `passwordHash`, `keyDerivationSaltHash` and the encryption key itself, in the same sequence `initKeysSage` uses, so there is nothing left in the payload for a sender to choose. `create-password-worker.ts` is deleted with its last caller. The reason it lived on a page was to keep scrypt off the worker's single thread, but `scryptAsync` yields to the event loop between blocks, and onboarding already accepts exactly this — the full popup e2e suite (52 passed, 1 skipped) runs the real path.

**And the current password is verified in the background**, against the stored `keys.passwordHash` / `passwordSaltHash`, before any derivation. Worth spelling out why the page-side check does not substitute: `password-protection-page` verifies against the `passwordHash` it reads from `fetchPrivateState`, so a page that can read the hash can replay the check. The plaintext current password the saga now demands is not derivable from anything an extension page can read, which is what makes this a gate rather than a formality.

Also guarded: an empty new password is rejected with the shape guard, so `password: ''` cannot re-key the vault under `scrypt('')`.

**On `keysUpdated` taking a `Partial`** — the `undefined`-hashes case you describe is now unreachable from this saga: all three values are produced locally and are `string` by construction.

**Not done here, and on the follow-up list:** `unlockVault`, which rides the same ungated path with `newEncryptionKeyHash` and `newVaultCipher` in its payload and verifies nothing. It is the stronger surviving instance, as you say, and it is listed in the PR description's residuals rather than folded into this diff.

**Not done, deliberately:** gating the forwarded branch on `isTrustedUiSender`. As you note, a compromised extension page passes that gate, and a content script already fails it, so it would not change this threat model. Metering a failed verify through `loginRetryCount` is likewise skipped: the same attacker tier already reads `passwordHash` from `fetchPrivateState` and can brute-force offline at the same cost per guess, in parallel, without the background's involvement.

Comp0te (changes requested):
Conflicts

Comp0te (approved):

PR #1474, opened 2026-08-17
feat(vault): serve vault secrets on demand instead of broadcasting them (WALLET-1385)

ost-ptk:
Part 2 of 2 for [WALLET-1385](https://make-software.atlassian.net/browse/WALLET-1385). **Stacked on #1473** — see "Dependency on #1473" below; the dependency is semantic, not just textual.

The background service worker broadcasts a `PopupState` to every extension page on every store change while unlocked. It carried the **decrypted secret phrase and every account's private key**, delivered to popup, onboarding, connect-to-app, signature-request and import-account-with-file alike — including the pages that never sign anything. This PR serves both on demand instead, and then removes them from the broadcast.

## The mechanism

`src/background/handlers/vault-secrets.ts` follows the `private-state.ts` precedent shipped in #1429, with one addition: **per-page authorization**.

| Request | Answered to | Returns |
| --- | --- | --- |
| `SECRET_PHRASE_REQUEST` | `popup.html` | the phrase |
| `ACCOUNT_SECRET_KEYS_REQUEST` | `popup.html`, `signature-request.html` | keys for the named accounts that hold one |
| `SUGGESTED_ACCOUNT_NAME_REQUEST` | `popup.html` | the next default account name |

`isTrustedUiSender` proves "an extension page", not "a page that needs this" — without the pathname allowlist, an XSS in connect-to-app (which renders dapp-controlled origin, favicon and site name) could simply ask for the phrase. Every request also refuses outright when the vault is locked, answers `null` rather than staying silent (a silent refusal leaves a legitimate page's promise pending forever), and builds its response map with `Object.create(null)`, since the keys are user-chosen account names.

## What changed per consumer

- **Backup phrase** fetches after the password gate, not at page mount, and drops the phrase when the wallet locks — a JS string cannot be zeroized, so dwell time is the only lever.
- **Create account** no longer holds the phrase to guess a default name. That loop derived key pairs in the UI and was duplicated verbatim in `createAccountSaga`; it is now one background helper that derives once per probed index instead of once per (index, account) pair.
- **Signing** — the four signature-request pages, transfer, staking, NFT transfer, the Ledger window, key export and mobile-sync QR fetch the key at the moment of use, into a local variable that never enters React state or a dependency array.
- `fetchAccountSecretKey` **never rejects**: a transport failure resolves to an empty key, so every caller lands in the refusal path it already implements. A rejection escaping into a signing page would have left the window open and the dapp's promise pending forever.
- Key export and QR treat a refused response as an error. Previously-shaped code would have produced an empty zip behind a Success screen, and a valid-looking QR that synced imported accounts to the phone **without their keys**.

## The blanking

`selectPopupState` now emits `secretPhrase: null` and every `secretKey` as the literal `''` — typed as that literal, so forgetting to blank one is a compile error rather than a silent leak.

Watch-only accounts were previously identified *by* their empty secret key, which blanking would have made indistinguishable from every other account. The sanitizer therefore emits an explicit `watching` flag. It is `secretKey === '' && hardware == null`: Ledger accounts also carry an empty key, and folding them in would have made the flag lie about its own name for the next consumer.

## Dependency on #1473

The blanking is only safe because the base branch moved the password change into the background. `change-password` used to re-encrypt the **replica's** vault and write the result back as the stored cipher — with that path still open, a sanitized vault would have been encrypted over the real one and the phrase and every private key lost at the next unlock.

The base has since changed shape in review, and further in this branch's favour: `changePassword` now carries `{ currentPassword, password }`, and the saga verifies the current password and derives all the key material itself, so nothing about the re-key comes from a page any more. This branch is merged with that base (faa7d4bc) and re-validated against it below. The one conflict was in `popup-state.ts`, where the base added a field-by-field `popupSession` helper immediately above this branch's `popupVault`; both are kept, and `session` is now built the same way `vault` is here.

## Threat model — what this does and does not buy

The page allowlist is **hardening, not a boundary**. The background serves the phrase to any request attributed to `popup.html` while the wallet is unlocked, with no password re-verification — the password screens are client-side. Code execution inside an allow-listed extension page can still ask.

What the PR removes is the standing exposure: secrets are no longer structured-cloned into five page contexts on every store change, and pages that never sign (connect-to-app, onboarding, import-account-with-file) can no longer receive them at all.

Also worth knowing for review:

- Worst case ~16 s (3 attempts × 5 s timeout + backoff) before a failure surfaces; two pages render blank during it.
- A transport failure reaches the dapp as `SIGNING_ACCOUNT_MISSING`. It settles the promise and closes the window, but the wording names the wrong cause. Distinguishing it needs a discriminated result across eleven call sites — deliberately not done here.
- Replica `secretKey` is now always `''`. `PopupState` pins that, but selectors typed against `RootState` erase the literal, so nothing stops future UI code from reading the field and quietly getting an empty string.
- Deliberately unchanged: `jsonById` / `eip712ById` still ride the broadcast. They are dapp-originated payloads the signing screens must render, not vault key material.

## Two commits that are not about vault secrets

Both are here because they unblock this branch, not because they belong to the feature:

- **`nosemgrep` on the rejected-sender warn.** `cw-logging-secrets` is a regex over the console call's argument text, so the word `vault` inside the message string tripped it and failed the blocking Semgrep step. The only value logged is the sender origin. Two suppressions of this same rule already exist in the tree.
- **`.claude/worktrees/` excluded from jest.** A git worktree checked out inside the repo is a full checkout with its own `node_modules`, and jest's `testRegex` sweeps `rootDir` — a local run collected 459 foreign suites on top of this branch's own and had ts-jest compile six copies of the codebase. The exclusion is anchored to `<rootDir>`, since an unanchored pattern also matches a worktree's own path and excludes everything. Both paths are absent on a CI runner, so the patterns are inert there.

## Verification

Re-run on the merge with the updated base, not on the pre-merge branch:

`npm run ci-check` clean (96 suites / 918 tests, coverage floors met, knip clean). Full popup e2e green **after** the blanking — 52 passed, 1 skipped — covering signature-request scenarios, transfers, staking, backup-secret-phrase, change-password and create-account. `build:chrome` and `build:firefox` both succeed.

`change-password` is the spec worth naming here: it changes the password, then opens the secret phrase with the **new** one, then locks and unlocks. On this branch that phrase arrives over `SECRET_PHRASE_REQUEST` rather than from the replica, and the re-encryption it is checked against runs entirely in the background — so it exercises both halves of the stack at once.

[WALLET-1385]: https://make-software.atlassian.net/browse/WALLET-1385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/background/redux/popup-state.ts
```diff
@@ -95,10 +113,17 @@ function popupSession(session: SessionState): PopupSliceOverrides['session'] {
...
   return {
-    secretPhrase: vault.secretPhrase,
-    accounts: vault.accounts,
+    secretPhrase: null,
+    accounts: vault.accounts.map(({ secretKey, ...account }) => ({
+      ...account,
+      // The literal type is the enforcement: forgetting to blank a key is a
+      // compile error. `watching` means watch-only specifically — a Ledger
+      // account also has an empty secretKey but is excluded via `hardware`.
+      secretKey: '' as const,
```
Comp0te, line 124: This blanks `secretKey` on every broadcast account, but `selectVaultAccountsAvailableForExport` filters on `Boolean(account.secretKey)`, and its only production call site reads the popup replica. The Download-account-keys window lists zero accounts and its button stays disabled — the wallet's only in-product key export, unreachable for every user.

Filter on broadcast facts instead, or build the list from what `fetchAccountSecretKeys` returns.

- `vault/selectors.ts:34-37` filters `Boolean(account.secretKey)`; `download-account-keys/download.tsx:43` is its only production call site.
- The window is `popup.html#/download-account-keys` (`export-keys-window-saga.ts:28`), and the popup store is the replica (`apps/popup/index.tsx:50` → `create-main-store-replica.ts:24`). Nothing repopulates `secretKey` there, so `download.tsx:52-60` renders no rows and `index.tsx:107` is `disabled={!selectedAccounts.length}`.
- Green because `vault/selectors.test.ts:7-28` feeds the selector background-shaped accounts with real keys, and nothing under `e2e-tests/` opens the window.
- `sign-eip712/index.tsx:142` was migrated to `watching` in this PR; this selector is the one `secretKey` check left behind.
- `!account.watching && account.hardware == null` is the closest equivalent, but `watching` alone excludes Ledger by construction, so a naive swap changes which accounts are offered.

</details>

src/apps/popup/pages/backup-secret-phrase/index.tsx
```diff
@@ -31,6 +37,14 @@ export function BackupSecretPhrasePage() {
     );
   }
 
+  if (error) {
+    return <PrivateStateErrorPage layout="popup" onRetry={retry} />;
+  }
+
+  if (secretPhrase == null) {
```
Comp0te, line 50: Returning `null` here renders a blank popup — no header, no spinner — while the phrase fetch is in flight, worst case 15.75s against a degraded background. `create-account/index.tsx:132-134` does the same. Before this PR both read the replica synchronously, so no delay was possible.

Expose `isLoading` and render the loading affordance, as `wallet-qr-code` already does.

- 15.75s = `request-with-retry.ts:1` `FETCH_TIMEOUT_MS = 5000` × 3 attempts, plus `RETRY_DELAYS_MS = [250, 500]`.
- Genuinely empty: both routes are bare elements (`app-router.tsx:337-340`, `:245-248`), each page owns its own `PopupLayout`, and only `SagaErrorBanner` mounts outside the router. `Suspense` at `:228` covers lazy chunks only — `lazy-page-fallback.tsx:5-12` says so.
- No flag to render: `use-secret-phrase.ts:10-14,57` returns `{ secretPhrase, error, retry }`; `create-account/index.tsx:103-126` exposes `suggestedName`/`error`/`retry`. The pattern is already in this diff at `password-protection-page/index.tsx:37,218`, passed from `wallet-qr-code/index.tsx:139`.
- `password-protection-page/index.tsx:185-186` already returns `null` while private state loads, so blank-while-fetching is a house pattern — but that wait isn't behind `requestWithRetry`.

</details>

src/apps/popup/pages/wallet-qr-code/index.tsx
```diff
@@ -40,29 +46,56 @@ export const WalletQrCodePage = () => {
...
+    }
+
+    // The worker sends only derived accounts' names — the mobile client re-derives their keys from the phrase.
+    const secretKeys = await requestWithRetry(() =>
+      fetchAccountSecretKeys(importedAccounts.map(a => a.name))
+    ).catch(() => null);
+
+    // A null/failed response must not turn into a partial payload: that would
+    // silently sync imported accounts to the phone without their keys.
+    if (!secretKeys) {
```
Comp0te: This guard catches `null` — a path `strict` already pins. Unpinned is the empty map: `{}` passes the guard and yields `secretKey: ''` for each account, reaching the success screen with a QR that syncs keyless accounts. The page has no test at all.

Extract the fetch/refuse/assemble decision into a plain module and unit-test the `{}` case.

- `secretKeys` is `Record<string, string> | null` (`vault-secrets.ts:151-158`) and `tsconfig.json:9` sets `"strict": true`, so deleting this guard fails `tsc` at `:81` — the `null` half is pinned by the type system. `{}` is not: it satisfies the guard, and `:64` can produce it when every imported account is watch-only.
- No coverage of any kind: `WalletQrCodePage|generateQRCode|wallet-qr-code` matches only `router/paths.ts`, `app-router.tsx`, and the page's own two files. `workers.test.ts:53-66` drives the worker with empty account lists and a forced `scryptAsync` rejection, never the page's payload assembly.
- The export half has this — `run-keys-download.test.ts:88-99`, "routes a refused secret-keys request to Failure, not an empty zip".

</details>

Comp0te (commented):
Read through the on-demand vault-secrets path — the broadcast shape in `popup-state.ts`, the new handler and its clients, and the pages that moved from reading the replica synchronously to fetching. The comments below share one through-line: the keys left the replica, and a few readers still assume either that they are still there, or that what comes back over the message boundary is well-formed.

src/background/redux/popup-state.ts
ost-ptk: Confirmed — the window listed zero accounts and the button never enabled. Fixed in 6d5a1195.

`selectVaultAccountsAvailableForExport` now filters on `!account.watching` instead of `Boolean(account.secretKey)`. The naive-swap hazard you flagged doesn't bite here: `selectVaultAccountsExceptLedgersAccounts` already drops `hardware` upstream, so `watching` only has to carry the watch-only half.

The test was the other half of the bug — it fed background-shaped accounts with real keys, which is exactly why the blanking didn't break it. It now uses the replica shape the selector actually runs against, so a revert to a `secretKey` check fails it.

src/apps/popup/pages/backup-secret-phrase/index.tsx
ost-ptk: Confirmed. Fixed in 703074db.

Both pages now render `PrivateStateLoadingPage` instead of `null` — a spinner under the same header as the page it precedes, so the frame doesn't shift when the data lands. It sits beside `PrivateStateErrorPage` in `@libs/layout`, which these two routes already render on failure. `useSecretPhrase` exposes `isLoading` rather than letting the page infer it from a null phrase: the hook also nulls the phrase on lock, and inferring would spin forever there.

One correction to the basis, since it changes the conclusion: `password-protection-page/index.tsx:185` is *not* outside `requestWithRetry` — `use-private-state.ts:24` returns `requestWithRetry(fetchPrivateState)`, so that blank render carries the same ~15.75s worst case. Left as is here to keep this commit to the two pages you flagged; it needs a `layout` prop on the loading page, since that page also renders in the window and tab variants. Happy to fold it in if you'd rather it ship together.

src/apps/popup/pages/wallet-qr-code/index.tsx
ost-ptk: Dug into this one, and the `{}` framing doesn't hold — but there is a real gap next to it.

`{}` is not reachable as a defect. `selectVaultImportedAccounts` filters on `imported`, which covers real imported accounts and watch-only ones (`add-watch-account/index.tsx:46`) but not Ledger: `connected-ledger.tsx:119-124` sets `hardware` without `imported`, and the `accountsImported` reducer stores the payload verbatim. The handler omits exactly the accounts with `secretKey === ''`, so `{}` means "every imported account is watch-only" — and `secretKey: ''` is then the truth, not a loss.

What is genuinely missing is the completeness invariant `run-keys-download.ts:46-55` enforces and this path doesn't. The map is keyed by account name and matched by exact equality against the live vault, while the names come from the replica snapshot at click time. A rename or a removal in another window between the last broadcast and the request drops a real key into `?? ''`, behind a Success screen and a QR the phone accepts. `watching` is the discriminator that separates that from a legitimately keyless account — the same one you pointed at in the export selector above.

Note also that `forSync` runs over `derivedAccounts` too, whose names are never in the map by construction (the worker reads only `da.name`), so any check has to exclude them.

Agreed on the extraction, and it's the only way to test this at all: `testEnvironment` is `node` with no jsdom, and the worker's `new URL(..., import.meta.url)` won't compile under ts-jest — so the fetch/refuse/assemble half has to come out of the component while orchestration stays in. Landing as a separate commit on this branch, mirroring `run-keys-download.test.ts`, with `{}`/all-watch-only pinned as legitimate and a non-`watching` imported account missing from the map pinned as an error.

ost-ptk: Landed in 1a4c7f11.

`buildQrSyncPayload` now owns the fetch/refuse/assemble decision and requires a key for every imported account the broadcast did not mark `watching`, logging a count rather than a name. Derived accounts no longer consult the map at all — their names are never in it by construction, so the shared `?? ''` was carrying two different meanings in one expression.

Worker construction stayed in the component, which is what makes the rest testable. Eleven tests, the page's first: the two anchors are the all-watch-only empty map pinned as legitimate, and a non-`watching` imported account missing from the map pinned as an error.

Return type is `null` rather than a discriminated result, deliberately — the page renders the same `PrivateStateErrorPage` with the same retry for every failure, so a `reason` field would have no consumer today. Easy to widen if one appears.

src/apps/popup/pages/backup-secret-phrase/index.tsx
ost-ptk: Folded `password-protection-page` in after all — 3019e65c.

`PrivateStateLoadingPage` now takes an optional header so it can carry that page's own: the dedicated-window variant swaps the back link for a close action, and the default header would have offered a back link that traps the user in a single-entry history.

Comp0te (approved):

Issue #10, opened 2022-01-20
As a Signer user I can import an existing account


ghost:
Figma: https://www.figma.com/file/mBlqs6l6I2d3ihX80Mswm4/Casper-Signer?node-id=115%3A933

**Requirements**
This ticket will enable legacy Signer users to import their account using a private key .PEM file in Signer 2.
1. Logged into their vault, a user will have an option to Import Account.
![image](https://user-images.githubusercontent.com/97249036/166694889-2c5021cc-2fcf-43ea-94bd-8ff7424f1e89.png)
2. The user will click 'Import with a file' and be prompted to upload their Private Key file. If the user uploads an unsupported file format, an error will display 'Please upload a .PEM containing your private key.'
![image](https://user-images.githubusercontent.com/97249036/166695195-e3c1e542-e7e3-446b-a3a4-205801ba4b63.png)
3. Once a valid key file is uploaded, a user is directed to a new screen with the filename displayed. A text input field allows the user to name their account. A max of 23 characters is allowed, including spaces. A back button allows the user to move back to the upload screen.
![image](https://user-images.githubusercontent.com/97249036/166695587-5ef7133d-c747-4d6f-8b45-61b0cd670afc.png)
4. Once a valid account name is included in the input field the user will be able to click the Import button to finish importing
![image](https://user-images.githubusercontent.com/97249036/166695904-ba857aec-ed8c-452c-8d09-d4fac4161d99.png)
5. Once imported, a confirmation screen appears confirming the successful import. By clicking Thanks the user is taken to the home screen with their new account set as the Active account. If user imports multiple accounts back to back, display them in a list on the home screen by the account name.
![image](https://user-images.githubusercontent.com/97249036/166696415-a94cd242-c057-4f53-bda8-89262ffd8822.png)

**Out of Scope in this ticket**
Importing account with Secret Phrase (will be added in as part of https://github.com/make-software/casper-signer-v2/issues/1

PR #1476, opened 2026-08-18
fix: WALLET-1394 — keep the approval window out of reuse during a Ledger confirmation

ost-ptk:
Fixes [WALLET-1394](https://make-software.atlassian.net/browse/WALLET-1394).

## The bug

An **already-permitted** Ledger device opens no permission window, so the device call runs in the page inside the **shared approval window**. When the next dapp request arrives, that window is reused — and reuse runs `tabs.update(tab.id, { url })` on its tab (`create-open-window.ts:100-110`), a full navigation. `ledger` is a **per-document** singleton (`libs/services/ledger/ledger.ts`), so the navigation destroys the document, the HID session and the pending `signTransaction` promise together.

Confirmed on hardware, 3 runs. The ticket describes a second mechanism — the signature reaching `sdk-response-to-tab` and being dropped by the supersede tombstone — and that one is real too, but it happened in only 1 of 3 runs. **Which one wins is a race** with how fast the replacement page commits its navigation (bundle cache, disk, load). Both end in `signResponse({ cancelled: true })` to the first dapp.

This matters for the fix: nothing on the tombstone side can help, including WALLET-1364 P2 #10 — in the dominant case there is no longer anything to deliver. Only preventing the reuse works.

PR #1427 / #1462 covered only the _other_ path, where a separate permission window gives the request a second display and the reuse leaves it alive.

## The fix

- `awaitingDeviceConfirmation` on the **open** `Request` descriptor. Deliberately not a slice-level set: the tombstone replaces the whole entry and `windowDetachedFromRequests` shrinks `windowIds`, so the flag cannot outlive what it describes and nothing has to remember to clear it.
- `selectIsWindowBusyWithDevice(state, windowId)` composes the answer from the request side.
- `openWindow` withholds the tracked window while it is busy. Passing `null` **redirects** the reuse rather than suppressing it: the new-window branch still runs `setWindowId` (`create-open-window.ts:154`), so the shared slot retracks to the newcomer and only the protected window leaves the rotation.
- UI half: `runWithDeviceConfirmationReported(requestId, run)` brackets the device call. The bracket lives in its own module for the reason `register-ledger-permission-window.ts` does — there is no React-hook harness in this repo, and a start without its end would withhold the shared window for the rest of that request's life. It also gives the previously un-awaited `ledgerAction()` a `catch`.
- The action is **intercepted** in `handleReduxAction`, not added to `FORWARDED_ACTION_TYPES` (which checks no sender at all). Gates: `isTrustedUiSender` plus `payload.requestId === recoverRequestId(sender.url)` — the same pair `closeLedgerFlowWindows` uses, so a page can only speak for the request its own URL names.

## Reviewer notes

**Deliberate behaviour change in the permission-window flow.** During the device call the request's `windowIds` is `[shared, permission]`, so both now read as busy and a second request opens a third window instead of reusing the shared one. Previously the reuse navigated the shared window and the request merely _survived_ it. Strictly safer — the waiting screen is no longer navigated away — but it is a change, and it is pinned by a selector test.

**Residual race, knowingly left open.** The flag is dispatched from the page and reduced in the background one `runtime.sendMessage` hop later. A second dapp request processed inside that hop (single-digit ms) still reuses the window. Previously the exposure was the entire device confirmation — seconds to minutes.

It was **not** closed by awaiting the dispatch ack, though the ack does mean the flag is in the store (`background/index.ts:216` responds only when the result carries a `response` key, and the branch dispatches synchronously before returning). The drop paths return `{ handled: true }` with **no** `response` key, so their promise never resolves — awaiting it before the device call would mean that any future tightening of the gate silently stops Ledger signing altogether. Bad trade against a millisecond window. Closing it properly means moving "this request is Ledger-signed" into the background, which it could derive from `signingPublicKeyHex` plus the vault at window-open time; that is a wider behaviour change and belongs in its own ticket.

**A burst opens one window per request while the device is busy.** The slot is only rewritten in the `windows.create` callback, so several requests arriving during a Ledger confirmation all see "tracked but busy" and each opens its own window. This is not new machinery — the same thing happens on base whenever no window is tracked yet (`selectWindowId` is `null` at cold start, so concurrent calls all take the create branch and each calls `setWindowId`); what changes is the condition that triggers it. Base collapsed such a burst into one window only by cancelling each previous request in turn, which for a Ledger confirmation is the exact P0 this PR exists to stop. Bounded near `MAX_STORED_PAYLOADS` for `sign`; `signMessage` has no equivalent cap, which is a pre-existing gap of its own.

**One line is covered only by the manual test:** the wiring at `use-ledger.ts:135`. Reverting it to a bare `ledgerAction()` leaves the whole suite green — same gap `registerLedgerPermissionWindow` lives with, for the same reason.

## Testing

`tsc`, `lint`, `format:check`, `knip` clean; 1047 unit tests pass; the coverage gate passes (`windowManagement` 100%, `handlers` above its floor).

Manual, on a real Ledger with the HID permission already granted (so no permission window appears — that is the precondition for this bug):

1. Dapp A → `sign` from a Ledger account, stop at "Confirm on device".
2. Dapp B (second tab) → `signMessage`.
3. **Before:** A's window is repainted with B's request; A's dapp receives `cancelled: true` ~250 ms later; pressing Approve on the device does nothing. **After:** A's window is untouched, B opens its own window, A signs through, then B signs through.

## Found while testing — not fixed here

Two Ledger flows can now be live at once, which was unreachable before this change because the second request used to consume the first one's window. If the user leaves A waiting on the device and presses **Sign with Ledger** in B, B's document opens the device and sends its own APDUs while A's prompt is up. The Casper app has no queue: it tears down A's prompt and answers B with `0x6985`, so B shows "You rejected to sign the deploy" (which the user did not do) and A can no longer be completed — the window has to be closed and the flow restarted.

This is contention over one physical device, not window management, and it is **not** made worse by this PR: previously A was lost _unconditionally and silently_, whereas now it is lost only if the user deliberately starts a second signing flow, and the failure is visible. Note that `LedgerEventStatus.WaitingToSignPrevDeploy` already exists with exactly the right copy for this situation — B just never gets to it.

It cannot be arbitrated inside a page (the `ledger` singleton is per document); the only shared authority is the background store, which is where this PR's flag already lives. It also affects the internal flows (transfer, staking, import-from-ledger), which carry no `requestId` and so are not covered by that flag.

Investigated and filed as **[WALLET-1429](https://make-software.atlassian.net/browse/WALLET-1429)**. Two findings from it are worth knowing while reviewing this PR:

- **Request A hangs rather than fails.** There is no timeout anywhere on the signing path — `TransportWebHID` never consults `exchangeTimeout`, and `exchangeAtomicImpl`'s 15 s `unresponsiveTimeout` only emits an event. So `runWithDeviceConfirmationReported`'s `finally` never runs and `awaitingDeviceConfirmation` stays set. The blast radius is still what this PR describes — the slot retracks to the newcomer's window and later requests reuse that normally, so it costs exactly one held window until A's window is closed, not one per request.
- **`LedgerEventStatus.WaitingToSignPrevDeploy` is unreachable**, and not only because of ordering: its trigger `returnCode === 65535` (`ledger.ts:135`) cannot fire for a busy device, since Zondax's `processErrorResponse` returns the real status code whenever the error carries one and `0xffff` only as the fallback for errors without one. Any future fix leaning on that copy must trigger it from a wallet-held fact.

[WALLET-1394]: https://make-software.atlassian.net/browse/WALLET-1394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WALLET-1429]: https://make-software.atlassian.net/browse/WALLET-1429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/background/open-window.ts
+++ b/src/background/open-window.ts
@@ -13,7 +13,10 @@ import {
   windowIdChanged,
   windowIdCleared
 } from '@background/redux/windowManagement/actions';
-import { selectWindowId } from '@background/redux/windowManagement/selectors';
+import {
+  selectIsWindowBusyWithDevice,
+  selectWindowId
+} from '@background/redux/windowManagement/selectors';
 
 export interface OpenApprovalWindowProps extends OpenWindowProps {
   /**
@@ -52,8 +55,21 @@ export function openWindow(
     );
   };
 
+  // Withheld while a Ledger confirmation is in flight in it — reuse navigates
+  // that window's tab out from under the device call (see
+  // `awaitingDeviceConfirmation` in windowManagement/types). `null` REDIRECTS
+  // the reuse rather than suppressing it: the new-window branch still runs
+  // `setWindowId`, so the slot retracks and only the protected window leaves
+  // the rotation. WALLET-1394.
+  const trackedWindowId = selectWindowId(store.getState());
+  const reusableWindowId =
+    trackedWindowId != null &&
+    selectIsWindowBusyWithDevice(store.getState(), trackedWindowId)
+      ? null
+      : trackedWindowId;
+
   const chain = createOpenWindow({
-    windowId: selectWindowId(store.getState()),
+    windowId: reusableWindowId,
     setWindowId: (id: number) => store.dispatch(windowIdChanged(id)),
     clearWindowId: () => store.dispatch(windowIdCleared())
   })(openWindowProps).then(
--- a/src/hooks/ledger-device-confirmation.ts
+++ b/src/hooks/ledger-device-confirmation.ts
@@ -2,6 +2,47 @@ import { dispatchToMainStore } from '@background/redux/utils';
 import { windowRequestDeviceConfirmationChanged } from '@background/redux/windowManagement/actions';
 
 /**
+ * How many brackets currently hold each request, so overlapping ones report the
+ * flag once between them rather than each releasing it for the others.
+ *
+ * They do overlap: neither signing page disables its submit control while a
+ * call is in flight, and the page state that hides it is only flipped after
+ * `getPreferredTransport()` and `beforeLedgerActionCb()` resolve — so a second
+ * click lands, fails fast on the busy transport (`TransportRaceCondition`) and
+ * would otherwise unprotect the window while the first call is still on the
+ * device.
+ *
+ * Module scope is the right scope: a transport is per document, so brackets can
+ * only overlap within one, and the reducer's equality guard already absorbs a
+ * repeated `true` from anywhere else.
+ */
+const heldByRequest = new Map<string, number>();
+
+/** @returns whether this is the first holder, i.e. whether to report the start. */
+function acquire(requestId: string): boolean {
+  const held = (heldByRequest.get(requestId) ?? 0) + 1;
+  heldByRequest.set(requestId, held);
+
+  return held === 1;
+}
+
+/** @returns whether this was the last holder, i.e. whether to report the end. */
+function release(requestId: string): boolean {
+  const held = (heldByRequest.get(requestId) ?? 1) - 1;
+
+  if (held > 0) {
+    heldByRequest.set(requestId, held);
+    return false;
+  }
+
+  // Deleted rather than left at zero: `requestId` is dapp-controlled, and a page
+  // that signs repeatedly would otherwise grow this map for its whole life.
+  heldByRequest.delete(requestId);
+
+  return true;
+}
+
+/**
  * Runs a Ledger device call with the background told, for its whole duration,
  * that this request is on the device — which is what keeps the window it runs
  * in out of the reuse rotation (`awaitingDeviceConfirmation` in
@@ -25,6 +66,10 @@ export async function runWithDeviceConfirmationReported(
   // dapp request behind them, so there is no descriptor to flag.
   const report = (awaiting: boolean) => {
     if (requestId == null || requestId === '') {
+      return;
+    }
+
+    if (!(awaiting ? acquire(requestId) : release(requestId))) {
       return;
     }
 
--- a/src/hooks/ledger-device-confirmation.ts
+++ b/src/hooks/ledger-device-confirmation.ts
@@ -1,0 +1,50 @@
+import { dispatchToMainStore } from '@background/redux/utils';
+import { windowRequestDeviceConfirmationChanged } from '@background/redux/windowManagement/actions';
+
+/**
+ * Runs a Ledger device call with the background told, for its whole duration,
+ * that this request is on the device — which is what keeps the window it runs
+ * in out of the reuse rotation (`awaitingDeviceConfirmation` in
+ * windowManagement/types).
+ *
+ * The bracket lives here rather than as two calls in `useLedger` for the reason
+ * `register-ledger-permission-window.ts` exists: the repo has no React-hook
+ * harness, so inside the hook the pairing would be two lines that can drift
+ * apart — and a start without its end withholds the shared window from every
+ * later request for the rest of that request's life.
+ *
+ * Never rejects. `run` is invoked fire-and-forget by the hook, exactly as the
+ * bare `ledgerAction()` it replaces was; a rejection propagated from here would
+ * be the unhandled rejection that call site already produced.
+ */
+export async function runWithDeviceConfirmationReported(
+  requestId: string | undefined,
+  run: () => Promise<void>
+): Promise<void> {
+  // The internal flows (transfer, staking, `import-account-from-ledger`) have no
+  // dapp request behind them, so there is no descriptor to flag.
+  const report = (awaiting: boolean) => {
+    if (requestId == null || requestId === '') {
+      return;
+    }
+
+    dispatchToMainStore(
+      windowRequestDeviceConfirmationChanged({ requestId, awaiting })
+    );
+  };
+
+  report(true);
+
+  try {
+    await run();
+  } catch (error) {
+    // Name only: a Ledger error carries the public key and transaction hash it
+    // failed on, and this line is a diagnostic, not the user-facing surface —
+    // the same failure already reaches the screen through the event subject.
+    console.error('useLedger: the device action failed', {
+      errorName: (error as Error)?.name
+    });
+  } finally {
+    report(false);
+  }
+}
--- a/src/background/handlers/close-ledger-flow-windows.test.ts
+++ b/src/background/handlers/close-ledger-flow-windows.test.ts
@@ -32,7 +32,8 @@ const openRequest = (windowIds: number[]): Request => ({
   tabId: 1,
   origin: 'https://dapp.example',
   method: 'sign',
-  windowIds
+  windowIds,
+  awaitingDeviceConfirmation: false
 });
 
 const removedIds = () => removeMock.mock.calls.map(([id]) => id).sort();
--- a/src/background/handlers/redux-actions.parity.test.ts
+++ b/src/background/handlers/redux-actions.parity.test.ts
@@ -142,6 +142,12 @@ const EXCLUSIONS: ReadonlySet<string> = new Set(
     // request whose set can never shrink to the window that went away is a
     // request nothing can ever cancel.
     windowManagementActions.windowRequestWindowAttached,
+    // UI-dispatched (use-ledger, around the device call), but intercepted by its
+    // own branch in handleReduxAction — deliberately not in the forwarding set,
+    // which checks no sender at all. It decides whether the shared approval
+    // window may be reused, so a page must only be able to set it on the request
+    // its own URL names.
+    windowManagementActions.windowRequestDeviceConfirmationChanged,
     // UI-dispatched (use-ledger, when a Ledger flow ends), but intercepted by
     // the dedicated `closeLedgerFlowWindows` branch in handleReduxAction —
     // deliberately not in the forwarding set. It has no reducer case at all, so
--- a/src/background/handlers/redux-actions.test.ts
+++ b/src/background/handlers/redux-actions.test.ts
@@ -7,7 +7,10 @@ import { MainStore } from '@background/redux/get-main-store';
 import { closeLedgerFlowWindows } from '@background/redux/ledger/actions';
 import { lockVault, resetVault } from '@background/redux/sagas/actions';
 import { accountRenamed } from '@background/redux/vault/actions';
-import { windowRequestWindowAttached } from '@background/redux/windowManagement/actions';
+import {
+  windowRequestDeviceConfirmationChanged,
+  windowRequestWindowAttached
+} from '@background/redux/windowManagement/actions';
 
 import { handleCloseLedgerFlowWindows } from './close-ledger-flow-windows';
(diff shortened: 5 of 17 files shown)


src/hooks/ledger-device-confirmation.ts
```diff
@@ -0,0 +1,50 @@
...
+  run: () => Promise<void>
+): Promise<void> {
+  // The internal flows (transfer, staking, `import-account-from-ledger`) have no
+  // dapp request behind them, so there is no descriptor to flag.
+  const report = (awaiting: boolean) => {
+    if (requestId == null || requestId === '') {
+      return;
+    }
+
+    dispatchToMainStore(
```
Comp0te, line 76: This dispatch is discarded, and its action type is not in `SURFACED_DISPATCH_ACTIONS` — so a dropped message produces no banner, no warning and no retry. A lost `awaiting: true` silently reopens the race this PR closes.

Add the type to the set — or, since that set's criterion is *visible* results and this failure has none, a `console.warn` here?

- `ledger-device-confirmation.ts:31-33` — the promise from `dispatchToMainStore` is neither returned, awaited, nor voided.
- `redux/utils.ts:23-54` — it never rejects (the `.catch` at :37 returns `false`), and `reportUiError` fires only for members of the set (:49-51). A dropped dispatch leaves only the background `console.error` at :45.
- `surfaced-dispatch-actions.ts:24-56` — 11 members, this type not among them; both siblings serving the same Ledger-window purpose are: `windowRequestWindowAttached` (:32) and `ledgerNewWindowIdChanged` (:37, whose comment notes neither has a retry path). The criterion is at :14-20.
- The other direction is bounded: a lost `awaiting: false` leaves the flag stuck true, but the redirect at `open-window.ts:65-69` still retracks the slot through `setWindowId` (`create-open-window.ts:154-155`), so it costs one extra window and strands the protected one for the rest of the request's life — not one window per later request. Nothing else clears the flag: one reducer case (`reducer.ts:77-101`), and no cancel or close handler touches it.

</details>

src/hooks/ledger-device-confirmation.ts
```diff
@@ -0,0 +1,50 @@
...
+    await run();
+  } catch (error) {
+    // Name only: a Ledger error carries the public key and transaction hash it
+    // failed on, and this line is a diagnostic, not the user-facing surface —
+    // the same failure already reaches the screen through the event subject.
+    console.error('useLedger: the device action failed', {
+      errorName: (error as Error)?.name
+    });
+  } finally {
+    report(false);
```
Comp0te, line 93: This `finally` clears `awaitingDeviceConfirmation` for every concurrent bracket — the flag is a bare boolean with no holder. A second Sign click stays possible until `getPreferredTransport()` and `beforeLedgerActionCb()` resolve — narrow, but not closed — and whichever bracket settles first releases the window while the other is still on the device.

A holder token, or a held-count, would fix it.

- `sign-transaction/index.tsx:462-467` — the `disabled` expression covers `transaction`, `signatureRequest`, loading and wasm-approval only, nothing in-flight; `sign-message/index.tsx:259-263` has no `disabled` prop at all.
- `use-ledger.ts:120-141` — the ref test at :120-122 reads `selectedTransportRef` but never claims it before awaiting `getPreferredTransport()`; the state flip that hides the button is `await beforeLedgerActionCb()` at :132. No lock or ref guard runs before those awaits.
- `reducer.ts:83-89` applies `awaiting: false` to any live open request unconditionally. The second bracket's `report(true)` no-ops on the equality guard (:86), and its `report(false)` clears the flag — a fast-failing second call still runs its `finally`.
- Both signing pages are unchanged here: the un-disabled window is pre-existing, but the window-release consequence exists only because of this PR.

</details>

src/background/open-window.ts
```diff
@@ -52,8 +55,21 @@ export function openWindow(
...
+  // that window's tab out from under the device call (see
+  // `awaitingDeviceConfirmation` in windowManagement/types). `null` REDIRECTS
+  // the reuse rather than suppressing it: the new-window branch still runs
+  // `setWindowId`, so the slot retracks and only the protected window leaves
+  // the rotation. WALLET-1394.
+  const trackedWindowId = selectWindowId(store.getState());
+  const reusableWindowId =
+    trackedWindowId != null &&
+    selectIsWindowBusyWithDevice(store.getState(), trackedWindowId)
+      ? null
```
Comp0te, line 68: This `null` redirects every concurrent request, not just one: `windowId` is rewritten only in the `windows.create` callback, so a burst arriving while the slot is tracked-and-busy all take the new-window branch — several approval popups while the user is on their Ledger, where base collapsed them into one.

If that is unacceptable, claim the slot synchronously at the reuse decision.

- `create-open-window.ts:154-155` (reached via `open-window.ts:73`) is the only non-test writer of `windowIdChanged`; nothing claims the slot synchronously. At base, `open-window.ts:56` passed the tracked id unconditionally, so the same burst reused one window.
- How many: for `signRequest` the count is bounded near `MAX_STORED_PAYLOADS = 10` — `sdk-methods.ts:243-262` refuses with `tooManyPendingRequests` and returns before `openWindow` once `storePayload` refuses (`vault/reducer.ts:117-122`, cap at :38). `signMessageRequest` (`sdk-methods.ts:284-318`) stores no payload and has no equivalent cap.

</details>

Comp0te (commented):
Read this against the WALLET-1394 failure it closes, along with the reducer, selector and `openWindow` tests it adds. The fix is sound and the tests pin it. The comments below all sit on one seam: `awaitingDeviceConfirmation` is a single boolean whose truth has to survive a message hop, a chain of browser round-trips, and any concurrent holder — each comment is one place where it does not.

src/hooks/ledger-device-confirmation.ts
ost-ptk: Fixed in 6d7fc413 — this one was right and reachable.

The overlap window you describe is real: `use-ledger.ts:120-132` awaits `getPreferredTransport()` and then `beforeLedgerActionCb()` before anything hides the control, so a second click lands, fails fast on `TransportRaceCondition`, and its `finally` unprotected the window while the first call was still on the device.

Went with the held-count. `heldByRequest` counts holders per `requestId`; the start is reported only by the first and the end only by the last. Module scope is deliberate and I think sufficient: a transport is per document, so brackets can only overlap within one, and a repeated `true` arriving from anywhere else is already absorbed by the reducer equality guard. The count is deleted rather than left at zero, since `requestId` is dapp-controlled and a page that signs repeatedly would otherwise grow the map for its whole life.

Three tests cover it: overlap does not release early, separate requestIds are counted separately, and — the failure a counter could introduce — the count is released so a later call reports again.

src/hooks/ledger-device-confirmation.ts
ost-ptk: Pushing back on this one, with a correction to the premise.

A dropped dispatch is not silent today: `dispatchToMainStore` logs `Dispatch to Main Store failed: <action.type>` with the error at `redux/utils.ts:45` on every failure, before the `SURFACED_DISPATCH_ACTIONS` check at :49. Your own basis notes that line, so I think the disagreement is only over whether a log is enough here rather than over what happens. Adding a `console.warn` in this module would print the same failure twice.

On the set itself — its criterion is at `surfaced-dispatch-actions.ts:14-20`: the dispatch IS the sole source of the visible result, so a drop means the user sees nothing happen. This one has no visible result at the moment it is dropped. Both siblings you cite do: without `windowRequestWindowAttached` the request can never be cancelled at all, and without `ledgerNewWindowIdChanged` the permission window is never closed on success. Ours only matters if a second dapp request happens to arrive afterwards.

The stronger reason not to surface it is where the banner would land. `SagaErrorBanner` is mounted over the approval screens, so this one would render on top of the Ledger confirmation prompt — the exact placement `cancel-requests.ts:135-140` already argues against for the recovered-cancel case, and at the one moment we least want to distract someone reading a transaction off their device.

I did consider closing the hole rather than reporting it, by awaiting the dispatch ack before starting the device call — `background/index.ts:216` responds only when the handler result carries a `response` key, and this branch dispatches synchronously before returning, so the ack really does mean the flag is in the store. I rejected it: the drop paths return `{ handled: true }` with no `response` key, so their promise never resolves, and awaiting it would mean any future tightening of the sender gate silently stops Ledger signing altogether. Bad trade against a one-hop window.

Happy to add the type to the set if you still disagree — but I would rather leave the banner off the signing prompt and keep the existing log as the record.

src/background/open-window.ts
ost-ptk: Correct about the burst, and it is worth stating in the PR — but I do not think it is new machinery, so I would rather document it than add a synchronous claim here.

The unclaimed-slot burst already exists on base whenever no window is tracked: `open-window.ts:56` passes `selectWindowId(...)`, which is `null` at cold start, so concurrent calls all take the `openNewWindow` branch and each one calls `setWindowId` (`create-open-window.ts:154`). One window per request, same as here. What this PR changes is the condition that produces it — from "no window tracked yet" to "tracked but on the device" — not the absence of a claim.

It is also self-limiting in a way the base case is not: the burst has to arrive during a Ledger confirmation, and every window it opens is a real pending approval the user has to answer anyway. Base collapsed them into one window, but only by cancelling each previous request as it went — which for a Ledger confirmation is precisely the P0 this PR exists to stop.

Claiming the slot synchronously is the right shape if we do act, but it is not free: there is no id to claim before `windows.create` resolves, so it would mean a background-local "an open is in flight" guard with its own failure mode — if the create then rejects, the guard has to be unwound or every later request reuses a window that never existed. That is a new invariant for a case bounded near `MAX_STORED_PAYLOADS` for `sign`, and I would rather not add it under this fix.

`signMessageRequest` having no equivalent cap is a fair catch and genuinely separate from this PR — it is an unbounded-pending-requests gap on its own. Want me to file it, or do you read it as in scope here?

Adding the burst behaviour to the PR description now so it is not a surprise at merge.

src/hooks/ledger-device-confirmation.ts
Comp0te: You're right on the criterion, and that settles it. `surfaced-dispatch-actions.ts:14-20` scopes the set to dispatches that are the sole source of a visible result — a dropped flag here has no effect until a later `openWindow` reads it, so the omission is conformant, not an oversight. The banner placement seals it. Withdrawing; the log at `redux/utils.ts:45` is the right record.

src/background/open-window.ts
Comp0te: Base comparison holds — `open-window.ts:56` passed the tracked id unconditionally, so an untracked slot already produced one window per request; this PR changes the condition, not the claim. Documenting it is enough for me; the synchronous claim brings its own unwind path. On `signMessageRequest`: separate — file it, it predates this PR and shouldn't ride on a Ledger fix.

Comp0te (approved):

PR #1484, opened 2026-08-19
fix(background): cross-check the SDK response delivery target

ost-ptk:
## Description

The SDK response — which can carry `signatureHex`, `encryptedMessage` or a decrypted plaintext — was delivered to the `tabId` carried in the UI message, with no cross-check against the request descriptor that recorded who actually asked (`src/background/handlers/sdk-response-to-tab.ts`).

Two ways that leaks:

1. **The tab navigated away.** Tab ids survive navigation and the content script is injected on every http(s) origin, so once the user opens an approval and the original tab goes somewhere else, the new document is a live receiver for a signature it never requested. A page can take the private SDK port itself — the content script hands it over with `window.postMessage(…, [port2])`, and every listener in the page world sees `event.ports[0]` — so the SDK's own `requestId` filter only protects an honest page.
2. **Every frame of the tab.** All three manifests declare `"all_frames": true`, so a bare `tabs.sendMessage(tabId, …)` is delivered to every frame, third-party iframes included — no navigation involved at all.

### What changed

`handleSdkResponseToTab` now:

- reads the request descriptor **before** the optimistic `markRequestResponded` (that call writes the tombstone `{ status: 'responded' }`, which drops `tabId` / `origin` / `frameId`) and refuses the direct send when the message names a tab that is not the requesting one;
- verifies the tab still hosts the requesting origin (`tabs.get`) before handing over the response. Gated to top-frame requests: `tabs.get` reports the **top** document's url, so a sub-frame request's own origin is not comparable to it and would refuse every legitimate iframe-embedded dapp;
- targets `tabs.sendMessage` at the `frameId` now recorded on the descriptor (`sender.frameId`, all six approval flows);
- narrows the same-origin fallback to frame 0 when the request came from the top frame — frame ids are per-tab, so a sub-frame id names a different frame, or none, in whatever other tab the origin match selected.

### Notes for review

- **Ordering is load-bearing.** The tab comparison is synchronous and runs before the mark; the origin check awaits, so it runs after it. Any await between the dedupe status read and `markRequestResponded` reopens the double-delivery race that file already documents.
- **Fails closed.** Tab gone, no url, opaque origin, or no descriptor *and* no `?origin=` on the approval window → the response is withheld and the existing delivery-failed banner is surfaced.
- **No new surface.** No new user-facing copy, no i18n key, no manifest or permission change (`tabs` was already declared). Logs carry identifiers and origins only — never the payload.
- After an MV3 service-worker restart the descriptor is gone, so the origin is recovered from the approval window's `?origin=` query param rather than failing closed; that path keeps working.

### Residual risks, named deliberately

- **TOCTOU** — the tab can still navigate between the origin read and the send. Unavoidable with this API; the window shrinks from the whole approval to one event-loop turn.
- **Sub-frame navigation** — an iframe that made the request and then navigates keeps its `frameId` and would still receive the response. Closing that needs the `webNavigation` permission, deliberately out of scope.
- **Terminal refusal** — a withheld response leaves the request `'responded'`, so no retry can deliver. Consistent with the pre-existing "delivery failure is terminal" decision.
- **iframe dapp + service-worker restart** — with no descriptor the frame is unknown, so the gate compares the top document's origin and an iframe-embedded dapp is refused outright. Rare (keep-alive holds the worker while the vault is unlocked), fails closed, and fixable later by threading `frameId` through the approval-window query string.

### Testing

- Unit: 703 tests green across `src/background` (61 suites), `handlers` coverage 97.13 / 91.41 / 100 / 97.09 against the 85 / 100 / 95 / 95 gate. `tab-origin.ts` and `deliver-via-origin.ts` at 100%.
- Manual, Chrome:
  - control — sign from a top-level dapp resolves with a signature;
  - **navigate away mid-approval** — response withheld, background logs `target tab no longer hosts the requesting origin; response withheld` with `delivered: 0`, wallet shows the delivery-failed banner, and a probe page on the new origin that grabs the SDK port receives nothing;
  - **sub-frame dapp** (frame origin ≠ top document origin) still receives its signature — the frame-scoping regression gate.

### Out of scope, noticed while testing

Not caused by this branch, both in paths it does not touch — worth their own tickets:

- `requestConnection()` called from a **sub-frame** resolves `true` while `isConnected()` stays `false` (reproduced twice); the same call at top level both resolves `true` and flips `isConnected()` to `true`. A dapp in an iframe can believe it is connected when the wallet says it is not.
- Related, already documented in code as KNOWN NOISE: `select-account/index.tsx` sends `connectResponse(true)` from the render body with no guard.

## Linked tickets

[WALLET-1386](https://make-software.atlassian.net/browse/WALLET-1386)

## Checklist

- [x] If the PR adds any new text to the UI, make sure they are localized — no new UI text; the existing `sagaError` surface is reused

- [x] Include a screenshot or recording if implementing significant UI or user flow change — no UI change

[WALLET-1386]: https://make-software.atlassian.net/browse/WALLET-1386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/background/handlers/deliver-via-origin.ts
```diff
@@ -11,11 +11,21 @@ import { SdkMethod } from '@content/sdk-method';
...
+  action: SdkMethod,
+  frameId?: number
 ): Promise<number> {
   if (!origin) return 0;
+
+  // Frame ids are per-tab, so a sub-frame's id names a different frame — or no
+  // frame — in whatever tab the origin match selected. Only frame 0 is the same
+  // frame everywhere, and the origin match already established that this tab's
+  // top document belongs to the dapp.
+  const targetFrameId = frameId === 0 ? 0 : undefined;
```
Comp0te: For a sub-frame request this passes `undefined`, so the fallback reaches every frame of each matched tab. Unscoped at the base too — left open, not opened. Same-origin iframe → `signatureHex` reaches sibling frames; cross-origin → an unrelated tab gets it, `:220-222` tombstones the request, and the asker never does.

Restrict the fallback to the originating tab, or drop it and surface the loss?

- `utils.ts:70-72` turns an `undefined` frame into the two-argument `tabs.sendMessage(tab.id, action)`, which under `all_frames: true` (`manifest.v3.json:50`) reaches every frame.
- The match at `utils.ts:67` is `getUrlOrigin(tab.url) === origin` over active tabs — the **top** document. A sub-frame's recorded origin is its own (`sdk-methods.ts:102` → `:138` → `sdk-response-to-tab.ts:183-184`), so same-origin iframes match their own tab and cross-origin ones match unrelated tabs.
- `content/index.ts:41-74` posts any `signResponse` onto the receiving frame's port; the requestId check happens later, in the page world at `content/sdk.ts:103`.
- Reached with a positive `frameId` from `:204` and `:305`, not `:266` — `:262` gates that one to `frameId == null || frameId === 0`.

</details>

src/background/handlers/sdk-methods.ts
```diff
@@ -135,6 +135,7 @@ export async function handleSdkMethod(
         windowRequestOpened({
           requestId: action.meta.requestId,
           tabId: senderTabId,
+          frameId: sender.frameId,
```
Comp0te, line 138: All six `windowRequestOpened` dispatches record `frameId: sender.frameId`, and `0` — the top-frame value, and the common case — never reaches them in any test. `sdk-methods.test.ts` uses one shared `SENDER` with `frameId: 3` and pins `3` in all six assertions.

Add a `frameId: 0` sender variant asserting the dispatched payload carries `frameId: 0`.

`0` is falsy, so the usual `sender.frameId || undefined` slip erases exactly the top-frame case. Change all six sites and run `jest sdk-methods.test.ts sdk-response-to-tab.test.ts windowManagement/reducer.test.ts` — 123/123 still pass, with every top-frame descriptor now carrying no frame and `sdk-response-to-tab.ts:297-299` back on the unscoped two-argument send.

At head the only `frameId: 0` values are `utils.test.ts:210` (passed straight to the util) and `sdk-response-to-tab.test.ts:829/835/848` (hand-built descriptors) — never one that travelled from a sender.

</details>

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -200,14 +241,68 @@ export async function handleSdkResponseToTab(
...
+          tabId,
+          expectedOrigin,
+          liveOrigin,
+          type: action?.type,
+          delivered
+        }
+      );
+
+      store.dispatch(deliveryFailedError(tabId, delivered > 0));
+      void closeLedgerWindowsAfterResponse(store, displays);
```
Comp0te, line 282: Three exits of this handler's new logic are unpinned — a mutation to each leaves both suites green. The one anchored here: the withhold branch runs after the optimistic `markRequestResponded`, so `closeLedgerWindowsAfterResponse` at `:282` is the only teardown, and no test reaches it with a Ledger flow.

Three cases, one pass over the two suites.

<details><summary>The three</summary>

**`sdk-response-to-tab.ts:282` — Ledger teardown.** `makeStore`'s `ledgerWindowId` defaults to `null` (`sdk-response-to-tab.test.ts:83`) and no call site passes it, so `isLedgerFlow` is always false; `close-windows-on-response.test.ts` fixes `tabs.get` once at `:122-124` to a matching origin, so it never enters the branch. Delete the line — 59/59 still pass. A user whose tab navigated mid-approval is left with a signature window and a `type: 'normal'` permission window (`create-open-window.ts:187-200`) that its own Close button cannot remove.

**`sdk-response-to-tab.ts:183` — origin precedence.** `request?.origin ?? recoverDappOrigin(sender.url)` is what makes the descriptor authoritative and the page's `?origin=` a fallback. Every descriptor in both suites carries `https://dapp.example`, and the only origin-carrying sender (`sdk-response-to-tab.test.ts:65-68`) carries the same value, so the two sources are never distinguishable. Invert it to `recoverDappOrigin(sender.url) ?? request?.origin ?? null` — 59/59 still pass. The param is not attacker-nominated (`isTrustedUiSender`, and only `sdk.bundle.js` is web-accessible), but it can be stale or belong to another request, and it steers both the comparison at `:265` and the broadcast target at `:266`/`:305`.

**`sdk-response-to-tab.test.ts:342` — the no-origin refusal.** The base's `'invalid tabId, no origin → …'` case asserted `emitToOriginMock` was never called. `:183` now prefers the descriptor, so that same setup takes the delivering path, and the replacement does not restore the assertion. Repo-wide, `emitToOriginMock).not` has zero hits — nothing anywhere pins "no fallback attempted".

</details>

Comp0te (commented):
Read the delivery path end to end — the `frameId` recorded on the descriptor, the live-origin cross-check, and the same-origin fallback — against the base. The through-line of these comments is the fallback: it is the one path the new frame scoping does not reach, and the branches around it are the ones the suites do not pin.

src/background/handlers/deliver-via-origin.ts
ost-ptk: Dropped it, rather than restricting to the originating tab — c58a1d76.

Restricting to the originating tab is incoherent for both callers that can reach here with a positive `frameId`: at `:305` the direct send to that tab has just failed, and at `:204` we are refusing that tab on purpose. So a non-zero `frameId` now returns `0` without broadcasting, and the caller surfaces the existing "not delivered" error — the loss is visible instead of silently landing in the wrong document.

Two cases stay as they were: `frameId === 0` still forwards (same frame in every tab, and the origin match already established that tab's top document is the dapp), and `frameId === undefined` — no descriptor, the MV3-restart path — keeps broadcasting unscoped.

Pinned by `the same-origin fallback is never attempted for a sub-frame request`: asserts `emitToOriginMock` is not called and the sagaError says not delivered. Reverting the guard to the old `targetFrameId = frameId === 0 ? 0 : undefined` fails it.

src/background/handlers/sdk-methods.ts
ost-ptk: Added in c58a1d76: `SENDER_TOP_FRAME` (`frameId: 0`) and `a top-frame sender (frameId 0) reaches the dispatch as frameId 0`, asserting the dispatched `windowRequestOpened` payload carries `frameId: 0`.

Verified it discriminates: with `sender.frameId || undefined` at the six sites the new test fails while the rest of the suite stays green, which is the slip you described. `0` is the expectation that does the work — jest's argument equality ignores an `undefined` value but not a `0`, which is also why the existing six assertions on `frameId: 3` could not catch it.

src/background/handlers/sdk-response-to-tab.ts
ost-ptk: All three pinned in c58a1d76, each checked by applying the mutation you named and watching the new test fail alone.

**Ledger teardown (`:282`).** New test in `close-windows-on-response.test.ts` — a descriptor whose `windowIds` are `[10, 11]` with the ledger slot on `11`, `tabs.get` overridden per-test to a different origin so the withhold branch is actually entered, and the message tab matching the descriptor so it does not divert into the tabId-mismatch branch. Asserts both windows are removed and `tabs.sendMessage` was never called. Deleting the teardown line now fails it.

**Origin precedence (`:183`).** New test with the two sources deliberately different — descriptor on `https://dapp.example`, sender url carrying `?origin=https://impostor.example` — asserting the fallback is invoked with the descriptor's origin. Inverting the `??` now fails it.

**No-fallback refusal.** Restored as an assertion on the existing `withholds the response when no origin can be established` case: `expect(emitToOriginMock).not.toHaveBeenCalled()`. That is the only occurrence in the repo again, and removing `if (!origin) return 0` fails it.

src/background/handlers/sdk-methods.ts
Comp0te: Confirmed, and it discriminates. But the assertion pins `method: 'sign'` — one of the six recording sites. The other five have no frameId-0 sender anywhere in the tree, so the same slip at `:305` or `:410` un-scopes `signatureHex` / `encryptedMessage` with a green suite.

One more case parameterised over the six, or a deliberate leave-it?

The six: `:138` connect, `:176` switchAccount, `:270` sign, `:305` signMessage, `:374` signTypedData, `:410` decryptMessage. `SENDER_TOP_FRAME` is declared at `sdk-methods.test.ts:132` and used at exactly one call site, `:521`, inside a `signRequest` case asserting `method: 'sign'` — so `:270` is pinned and the other five are not. `:305` and `:410` are the two whose responses carry payload rather than a status.

</details>

src/background/handlers/sdk-response-to-tab.ts
```diff
@@ -163,19 +167,56 @@ export async function handleSdkResponseToTab(
...
-    const delivered = await deliverViaOrigin(origin, action);
+  if (!validTab || !tabIdMatchesRequest) {
+    // No usable tab, or not the tab that made the request. Attempt the
+    // same-origin fallback (broadcast to any active tab of the dapp origin).
+    // Only mark responded when the fallback ACTUALLY delivered to at least one
+    // tab — we ARE delivering, so a later duplicate must dedupe. When nothing
+    // was delivered (no origin, or the broadcast matched zero tabs), do NOT
+    // mark responded: a valid retry must still be able to deliver. Either way
+    // surface the non-fatal error.
+    const delivered = await deliverViaOrigin(expectedOrigin, action, frameId);
```
Comp0te, line 204: Dropping the third argument here leaves both suites green — every test reaching this branch has no `frameId`. It's defence-in-depth (reaching it needs a bad `tabId` from our own UI), but the mutation broadcasts a sub-frame response unscoped, and `delivered > 0` at `:220-222` tombstones it, so the asking frame can't retry.

A sibling of the `:884` case pins it.

The other two call sites are covered: `:266` is gated by `:262` on `frameId == null || frameId === 0`, and `:305` is pinned twice — `sdk-response-to-tab.test.ts:871-882` on the 3-arg shape, `:884-897` on `emitToOriginMock` not being called for `frameId: 4`.

`:204` is not. The third parameter is optional (`deliver-via-origin.ts:16`), so the 2-arg call compiles, and the tests reaching this branch already assert the 2-arg shape. `OPEN_REQUEST` (`:72-79`) carries no `frameId` and `openWith` in `close-windows-on-response.test.ts` (`:95-107`) dispatches without one; the `frameId: 4` descriptors (`:834`, `:843`, `:888`) all use a matching `tabId`, so they reach the direct send or `:305`.

The guard is *bypassed*, not removed: `frameId` is undefined inside the function, so `:27` doesn't fire and `:36-38` takes the unscoped branch into `utils.ts:70-72`.

Suggested: `makeStore({ ...OPEN_REQUEST, frameId: 4 })` with `makeMessage(3)`, asserting `emitToOriginMock` not called.

</details>

src/background/redux/windowManagement/reducer.ts
```diff
@@ -63,6 +64,7 @@ const slice = createSlice({
           [action.payload.requestId]: {
             status: 'open',
             tabId: action.payload.tabId,
+            frameId: action.payload.frameId,
```
Comp0te, line 67: `frameId: action.payload.frameId || undefined` here keeps the suite green. `0` is pinned at both ends of the chain — the sender→action hop and the emit util — but not at this hop, the action→descriptor write. Every top-frame request would then record no frame, and `sdk-response-to-tab.ts:297-299` falls to the unscoped `tabs.sendMessage(tabId, action)`.

A `reducer.test.ts` case dispatching `frameId: 0` pins it.

What the mutation survives: `reducer.test.ts:134-155` uses `frameId: 4` (`4 || undefined === 4`), `:157-169` omits it, and the helper at `:121-132` asserts with `toEqual`, which ignores `undefined`-valued keys. `sdk-methods.test.ts:512-536` stops at the action — `makeStore` (`:112-121`) returns `dispatch: jest.fn()`. Nothing in the tree dispatches `windowRequestOpened` with `frameId: 0`.

The pinned ends are `sdk-methods.test.ts:512-536` for sender→action and `utils.test.ts:202-212` for the emit util — the latter calls `emitSdkEventToActiveTabsWithOrigin` directly, so it can't observe this line.

Suggested: dispatch `windowRequestOpened({ ..., frameId: 0 })` and assert the whole descriptor with `toEqual`, so an erased `0` can't pass as an ignored `undefined`.

</details>

Comp0te (commented):
Second pass over the pushed fixes — the three pinned exits and the top-frame sender assertion all hold up under the mutations they name. What's left is the same shape one hop out: `frameId: 0` is pinned at the sender hop and at the emit util, but neither of the two hops in between — the reducer that writes the descriptor, and the tab-mismatch call site — would catch a falsy-vs-nullish slip.

src/background/handlers/sdk-methods.ts
ost-ptk: All six, in 9a633c1a — `SENDER_TOP_FRAME` now reaches connect, switchAccount, sign, signMessage, signTypedData and decryptMessage, each asserting `frameId: 0` on the dispatched payload.

The existing six assertions on `frameId: 3` stay as they are: a `frameId: 0` expectation alone would also pass against a hard-coded `0`, so the two together are what pin "the value travels, and it is not erased".

One test per flow rather than a loop — each case builds its own method-specific payload and its own `openWindow` assertion, which is the shape the rest of that file already has.

src/background/handlers/sdk-response-to-tab.ts
ost-ptk: Pinned in 9a633c1a, with the setup you suggested: `makeStore({ ...OPEN_REQUEST, frameId: 4 })` and `makeMessage(3)` against the descriptor's tab `7`, so it lands on the tab-mismatch branch rather than the direct send, asserting `emitToOriginMock` was not called.

Dropping the third argument fails it: `frameId` is `undefined` inside `deliverViaOrigin`, the sub-frame guard does not fire, and the unscoped branch calls the emit once.

src/background/redux/windowManagement/reducer.ts
ost-ptk: Pinned in 9a633c1a: `windowRequestOpened records frameId 0 rather than erasing it` dispatches `frameId: 0` and asserts the whole descriptor with `toEqual` — including `awaitingDeviceConfirmation: false`, so the assertion fails on the erased frame rather than on a missing sibling key.

`action.payload.frameId || undefined` fails it, while the neighbouring `frameId: 4` case stays green — which is the asymmetry that let this hop through.

Comp0te (approved):

PR #1496, opened 2026-08-20
fix(vault): arm the login-retry lockout in the background

ost-ptk:

### The bug

The login-retry lockout is a UI-only control armed by a message that can vanish.

`useLockWalletWhenNoMoreRetries` was the only thing that ever set `loginRetryLockoutTime`, and it did
so with a fire-and-forget `dispatchToMainStore`. If that message was dropped, background state kept
`null`, nothing armed the lockout, and nothing noticed — not the code, not the user. The screen that
enforces the lockout renders from the replica, so the whole control lived in state a lost message
could desynchronise.

### The fix

The background now arms the lockout itself, in `armLockoutSaga`, on every retry increment. The
increment and the arming happen in the same process, so there is no message in between to lose. The
saga also locks a still-unlocked vault — the case the deleted hook covered — and reconciles on resume,
for a persisted count that passed the limit while nothing armed it.

### Note on the ticket's suggested direction

The ticket proposed refusing an unlock in `unlockVaultSaga` while a lockout is active. That check is
included, but **it is defence in depth, not the fix**: if `loginRetryLockoutTimeSet` never reached the
background, the selector is `null` and the check passes. That is why the enforcement had to move
rather than being guarded.

### Behaviour change to review carefully

The password-field reset is re-keyed to the lockout transition. The background now arms the lockout in
the same dispatch that increments the counter, so a replica may never observe the old condition
(`!hasLockout && retriesLeft <= 0`) — left as it was, the reset would silently stop firing.

`loginRetryLockoutTimeSet` leaves `FORWARDED_ACTION_TYPES` (with its `EXCLUSIONS` entry, required by
the parity test in the same commit): no UI page dispatches it any more, so no page needs to.

### Testing

`armLockoutSaga` unit tests (below limit, at limit, already armed, unlocked vault, stale count on
resume) and the `unlockVaultSaga` gate. Existing `unlockVaultSaga` cases updated for the new selector.

### Follow-ups from review

**The onboarding tree a locked vault lands in.** Arming the lockout in the background means five wrong
passwords on an onboarding tab lock the vault, which cleared `encryptionKeyDoesExist` and put
`WelcomePage` — one unconfirmed click from `resetVault()` — in front of a keyed wallet. Routing that to
`unlock` closed the wipe but landed it in a tree that cannot succeed: the re-auth form only proves the
password, and onboarding is deliberately off the unlock allowlist (#1498).

`unlock` was doing two jobs and is now split into `reauth` (this tab's session is stale, the vault is
open) and `locked` (the vault is shut). `locked` renders `VaultLockedPage` — no password field, the
checkbox-gated reset as its only action. Abandoned onboarding lands there too: the same reset they got
at base, now behind a confirmation.

`initKeysSage`'s two puts are reordered as part of this. The store subscriber broadcasts after every
dispatch, so `keys && !session` really does reach the tab and would flash the locked screen over the
create-password page; `session && !keys` reads as a fresh install, whose routes still include that page.

**The refusal path stranding the unlock page.** `unlockVaultSaga` refuses with a `sagaError` and a bare
return, which left the spinner running, submit at `pointerEvents: 'none'` and Reset wallet disabled for
good. A submit is now refused before it sets the spinner while a lockout is armed, and the edge into a
lockout clears one already running. `shouldClearPasswordField` becomes `didLockoutArm`, since that edge
now drives two consequences.

**Not covered:** an SSR test for `VaultLockedPage`. `mac-scrollbar` ships an ESM-only entry, so anything
importing `@libs/ui/components` — which `@libs/layout` does — fails to resolve under jest. The routing
decision is pinned in `resolve-route.test.ts`; onboarding e2e is 9/9.


src/background/redux/sagas/vault-sagas.ts
```diff
@@ -440,6 +456,30 @@ export function* setDelayForLockoutVaultSaga(
...
+  }
+
+  // Not just a burst guard: without it a resume would restart the lockout clock.
+  if (yield* sagaSelect(selectHasLoginRetryLockoutTime)) {
+    return;
+  }
+
+  yield put(loginRetryLockoutTimeSet(Date.now()));
+
+  if (!(yield* sagaSelect(selectVaultIsLocked))) {
```
Comp0te, line 478: `lockVault()` here also fires for the onboarding re-auth gate — a surface the hook never covered. Locking resets `encryptionKeyDoesExist`, so the tab falls to `WelcomePage`: one button, no confirmation, wipes the wallet. That screen isn't new; the five-failure route into it is. Needs an unlocked vault and a session-fresh tab.

Should the onboarding router treat `keysDoesExist && !encryptionKeyDoesExist` as locked?

Dispatcher: `form-validation-rules.ts:70` → `useUnlockWalletForm` (`unlock-wallet.ts:16`) → `onboarding/pages/unlock-wallet/index.tsx:92`. At base the hook was mounted only at `unlock-vault/index.tsx:236` and `unlock-protected-page-content/index.tsx:52`.

Chain: `:478-480` → `lockVaultSaga` → `sessionReseted` (`session/reducer.ts:16`) → `onboarding/app-router.tsx:43` → `welcome/index.tsx:19-25` → `resetVault()` → `onboarding-sagas.ts:54-76`, clearing keys, cipher, contacts and `storage.local`. At base that point showed the 0-tries screen (`unlock-wallet/index.tsx:99-126`) routing to the checkbox-gated `ResetWalletPage` (`reset-wallet/index.tsx:35-44`).

Mid-onboarding is safe: `initKeysSage` sets `encryptionKeyHashCreated` without `vaultUnlocked` (`onboarding-sagas.ts:95-104`), so `isLocked` is true there and `:478` skips the lock. To reach it: open onboarding while locked, then unlock in the popup — the tab re-renders into the gate (`onboarding/index.tsx:27-47`) with `isLocked` false and `isLoggedIn` fresh (`use-session-storage.ts:25`).

</details>

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -538,6 +578,24 @@ export function* unlockVaultSaga(action: ReturnType<typeof unlockVault>) {
...
   try {
+    // Append-only errors + a route-independent banner mean a refused attempt's
+    // banner would otherwise outlive the retry that succeeds.
+    yield put(dismissSagaErrorsBySource('unlockVaultSaga'));
+
+    // Defence in depth, not the fix for WALLET-1424: a DROPPED
+    // `loginRetryLockoutTimeSet` leaves this selector null, so this cannot
+    // catch that. It closes the forged-`unlockVault` path, and once the
+    // background owns the unlock it is a belt-and-braces assertion.
+    if (yield* sagaSelect(selectHasLoginRetryLockoutTime)) {
```
Comp0te, line 589: The refusal path puts `sagaError` and returns, touching nothing that clears the page's `isLoading`. The banner does explain why — but the spinner stays, submit keeps `pointerEvents: 'none'`, Reset wallet stays disabled, and the new effect only resets the field. The page never recovers, even after the lockout expires.

Clear `isLoading` on the refusal outcome rather than relying on unmount?

`setIsLoading` has exactly three sites in `unlock-vault/index.tsx`: `:117` true, `:140` and `:164` false. The success path never clears it — `popup/app-router.tsx:171-175` unmounts `LockedRouter` instead. Stuck true, `:269` is `pointerEvents:'none'`, `:290` is `disabled={isLoading}`, and `:108` early-returns; the effect at `:100-105` touches only `resetField`.

Reach: `signature-request/app-router.tsx:25-27` and the connect-to-app equivalent render the same page, so a second surface's fifth failure can arm the lockout inside the submit→dispatch window. A cold start does it too — `get-main-store.ts:145` dispatches `startBackground` at init, ahead of the `unlockVault` message, so a persisted count at the limit with no lockout arms on resume.

Bounded: it strands one page instance. The popup can be closed and reopened; a signature-request or connect-to-app window has to be closed, which abandons the pending request.

</details>

src/background/redux/sagas/vault-sagas.ts
```diff
@@ -114,6 +122,14 @@ export function* vaultSagas() {
     [loginRetryLockoutTimeSet.type, popupWindowInit.type, startBackground.type],
     setDelayForLockoutVaultSaga
   );
+  // Registered AFTER setDelayForLockoutVaultSaga on purpose. That saga is a
+  // takeLatest that also watches startBackground; arming first on a resume would
+  // start a run the still-queued startBackground delivery then cancels, possibly
+  // before the deadline write.
+  yield takeEvery(
+    [loginRetryCountIncremented.type, startBackground.type],
```
Comp0te, line 130: Replacing this whole array with `[loginRetryCountReseted.type]` leaves `vault-sagas.test.ts` at 68/68. The increment half is caught by two e2e specs — outside `ci-check`. The `startBackground` half is caught nowhere: all nine `startBackground` uses in the suite invoke a saga directly, and no e2e restarts the worker.

Worth pinning with `expectSaga(vaultSagas).dispatch(...)`, as `:550` and `:585` already do for the debounce list?

To repeat it: swap the array at `:130` for `[loginRetryCountReseted.type]` — already imported at `:483`, so lint, tsc and knip stay clean — and run `npx jest src/background/redux/sagas/vault-sagas.test.ts`. 68/68, with the saga unwired from both triggers.

The four `armLockoutSaga` tests (`:255-302`) call the saga directly with no action. The nine `startBackground` sites (`:168, :188, :225, :240, :326, :338, :362, :372, :1320`) are all direct invocations. `e2e-tests/fixtures.ts:108-129` reads `background.url()` only, to derive the extension id.

On the description: Testing lists "stale count on resume", but the test of that name (`:280-289`) provides `hasLoginRetryLockoutTime: true` and asserts `.not.put(...)` — it pins the no-re-arm guard, not that a resume reaches the saga.

`get-main-store.test.ts:34-49` is the only harness that dispatches `startBackground` into the real root saga; seeding a count past the limit with no lockout there would pin the trigger and the rehydration together.

</details>

src/libs/layout/unlock-vault/index.tsx
```diff
@@ -89,6 +89,21 @@ export const UnlockVaultPage = ({ popupLayout }: UnlockVaultPageProps) => {
...
 
+  const hasLoginRetryLockoutTime = useSelector(selectHasLoginRetryLockoutTime);
+  const wasLockedOut = useRef(hasLoginRetryLockoutTime);
+
+  // Keyed on the transition, not on `count >= limit && !lockout`: the background
+  // arms the lockout in the same dispatch that increments, so that intermediate
+  // state may never reach a replica. The lockout screen unmounts this field and
+  // react-hook-form keeps values across unmount, so without this the wrong
+  // password is still there when the lockout expires.
+  useEffect(() => {
```
Comp0te, line 96: This false→true ref edge replaced the deleted hook's derived condition, and nothing exercises it: no test renders `UnlockVaultPage`, `jest.config.js` scopes coverage to reducers, handlers and sagas, and the e2e never returns to the form after the lockout. Remove or invert it and the remounted field keeps the known-wrong password.

One click then burns an attempt. Worth extracting a testable unit?

`UnlockVaultPage` is referenced only at its definition and `locked-router/index.tsx:5,21`. `jest.config.js:3` is `testEnvironment: 'node'` and `:44-56` limits `collectCoverageFrom` to reducers, handlers and sagas; every component test in the repo uses `renderToStaticMarkup`, which runs no effects. `popup/common/wallet.spec.ts:85-106` and `backup-secret-phrase.spec.ts:35-51` fill once, click five times, assert the heading, and never come back to the form.

The value does survive: `content.tsx:37-76` drops the `Input` for the whole lockout, `shouldUnregister` appears nowhere in the tree, and react-hook-form `^7.85.0` retains values across unmount by default.

Milder than it looks, though: `setDelayForLockoutVaultSaga` puts `loginRetryCountReseted()` at `:454` before `loginRetryLockoutTimeReseted()` at `:455`, so the count is already 0 when the field remounts — a resubmit goes to 1 of 5, not to another immediate lockout.

</details>

Comp0te (commented):
Reviewed the move of the retry lockout into the background: the saga wiring, the refusal branch in `unlockVaultSaga`, and the re-keyed password reset on the unlock page. The direction is right, and the note on why the ticket's suggested `unlockVaultSaga` check is defence in depth rather than the fix is correct. The comments below are about the edges it opens — a surface the deleted hook never covered that now reaches a lock, a refusal outcome the page cannot recover from, and the parts of the change the suite does not pin.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Confirmed the whole chain and fixed it in `6a12cf26` (now `481319a7`'s parent).

`lockVault()` → `sessionReseted` clears `encryptionKeyDoesExist` while the keys survive → `onboarding/app-router.tsx` gated the entire tree on `keysDoesExist && encryptionKeyDoesExist` → `NoVaultRoutes` → `WelcomePage`, whose button calls `resetVault()` with no confirmation.

Answering your question directly: yes. Keys are what prove a wallet exists; `encryptionKeyDoesExist` only says the session is live. The gate is now `keysDoesExist` alone, with `encryptionKeyDoesExist && isLoggedIn` deciding between the authorized and unlock trees.

The decision moved into a pure `resolveOnboardingRoute` with four cases, since the router itself is not unit-testable here.

One thing worth recording that the fix had to survive: `initKeysSage` sets the two flags in two separate `put`s, and `get-main-store.ts`'s subscriber has no debounce, so `keysDoesExist && !encryptionKeyDoesExist` really is broadcast mid-onboarding. That intermediate state now resolves to `unlock` rather than `no-vault`. The onboarding e2e suite passes unchanged (8 passed, 1 pre-existing SW-registration flake), so the transient does not disturb the flow.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed. The refusal path no longer strands the page.

Worth noting the shape of the fix: Layer 3 rewrote this handler with `try/finally`, so the top of the stack was already correct and only this PR carried the defect. Rather than leave it to be fixed above, the `finally { setIsLoading(false) }` is now here too — each PR in the stack should be shippable on its own.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Fixed in `481319a7`. Both triggers now dispatch through the real root saga, following the `:550`/`:585` pattern you pointed at.

Verified your mutation: replacing the array with `[loginRetryCountReseted.type]` now fails both new tests.

You were right that the `startBackground` half was covered nowhere — that is the trigger that arms a lockout for a count which survived a restart, and it had no coverage at any level.

A trap for whoever touches these next: arming wakes `setDelayForLockoutVaultSaga`, which reads `selectLoginRetryLockoutTime`. Without a provider for it the test fails with `Cannot read properties of undefined`, which looks like a wiring problem and is not. That is noted in the test.

Also corrected the description: the "stale count on resume" test pins the no-re-arm guard, not the resume trigger, exactly as you said.

src/libs/layout/unlock-vault/index.tsx
ost-ptk: Extracted, as suggested — `shouldClearPasswordField(wasLockedOut, isLockedOut)`, with four cases and the rationale moved onto the function.

Confirmed it catches the mutation: inverting the condition now fails two of the four.

Your mitigation is accurate and I left it out of the code comment as noise: `setDelayForLockoutVaultSaga` puts `loginRetryCountReseted()` before `loginRetryLockoutTimeReseted()`, so a resubmit goes to 1 of 5 rather than straight back to a lockout.

src/background/redux/sagas/vault-sagas.ts
Comp0te: The wipe path is closed — `resolveOnboardingRoute` returns `unlock` whenever `keysDoesExist`, so the unconfirmed `resetVault()` can no longer front a keyed wallet. The tree it routes to is inert, though: the re-auth gate's `onSubmit` never emits `encryptionKeyHashCreated`, so a correct password re-renders the same screen.

Can that gate reach `unlockVault`, or should `unlock` route somewhere that can?

<details><summary>Basis, at `481319a7`</summary>

`unlock-wallet/index.tsx:94-97` dispatches only `loginRetryCountReseted()` and `saveIsLoggedIn(true)`. `encryptionKeyHashCreated` has three producers — `onboarding-sagas.ts:103`, `vault-sagas.ts:371`, `vault-sagas.ts:619` — and `unlockVault` is dispatched from one place in the tree, `libs/layout/unlock-vault/index.tsx:217,226`, which onboarding does not render. So `encryptionKeyDoesExist` stays false, `resolve-route.ts:18` keeps returning `unlock`, and the only live control is "Start again" → the checkbox-gated `ResetWalletPage`.

Two populations land there. This finding's own scenario once the lockout expires — still far better than a one-click wipe. And abandoned onboarding: `initKeys` persists `keys` at the create-password step, `session` is not persisted (`get-main-store.ts:44-57,96-140`), and `isOnboardingCompleted()` is `keysDoesExist && vaultCipherDoesExist` (`background/index.ts:70-77`), so the toolbar still opens the onboarding tab for them. At base they got "Get started" → `resetVault()` — harmless, no vault to lose. Now they get a form that cannot succeed.

</details>

src/background/redux/sagas/vault-sagas.ts
Comp0te: The `finally { setIsLoading(false) }` isn't in `481319a7`. `unlock-vault/index.tsx` still has exactly two `setIsLoading(false)` sites — `:138` and `:162` — `handleUnlockVault` has no `try`/`finally`, and `vault-sagas.ts` has zero diff lines between the two heads. Did it get squashed into the wrong commit?

<details><summary>What's at the pushed head</summary>

`setIsLoading(true)` at `:115`; the only clears are `handleWorkerFailure` (`:138`) and the client-side wrong-password branch (`:162`) — the same two as the old head at `:140`/`:164`. The only change to that file in the increment is the `shouldClearPasswordField` extraction; the effect at `:95-103` still touches only `resetField`.

The refusal branch here is still `put(sagaError(...))` plus a bare `return`, and the saga's `finally` releases the service-worker anchor only. So the strand is unchanged: submit keeps `pointerEvents: 'none'` (`:267`), Reset wallet stays `disabled={isLoading}` (`:288`), and the re-entry guard at `:106` early-returns.

</details>

Comp0te (commented):
Please check 2 issues

src/background/redux/sagas/vault-sagas.ts
ost-ptk: Confirmed, and the answer to your question is no — it should not reach `unlockVault`. #1498 puts onboarding on `VERIFY_PASSWORD_REQUEST` but deliberately off `UNLOCK_REQUEST`, pinned by `admits onboarding for verify but not for unlock`. Letting this gate unlock would undo that. So the fix is the route, not the gate.

`unlock` was doing two jobs, and it is now split. `reauth` is this tab's session being stale while the vault is open — the existing form, which can succeed. `locked` is the vault itself being shut, and it gets `VaultLockedPage`: no password field, and the checkbox-gated `ResetWalletPage` as its only action.

Both populations you named land there. The locked vault, once the lockout expires. And abandoned onboarding — same reset they got at base, now behind a confirmation instead of one click.

One thing the fix had to survive, since you flagged the mid-onboarding transient earlier: with `locked` in the router, `keys && !session` between `initKeysSage`'s two puts would flash the locked screen over the create-password page. The puts are reordered — session first, where the intermediate state reads as a fresh install whose routes still include the page that dispatched. `onboarding-sagas.test.ts` pins it; restoring the old order fails it.

One gap worth stating: I could not add an SSR test for the page itself. `mac-scrollbar` ships only an ESM entry with no `main`, so anything importing `@libs/ui/components` — which `@libs/layout` does through its header — fails to resolve under jest. The routing decision is covered in `resolve-route.test.ts`; the page is markup with no branches. Onboarding e2e is 9/9.

src/background/redux/sagas/vault-sagas.ts
ost-ptk: You are right, and my previous reply was wrong. The `try`/`finally` only ever existed in `f3da535c` on the Layer 3 branch; nothing of it reached this one. No squash went astray — the backport I claimed simply never happened.

Fixed here now, and the shape had to differ from L3's: at this layer the unlock still runs in page workers behind async `onmessage`, so there is no single `try` to wrap. Both ends of the window are closed instead. A submit while a lockout is already armed returns before setting the spinner, and the edge into a lockout clears one that is already running — the same edge that resets the field, so the refusal that answers with a banner alone no longer strands the page. The submit button is inert for the duration, so the guard is never what a user runs into.

`shouldClearPasswordField` is renamed to `didLockoutArm`, since the edge now drives two consequences and naming it after one made the other read as an accident.

Comp0te (approved):

PR #1497, opened 2026-08-20
feat(background): route secret-carrying requests over a gated port

ost-ptk:

### Why

`runtime.sendMessage` delivers to every open page of this extension, not only the background.
`changePassword` carries two plaintext passwords over it, and the unlock flow (PR 3) is about to carry
more. Those requests move to a point-to-point `runtime.connect` port that only the background sees.

To be precise about the threat: `sendMessage` does **not** reach dapp pages or other extensions. The
exposure is to other pages of this extension — which is exactly the surface the on-demand secret
handlers already guard against.

### Gating

Same shape as the existing secret handlers: `isTrustedUiSender` plus a per-request page allowlist.
Being an extension page proves origin, not need.

### Two subtleties worth a reviewer's attention

**The listener is attached synchronously, before the store is awaited.** On a cold service-worker
start the page's already-sent message reaches only listeners present at that instant. Dropping it
would leave the port _connected_, so the caller's disconnect-retry would never fire and its promise
would never settle.

**Retry policy is asymmetric on purpose.** A disconnect before a response is retried — it proves the
message produced no result. A timeout is _not_ retried — the background may already have acted on it.
A generous liveness backstop covers the case that yields neither.

### Behaviour

No happy-path change. **The error path is new:** `dispatchToMainStore` swallowed a failed
change-password dispatch; failure now surfaces through the existing UI error channel. Please exercise
the failure path, not just the happy one.

`changePasswordSaga` itself is unchanged — only the transport moves — and `changePassword` leaves
`FORWARDED_ACTION_TYPES`.

### DIFF
--- a/src/apps/popup/pages/change-password/index.tsx
+++ b/src/apps/popup/pages/change-password/index.tsx
@@ -7,7 +7,6 @@ import { PasswordProtectionPage } from '@popup/pages/password-protection-page';
 import { RouterPath, useTypedNavigate } from '@popup/router';
 
 import { CHANGE_PASSWORD_REQUEST_TYPE } from '@background/handlers/privileged-port';
-import { changePassword } from '@background/redux/sagas/actions';
 
 import {
   FooterButtonsContainer,
@@ -18,14 +17,12 @@ import {
 import { requestOverPort } from '@libs/messaging/background-port';
 import { Button, PasswordInputs } from '@libs/ui/components';
 import {
-  clearUiError,
-  reportUiError
-} from '@libs/ui/components/saga-error-banner/ui-error-channel';
-import {
   CreatePasswordFormValues,
   useCreatePasswordForm
 } from '@libs/ui/forms/create-password';
 import { calculateSubmitButtonDisabled } from '@libs/ui/forms/get-submit-button-state-from-validation';
+
+import { submitPasswordChange } from './submit-password-change';
 
 export const ChangePasswordPage = () => {
   const [isPasswordConfirmed, setIsPasswordConfirmed] =
@@ -41,7 +38,7 @@ export const ChangePasswordPage = () => {
   const {
     register,
     handleSubmit,
-    formState: { isDirty, errors },
+    formState: { isDirty, isSubmitting, errors },
     control
   } = useCreatePasswordForm();
 
@@ -58,26 +55,20 @@ export const ChangePasswordPage = () => {
     setCurrentPassword(password);
   }, []);
 
-  const isSubmitButtonDisabled = calculateSubmitButtonDisabled({
-    isDirty
-  });
+  const isSubmitButtonDisabled =
+    calculateSubmitButtonDisabled({ isDirty }) || isSubmitting;
 
-  const onSubmit = (data: CreatePasswordFormValues) => {
+  const onSubmit = async (data: CreatePasswordFormValues) => {
     if (currentPassword == null) return;
 
-    requestOverPort({
-      type: CHANGE_PASSWORD_REQUEST_TYPE,
-      payload: { currentPassword, password: data.password }
-    })
-      .then(() => {
-        clearUiError('dispatch-failed', changePassword.type);
-      })
-      .catch((error: unknown) => {
-        console.error('Password change request failed:', error);
-        reportUiError('dispatch-failed', changePassword.type);
-      });
-
-    navigate(RouterPath.Home);
+    await submitPasswordChange(
+      () =>
+        requestOverPort({
+          type: CHANGE_PASSWORD_REQUEST_TYPE,
+          payload: { currentPassword, password: data.password }
+        }),
+      () => navigate(RouterPath.Home)
+    );
   };
 
   if (!isPasswordConfirmed) {
--- a/src/apps/popup/pages/change-password/index.tsx
+++ b/src/apps/popup/pages/change-password/index.tsx
@@ -6,8 +6,8 @@ import { ChangePasswordPageContent } from '@popup/pages/change-password/content'
 import { PasswordProtectionPage } from '@popup/pages/password-protection-page';
 import { RouterPath, useTypedNavigate } from '@popup/router';
 
+import { CHANGE_PASSWORD_REQUEST_TYPE } from '@background/handlers/privileged-port';
 import { changePassword } from '@background/redux/sagas/actions';
-import { dispatchToMainStore } from '@background/redux/utils';
 
 import {
   FooterButtonsContainer,
@@ -15,7 +15,12 @@ import {
   HeaderSubmenuBarNavLink,
   PopupLayout
 } from '@libs/layout';
+import { requestOverPort } from '@libs/messaging/background-port';
 import { Button, PasswordInputs } from '@libs/ui/components';
+import {
+  clearUiError,
+  reportUiError
+} from '@libs/ui/components/saga-error-banner/ui-error-channel';
 import {
   CreatePasswordFormValues,
   useCreatePasswordForm
@@ -60,9 +65,17 @@ export const ChangePasswordPage = () => {
   const onSubmit = (data: CreatePasswordFormValues) => {
     if (currentPassword == null) return;
 
-    dispatchToMainStore(
-      changePassword({ currentPassword, password: data.password })
-    );
+    requestOverPort({
+      type: CHANGE_PASSWORD_REQUEST_TYPE,
+      payload: { currentPassword, password: data.password }
+    })
+      .then(() => {
+        clearUiError('dispatch-failed', changePassword.type);
+      })
+      .catch((error: unknown) => {
+        console.error('Password change request failed:', error);
+        reportUiError('dispatch-failed', changePassword.type);
+      });
 
     navigate(RouterPath.Home);
   };
--- a/src/background/handlers/privileged-port.ts
+++ b/src/background/handlers/privileged-port.ts
@@ -1,0 +1,127 @@
+import { Runtime, runtime } from 'webextension-polyfill';
+
+import { MainStore } from '@background/redux/get-main-store';
+import { changePassword } from '@background/redux/sagas/actions';
+
+import { isTrustedUiSender } from './private-state';
+
+export const CHANGE_PASSWORD_REQUEST_TYPE = 'CHANGE_PASSWORD_REQUEST' as const;
+
+const POPUP_PAGE = '/popup.html';
+
+// isTrustedUiSender proves "an extension page"; it does not prove "a page that
+// needs this". Same shape and same reasoning as vault-secrets.ts.
+export const ALLOWED_PAGES: Record<string, readonly string[]> = {
+  [CHANGE_PASSWORD_REQUEST_TYPE]: [POPUP_PAGE]
+};
+
+export function isAllowedPage(
+  type: string,
+  sender: Runtime.MessageSender
+): boolean {
+  if (sender.url == null) {
+    return false;
+  }
+
+  const pages = ALLOWED_PAGES[type];
+
+  return pages != null && pages.includes(new URL(sender.url).pathname);
+}
+
+function isChangePasswordRequestPayload(
+  payload: unknown
+): payload is { currentPassword: string; password: string } {
+  const p = payload as
+    Partial<{ currentPassword: string; password: string }> | null | undefined;
+
+  if (p == null) {
+    return false;
+  }
+
+  return (
+    typeof p.currentPassword === 'string' &&
+    typeof p.password === 'string' &&
+    p.password.length > 0
+  );
+}
+
+/**
+ * `null` means refused or unknown: the caller disconnects without answering,
+ * matching `vault-secrets.ts`.
+ */
+export async function handlePrivilegedRequest(
+  request: { type: string; payload?: unknown },
+  sender: Runtime.MessageSender,
+  store: MainStore
+): Promise<unknown | null> {
+  if (!Object.hasOwn(ALLOWED_PAGES, request.type)) {
+    return null;
+  }
+
+  if (!isTrustedUiSender(sender) || !isAllowedPage(request.type, sender)) {
+    // Same-extension only, and origin only — a content script or web page can
+    // connect to this port too, and its sender.url can carry a query string.
+    if (sender.id === runtime.id) {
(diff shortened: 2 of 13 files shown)


src/background/handlers/privileged-port.ts
```diff
@@ -0,0 +1,127 @@
...
+export async function handlePrivilegedRequest(
+  request: { type: string; payload?: unknown },
+  sender: Runtime.MessageSender,
+  store: MainStore
+): Promise<unknown | null> {
+  if (!Object.hasOwn(ALLOWED_PAGES, request.type)) {
+    return null;
+  }
+
+  if (!isTrustedUiSender(sender) || !isAllowedPage(request.type, sender)) {
```
Comp0te, line 61: Four of the tests added here stay green under a mutation of the exact thing they exist to pin. On this line, the `!isTrustedUiSender(sender) ||` leg can be deleted and `privileged-port.test.ts` still reports 17/17 — every rejected fixture in it is already rejected by `isAllowedPage` alone.

One fixture and three assertions would close all four.

**1. `privileged-port.ts:61` — the two legs of the `||` are never separated.** `POPUP_SENDER` (test:21), `CONNECT_SENDER` (:26, `/connect-to-app.html`), `WEB_PAGE_SENDER` (:31), `FOREIGN_EXTENSION_SENDER` (:36), the inline `NO_URL_SENDER` (:167, :231) and `makePort(undefined)` (:341) all fail `isAllowedPage` on pathname alone. The leg no fixture isolates is same `id` + allow-listed pathname + foreign origin — and `isTrustedUiSender` is the only check that rejects it. `vault-secrets.test.ts:42-47` already defines exactly that fixture (`SPOOFED_ORIGIN_SENDER`, at `https://evil.example/popup.html`) and exercises it at :93-101; copying it across and asserting `handlePrivilegedRequest` resolves `null` with `dispatch` uncalled is the whole fix. To reproduce: delete `!isTrustedUiSender(sender) || ` from line 61, run `npx jest src/background/handlers/privileged-port.test.ts` — 17 passed, identical to baseline.

**2. `redux-actions.parity.test.ts:199` — the set algebra can't see a move between the two sets.** The six assertions at :215-244 are all relations that stay true when a type moves from `FORWARDED_ACTION_TYPES` to `EXCLUSIONS` or back. A lone re-add does fail disjointness (:220-223) and equality (:241-244) — but the paired edit, re-adding it while deleting the exclusion entry, passes all 35. The file already recognises this and closes it for one entry, the standalone `windowRequestWindowAttached` test at :246-257; `expect(FORWARDED_ACTION_TYPES.has(sagasActions.changePassword.type)).toBe(false)` gives `changePassword` the same pin. (Membership doesn't itself put anything on the wire — it's what makes the background *accept* such a message from any trusted-UI page.)

**3. `background-port.test.ts:53` — the client's wire contract is unasserted.** `mockConnect` is checked by count only (:69, :81, :95, :117), never `toHaveBeenCalledWith`, so `runtime.connect({ name: BACKGROUND_PORT_NAME })` at `background-port.ts:32` reduces to `runtime.connect()` with 5/5 still passing — while `background/index.ts:314-320` returns early for any other name *without disconnecting*, so that mutation costs the user the full 60s backstop with no retry. Every request in the file is payload-less, so `port.postMessage(request)` at :56 reduces to `postMessage({ type: request.type })` with 22/22 passing, though `privileged-port.ts:31-46` refuses anything missing `currentPassword`/`password`. Nothing imports both modules, so nothing checks that what the client sends is what the background accepts.

**4. `background-port.test.ts:65` — the backoff values are unasserted.** Both retry tests read `mockConnect` only after the advance, so `PORT_RETRY_DELAYS_MS = [0, 0]` passes unchanged. The retry *count* is genuinely pinned (`[250]` would make the :98-118 test see 2 connects against its `toHaveBeenCalledTimes(3)`). Note that asserting a count *before* advancing does not catch this — under `jest.useFakeTimers()` a 0 ms timer hasn't fired at that point either. What works: `await jest.advanceTimersByTimeAsync(249)` → expect 1 connect, then advance the remaining 1 ms → expect 2; or assert the exported constant directly.

</details>

src/apps/popup/pages/change-password/index.tsx
```diff
@@ -60,9 +65,17 @@ export const ChangePasswordPage = () => {
...
-      changePassword({ currentPassword, password: data.password })
-    );
+    requestOverPort({
+      type: CHANGE_PASSWORD_REQUEST_TYPE,
+      payload: { currentPassword, password: data.password }
+    })
+      .then(() => {
+        clearUiError('dispatch-failed', changePassword.type);
+      })
+      .catch((error: unknown) => {
```
Comp0te: These two branches are the first user-facing signal that a password change never reached the background — and nothing tests either. There's no unit or component test for this page anywhere in `src/`, and the e2e spec covers only the happy path and the mismatch validation.

Could you add a unit test shaped like `redux/utils.test.ts:97-108`?

Delete the `.catch` and nothing fails, but at runtime a timed-out or retry-exhausted request produces an unhandled rejection and no banner — while `navigate(RouterPath.Home)` on :80 has already run. The user is on Home believing the password changed; on the next unlock the old one is still live, and each failed attempt increments `loginRetryCount`, which `armLockoutSaga` (`vault-sagas.ts:130-131`) uses to arm the lockout. Delete the `.then` instead and a stale "didn't respond" row outlives the retry that worked — the case `ui-error-channel.ts:36-40` documents.

`create-reporting-open-window.test.ts:52-62` is a second example of the same shape.

For e2e, `breakTransport` (`error-surface.ts:26-49`) patches `chrome.runtime.sendMessage` only, so it can't drive this path — new work rather than a repair, since at base the failure was swallowed silently and there was no signal here to lose. A `breakPort` sibling patching `chrome.runtime.connect` would close it.

</details>

src/apps/popup/pages/change-password/index.tsx
```diff
@@ -60,9 +65,17 @@ export const ChangePasswordPage = () => {
...
+        console.error('Password change request failed:', error);
+        reportUiError('dispatch-failed', changePassword.type);
```
Comp0te: `surfaced-dispatch-actions.test.ts:19-35` pins which types may raise a `dispatch-failed` banner. This call goes direct rather than through `utils.ts:49-51`, so it raises one for a type the pin excludes — and the pin stays green.

Add `changePassword.type` to the allowlist, or widen the pin to cover direct call sites?

The pin's own comment says growing the list has to be "a deliberate edit, not a side effect". The three production `reportUiError` call sites are `utils.ts:50` (allowlist-gated), `create-reporting-open-window.ts:34` (kind `window-open-failed`, already outside the pin) and this one. So the pin has never described the whole banner — it governs the `dispatch-failed` kind specifically, and this is the first `dispatch-failed` report that bypasses it.

What that costs: a second ungated `reportUiError('dispatch-failed', X)` can be added anywhere later, and the test written to make that decision deliberate won't fire.

Not a defect either way — a decision worth making explicitly rather than by omission.

</details>

src/libs/messaging/background-port.ts
```diff
@@ -0,0 +1,85 @@
...
+/** A disconnect means the message never produced a result — safe to re-send. */
+const PORT_RETRY_DELAYS_MS = [250, 500];
+
+/**
+ * Liveness backstop. A message delivered before the background attached its
+ * listener produces neither a response nor a disconnect, so retry alone cannot
+ * see it. Generous on purpose: it must never fire during a real scrypt
+ * derivation, which `request-with-retry.ts`'s 5s could not bound.
+ */
+export const PORT_RESPONSE_TIMEOUT_MS = 60_000;
```
Comp0te, line 21: The justification above doesn't hold — the response never waits on scrypt. `handlePrivilegedRequest` produces `{ accepted: true }` at `privileged-port.ts:85-86` right after `store.dispatch`, and `changePasswordSaga`'s first suspending `call` is at `vault-sagas.ts:309`. This is 12× `request-with-retry.ts`'s 5s for a reason that doesn't apply.

Bring it down to that scale and correct the comment?

Everything before that `call` is `put`/`select` (`vault-sagas.ts:271-307`), so `dispatch` returns without awaiting it; the other awaited step, `getExistingMainStoreSingletonOrInit` (`get-main-store.ts:80-109`), is a `storage.local.get`. (The ack is produced at :85-86, posted at :116.)

Why the size matters rather than just being generous: `change-password/index.tsx:80` navigates to Home before the ack arrives — before any outcome is known — and `ui-error-channel.ts:1-5` states the channel is UI-local, never reaching the store or the replica. A user who closes the popup inside that minute never sees the banner and has no persisted trace on reopening, unlike a saga-level failure, which reaches the background store via `sagaError`.

Optional, separately: consider not navigating until the ack lands — `PasswordProtectionPage` in this flow has the `isSubmitting`/`readOnly` precedent.

</details>

src/libs/messaging/background-port.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    port.onDisconnect.addListener(() => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      reject(new PortDisconnectedError());
+    });
+
+    port.postMessage(request);
```
Comp0te: The one exit from `attempt()` that leaves `timer` armed — the listeners clear it at :44 and :52. A synchronous `postMessage` throw rejects with `settled` still false, holding the closure and both passwords for the remaining 60s. No reachable trigger today: an invalidated context throws at `runtime.connect` (:32), before the timer exists.

Same try/catch as the other two exits?

The rejection isn't a `PortDisconnectedError`, so `requestOverPort` (:77-79) rethrows immediately and the banner shows while the timer still holds the executor's closure — and with it `request`, carrying `currentPassword` and `password`.

On reachability, so this isn't overstated: nothing can disconnect the port between :32 and :56, so the only remaining synchronous-throw cause is a non-cloneable payload, and the sole call site (`change-password/index.tsx:68-71`) passes two strings. Treat it as exit-path uniformity on a secret-carrying exported API rather than a live leak.

Fix is the shape already used twice in this function: set `settled`, `clearTimeout(timer)`, reject.

</details>

Comp0te (commented):
Read the new port transport end to end against the base branch — the background handler and its sender gate, the client's retry and timeout behaviour, the parity suite, and the change-password call site. The through-line of the comments: the runtime shape of the change looks right, and the tests added alongside it largely don't pin the parts carrying the security property, so one-token mutations of the new code leave the suite green. Two comments are about the transport itself rather than its coverage.

src/background/handlers/privileged-port.ts
ost-ptk: All four fixed, in `6a8c32d6` and `96902d28`.

1. `SPOOFED_ORIGIN_SENDER` copied across from `vault-secrets.test.ts` and asserted to resolve `null` with `dispatch` uncalled. Verified your reproduction: deleting `!isTrustedUiSender(sender) || ` from line 61 now fails that test.
2. `changePassword` has a standalone membership pin, the same shape as the `windowRequestWindowAttached` one at `:246-257`.
3. `mockConnect` is now asserted with `toHaveBeenCalledWith({ name: BACKGROUND_PORT_NAME })`, and one request carries a real payload so `postMessage(request)` cannot be reduced to the type alone.
4. The backoff is pinned by advancing to 249ms → 1 connect, then 1ms more → 2, exactly as you described. Your note that asserting the count *before* advancing does not catch it saved a wrong fix — a 0ms timer has not fired at that point either.

All four mutations were re-run against the new tests and each one fails now.

src/apps/popup/pages/change-password/index.tsx
ost-ptk: Added, and the branch it guards changed too.

The logic moved into `submitPasswordChange`, tested in the shape you pointed at: success calls `clearUiError` and proceeds, failure calls `reportUiError` and does **not** proceed, plus a check that the payload never reaches the log.

The "does not proceed" half is the substantive part — see the reply on `:15`. The page no longer navigates before the ack, so the failure you describe (on Home, believing the password changed, old one still live, each later attempt counting toward the lockout) is no longer reachable.

Agreed on `breakPort` being new work rather than a repair; leaving it out of this stack.

src/apps/popup/pages/change-password/index.tsx
ost-ptk: Took the first option: `changePassword.type` is now in `SURFACED_DISPATCH_ACTIONS`, with a comment saying why it is there despite not going through `utils.ts` — so the set describes every `dispatch-failed` banner rather than only the ones the helper raises.

Both pins updated, the creator-derived one and the literal type-string one.

Your framing was the useful part: not a defect, but a decision better made explicitly than by omission. The cost you named — a second ungated reporter added later with the pin never firing — is what settled it.

src/libs/messaging/background-port.ts
ost-ptk: Half agreed, and I have not changed the value — reasoning below, happy to be overruled.

You are right about this branch: `handlePrivilegedRequest` produces `{ accepted: true }` right after `store.dispatch`, `changePasswordSaga`'s first suspending `call` is later, so nothing here waits on scrypt and the justification as written describes something that does not exist yet. That part of the comment was wrong and is rewritten.

What does not follow is lowering it. In PR #1498 this same port carries `UNLOCK_REQUEST` and `VERIFY_PASSWORD_REQUEST`, and those answer only *after* their derivation — which is serialised process-wide, so a queued request waits out every one ahead of it. Dropping it to ~5s here and raising it again one PR later is churn that leaves the middle of the stack wrong, so the value stays and the comment now says it is sized for the requests the next layer puts on this port.

Your optional half I did take: the page no longer navigates to Home before the ack. It awaits the outcome, and the submit button is disabled while a request is in flight — the `isSubmitting`/`readOnly` precedent you cited.

src/libs/messaging/background-port.ts
ost-ptk: Fixed with the shape you described — `settled`, `clearTimeout(timer)`, `port.disconnect()`, reject.

Your reachability analysis holds and I kept the framing in the commit message: no live trigger today, exit-path uniformity on a secret-carrying API rather than a leak.

Pinned by asserting `disconnect` is called exactly once even after advancing past twice the backstop — if the timer were still armed it would fire and disconnect again.

Comp0te (approved):

PR #1498, opened 2026-08-20
feat(vault): perform the unlock in the background

ost-ptk:

### What this closes

Password verification and the whole vault unlock move out of page-local Web Workers into the
background. The page sends a password over the privileged port and receives only a status.

This is what makes the retry lockout enforceable: the background observes every failed attempt,
counts it, arms the lockout and refuses — where before all three lived in replica state that a dropped
message could skip.

Two consequences beyond the ticket:

- **Offline grinding is closed.** `fetchPrivateState` handed `passwordHash`, `passwordSaltHash`,
  `keyDerivationSaltHash` and `vaultCipher` to any extension page, so a compromised page could grind
  passwords locally without ever touching the counter. No page receives any of them now, so the
  lockout becomes a real boundary rather than a control against casual access.
- **A cipher write sink is closed.** `unlockVaultSaga` accepted a **caller-supplied**
  `newVaultCipher` / `newEncryptionKeyHash` and persisted them to `storage.local`, and `unlockVault`
  was forwardable from any extension page. The background now computes those bytes itself.

`unlockVaultSaga` is deliberately unchanged: the background handler produces exactly the payload it
already accepted, keeping the most security-critical saga in the wallet out of this diff.

### Ordering and idempotency

The lockout check runs **before** the memo lookup, so a memoised verdict can never be a path past an
active lockout. A verdict is memoised by request type, attempt id and password for ten seconds, so a
response lost to a service-worker death cannot burn a second attempt on retry — and a repeat with a
_different_ password re-verifies rather than replaying a stale verdict.

Derivations are serialised process-wide. Without that, one compromised page could start eight
concurrent scrypt runs at ~256MB each and OOM the worker; serialising turns that into eight sequential
attempts that trip the lockout after five.

### Behaviour changes to know about

- The retry counter now resets at **verification** time rather than after the caller's follow-on
  action succeeds. Intended — the count measures password attempts.
- A service-worker death between the counter write and the response can over-count by up to the retry
  count. It fails toward locking out, never toward letting in.

### Deleted

`private-state.ts`, `use-private-state.ts`, `private-state-broadcast.ts`, and both page workers
(`verify-password-worker.ts`, `unlock-vault-worker.ts`). `unlockVault` and both retry-count actions
leave `FORWARDED_ACTION_TYPES`.

### Second commit — the scrypt offload

The unlock is now the most frequent background derivation in the wallet, and on the MV2 builds the
background is a **persistent page sharing the popup's renderer thread**. Left inline, this PR would
trade a UI freeze for a security fix on Firefox and Safari. So the derivation moves off that thread.

It also fixes a **pre-existing** bug: `changePasswordSaga` and `onboarding-sagas` already derive in
the background on `develop`, so those two flows already freeze the popup on MV2 today, with no
relation to this ticket.

**The measurement.** `scryptAsync` does not yield to the event loop. `nextTick` in `@noble/hashes` is
`async () => {}` — an empty resolved promise — so `asyncLoop` queues a **microtask**, and the loop
drains the whole microtask chain before timers, input, paint or message delivery. Measured here: one
derivation at `N = 2**18, r = 8` took **347ms, during which a 50ms `setInterval` fired zero times.**

(For the record, since review stated it the other way: Chrome has been MV3 since before this crypto
was first moved out of the background — `d4ea15d7:src/manifest.v3.json` is already
`"manifest_version": 3`. The thread problem only ever applied to Firefox and Safari.)

```
MV3 (Chrome)           service worker, own thread, no Worker constructor
                       -> derive inline; the popup is not on this thread anyway

MV2 (Firefox, Safari)  persistent background PAGE, shares the popup's thread
                       -> derive in a Worker the background owns
```

`scrypt-off-thread.ts` feature-detects `typeof Worker !== 'undefined'`; all six background scrypt call
sites route through it. **No security property changes** — the Worker is spawned by the background,
never by a page, and no secret joins any page-reachable channel. This is threading inside the trust
boundary.

Two implementation notes: `import.meta.url` (needed by webpack to emit the chunk) is a syntax error
under ts-jest's CommonJS output, so the spawn is isolated and mapped to a stub in `jest.config.js`;
and the worker reports failure as `postMessage({ error: true })`, because a rejection inside an async
`onmessage` raises no `error` event on the parent `Worker`.

**Manually confirmed on Firefox and Safari:** the popup stays responsive — spinner animating — for the
whole unlock, and the five-wrong-passwords lockout holds for its five minutes. Unlock takes the same
wall-clock time as before, as expected: this frees the thread, it does not make scrypt faster.

**Caveat — please read before approving.** The Firefox selenium e2e harness can no longer navigate to
`moz-extension://` (the untouched smoke spec fails identically on a clean tree), so **nothing
automated guards the MV2 path**. The manual check above is the only evidence, which is why the offload
is a separate commit in this PR. Repeat it if the offload or the worker chunking changes.

**Known residual:** the MV3 build still emits the ~9KB worker chunk it never loads.

### Testing

Full unit coverage of the verify/unlock handler (already-unlocked, lockout active, wrong password with
exact `attemptsLeft`, wrong at the limit, memo hit/miss/expiry, serialisation, eviction), the port
router, and vault normalisation. e2e specs pinning these flows need fixture-timing re-verification —
the page workers become a background round-trip.

The offload adds its own unit tests on both branches (inline fallback; Worker path with a fake worker,
covering the key, the terminate, and both failure shapes). `src/background/workers/` sits outside the
jest coverage universe, so those are deliberate rather than gate-driven.

Verified across the PR: `tsc`, `lint`, prettier clean; 114 suites / 1146 tests green; `build:firefox`
and `build:chrome` both succeed.


src/background/handlers/unlock-requests.ts
```diff
@@ -0,0 +1,239 @@
...
+
+const MEMO_LIMIT = 8;
+
+// Bounds how long a verdict can be replayed — not how long an entry stays in
+// memory (only FIFO eviction or a lookup on that exact key does that). Must
+// outlive the page's own retry window for a dropped response
+// (background-port.ts retries at +250ms, then +500ms).
+const MEMO_TTL_MS = 10_000;
+
+interface MemoEntry {
```
Comp0te, line 66: `MemoEntry` holds the caller's plaintext password in a module-level `Map`. The only removals are FIFO eviction at a 9th entry (`:95-102`) and a TTL checked lazily on lookup of that same key (`:228-229`) — neither bounds residency. The password that decrypts `vaultCipher` therefore outlives both auto-lock and manual lock; on the MV2 builds, the whole browser session.

Worth keying the memo on a per-process digest (random HMAC key + SHA-256) instead of the plaintext — `:230` only needs the equality check. If the plaintext stays, `memo.clear()` in `lockVaultSaga` plus `setTimeout(() => memo.delete(key), MEMO_TTL_MS)` in `remember` would at least make residency match the documented 10s.

- `memo` is module-private, so `lockVaultSaga` (`vault-sagas.ts:198-213`) cannot reach it: `encryptionKeyHash` is nulled while the more valuable secret stays.
- `:39-42` already concedes `MEMO_TTL_MS` bounds replayability, not residency. A within-TTL retry does re-look-up the key, but hits `:230-231` and leaves the entry in place.
- MV2 is persistent (`manifest.v2.json:44`). At `118c8657` this password went to a page-local Worker that was `terminate()`d, never to the background.
</details>

src/background/workers/scrypt-off-thread.ts
```diff
@@ -0,0 +1,78 @@
...
+): Promise<Uint8Array> {
+  if (!canOffloadToWorker) {
+    return scryptAsync(
+      password,
+      convertHexToBytes(saltHash),
+      createScryptOptions()
+    );
+  }
+
+  return new Promise<Uint8Array>((resolve, reject) => {
```
Comp0te, line 38: `deriveScryptKey` settles only on `onmessage` or `onerror` — there is no deadline. `serialise` chains everything on one module-level `derivationQueue`, and `derivationQueue = next.catch(...)` handles a rejection but cannot displace a promise that stays *pending*, so one never-settling derivation means `work` is never invoked again. Whether a Worker can die without firing `onerror` is unsettled; the missing deadline is not.

Worth giving the worker promise an explicit timeout that rejects and frees the queue slot, turning a permanent wedge into one failed attempt.

- Every other layer on this path already has one: `request-with-retry.ts` at 5s, and `background-port.ts` `PORT_RESPONSE_TIMEOUT_MS` at 60s, commented "Liveness backstop". The derivation has none.
- If it jams, every later `UNLOCK_REQUEST` / `VERIFY_PASSWORD_REQUEST` from all four pages in `privileged-port.ts:22-39` queues behind it, waits out the 60s timeout, and gets `Error('Background port timed out')` — not a `PortDisconnectedError`, so `background-port.ts:77-79` re-throws instead of retrying and the page shows the generic message.
- It clears only on a background restart — on MV2, where the page is persistent, that means a browser restart.
</details>

src/background/handlers/unlock-requests.ts
```diff
@@ -0,0 +1,239 @@
...
+) {
+  if (memo.size >= MEMO_LIMIT) {
+    // Map iterates in insertion order, so this is the oldest entry; size >=
+    // MEMO_LIMIT (> 0) guarantees at least one.
+    for (const key of memo.keys()) {
+      memo.delete(key);
+      break;
+    }
+  }
+  memo.set(memoKey, { password, result, createdAt: Date.now() });
```
Comp0te: `createdAt` is stamped at *enqueue* (`:236`, before the derivation runs), so the 10s replay window is spent deriving instead of starting when the verdict exists. A wrong-password request queued behind another unlock can then have its retry miss the memo, re-derive, and count one attempt twice — locking the user out early. Full reach unconfirmed; the stamping is wrong regardless.

Refreshing `createdAt` on settlement fixes it — but that extends how long the plaintext sits in the map, so the digest change suggested on `:45` is the better single fix: it makes residency irrelevant.

- The contract at `:39-43` defines the window against the client's *post-response* retry, measured from settlement; expiry is checked at `:228`.
- The double dispatch is visible in your own test: `unlock-requests.test.ts:166-186` takes the expired-miss path and produces two `loginRetryCountIncremented` dispatches.
- Only a wrong-password request can double-count — the correct path never increments (`:155-183`) and its retry hits the already-unlocked return at `:212-214`. So the latency comes from the process-wide queue (`:73-77`), not from its own single derivation at `:130`.
- The unconfirmed part is the rest of the conjunction: the port must disconnect after the handler settled, while the background is still alive.
</details>

src/libs/layout/unlock-vault/index.tsx
```diff
@@ -105,161 +75,43 @@ export const UnlockVaultPage = ({ popupLayout }: UnlockVaultPageProps) => {
...
-        // If the hidden property is undefined, set it to false
-        if (acc.hidden === undefined) {
-          return {
-            ...acc,
-            hidden: false
-          };
-        }
-
-        return acc;
+      // 'ok' unmounts this page via the broadcast, which needs no error here.
```
Comp0te, line 105: On `ok` the page does nothing locally, relying on the broadcast to unmount it — but `broadcastToReplicas` logs delivery failures and swallows them. If it is lost the spinner stops, the screen is unchanged, and nothing the user can do from there recovers it: resubmitting hits the already-unlocked return at `unlock-requests.ts:212-214`, which returns `ok` without dispatching, so no new broadcast.

Worth confirming the transition here rather than assuming it — re-read popup state on `ok`, or re-broadcast from the already-unlocked path so a dropped broadcast is retried rather than skipped.

- `broadcastToReplicas` at `get-main-store.ts:63-72`, called at `:150-153`.
- The routes that might self-heal do not: `useUserActivityTracker` attaches only when `!vaultIsLocked` (`use-user-activity-tracker.ts:33`) and the replica still reads locked; `unlockVaultSaga`'s puts (`vault-sagas.ts:603-618`) are upstream of the dropped broadcast. Remounting does re-sync (`use-subscribe-to-redux.ts:30-42`), so reopening the popup is the only fix from that screen — and nothing tells the user to.
- Not formally permanent — any unrelated later background dispatch re-broadcasts full state.
</details>

Comp0te (commented):
Reviewed the move of vault unlock into the background — the new privileged port and its allowlist, the scrypt offload and worker, the memo, and the three page-side response handlers that replaced the page-local crypto. The comments below share a through-line: the secret and the failure modes both moved into the background process, and the guarantees that used to be page-local — a Worker that got `terminate()`d, an error path the page owned — need equivalents on the new side of the boundary.

One point has no line in this diff to sit on, so it goes here. None of `ci-check.yml`, `e2e-popup-tests.yml` or `e2e-onboarding-tests.yml` triggers on this PR: all three filter `pull_request` on `[master, develop, release/**]` and this PR's base is `WALLET-1424-cw-privileged-port`, so `gh pr checks 1498` returns only CodeQL. The two specs that pin the flows this PR rewrites — `e2e-tests/popup/common/wallet.spec.ts:6-27` (unlock) and `:85-106` (five wrong passwords to lockout) — are byte-identical to the base branch and have not run against this change anywhere. Worth running them manually against this head, or adding the stack's base to the workflow filters for the life of the stack.

src/background/workers/scrypt-worker.ts
```diff
@@ -0,0 +1,30 @@
...
+  const { password, saltHash } = event.data;
+
+  try {
+    const key = await scryptAsync(
+      password,
+      convertHexToBytes(saltHash),
+      createScryptOptions()
+    );
+
+    postMessage({ key });
```
Comp0te, line 23: Six behaviours this PR adds are unconstrained by the suite: in five, I mutated the production code and the jest run stayed green; the sixth has no test at any level. This line is the clearest — `scrypt-worker.ts` is executed by no test, so the two sides of the worker message contract are never checked against each other.

Each is cheap to pin, and they are one pass rather than six.

<details><summary>The six, and the mutation that survived each</summary>

1. **`scrypt-worker.ts:23`** — `postMessage({ key })` → `postMessage(key)`; `scrypt-off-thread.test.ts` stayed 7/7. `isWorkerError` (`workers/types.ts:7-10`) then returns false for a bare `Uint8Array`, `event.data.key` is `undefined` at `scrypt-off-thread.ts:49`, and `convertBytesToHex` throws — breaking unlock, verify, onboarding `initKeysSage` and change-password together on MV2. The offload tests drive a hand-written fake (`scrypt-off-thread.test.ts:44-51`), the only importer of the real module is `spawn-scrypt-worker.ts:5`, and `jest.config.js:48-60` `collectCoverageFrom` excludes `src/background/workers/`.
2. **`unlock-requests.ts:130-137`** — appending `.catch(() => false)` to the `verifyPasswordOffThread` call stayed 22/22, and a failed derivation is then counted as a wrong password. The invariant is still written down page-side at `password-protection-page/index.tsx:81-83`, but only for that page's transport and `error` handling; nothing states or tests it for the background derivation, which is where the increment moved.
3. **`unlock-requests.test.ts:450-459`** — `await expect(Promise.all(results)).resolves.toBeDefined()` holds for every implementation of `remember`, including no eviction and including evicting the newest; with the `MEMO_LIMIT` guard at `:95` disabled the suite stayed 22/22. That guard is the only bound the memo has (see the comment on `:45`). Asserting *which* key survives would pin it.
4. **`privileged-port.ts:28-38`** — removing `SIGNATURE_REQUEST_PAGE` from the UNLOCK allowlist stayed 20/20. Of the eight allowlist entries exactly one is positively asserted (`privileged-port.test.ts:252-267`); the other two routing tests are negative and no signature-request sender is defined in the file. Losing that entry means a locked user who clicks Sign in a dapp can never unlock — refusal disconnects (`:145-148`), the retries at +250/+500ms reject, and the generic error is all they see.
5. **`unlock-requests.ts:188-190`** — replacing the `finally` body stayed 22/22, because `unlock-requests.test.ts:23-25` stubs `anchorServiceWorker` wholesale and `release()` is unobservable. A leaked anchor leaves `anchorCount` above zero, so `clearInterval` never runs and the 20s `getPlatformInfo()` heartbeat (`sw-keep-alive-anchor.ts:52-75`) outlives every Chrome user's first unlock.
6. **`unlock-vault/index.tsx:82-114`, `password-protection-page/index.tsx:65-121`, `form-validation-rules.ts:43-77`** — the four-branch response handler and its transport `catch` have no test at any level; no test file references these three pages, and the popup e2e reaches only the broadcast-delivered path, not the `lockedOut` fallback at `unlock-vault/index.tsx:96-105` or the `catch` at `:107-111`.
</details>

ost-ptk:
On the CI point: confirmed, and run locally against this head instead.

`ci-check.yml`, `e2e-popup-tests.yml` and `e2e-onboarding-tests.yml` all filter `pull_request` on `[master, develop, release/**]`, so with a stack base only CodeQL runs here. Results against this branch:

| Suite | Result |
|---|---|
| `popup/common/wallet.spec.ts` + `settings/backup-secret-phrase.spec.ts` | 6 passed |
| full `e2e-tests/popup` | 56 passed, 2 flaky |
| the 2 flaky, re-run with `--retries=0` | 6 passed |
| `e2e-tests/onboarding-flow` | 8 passed, 1 flaky |
| unit + tsc + lint + prettier + knip | 117 suites / 1172 tests, all clean |

The two flaky are `should made a failed transfer` in the native and ERC-20 transfer specs — green on a clean re-run, and nothing on this branch touches transfers. The onboarding flake is `dropped-dispatch-flow.spec.ts` failing with "Service worker did not register in time"; the `extensionId` fixture waits 5s for the SW and occasionally loses that race on a cold start. Both predate this stack.

Worth noting for `change-password.spec.ts`: it passes, which is the check that mattered after the page stopped navigating to Home before the ack lands.

We're leaving the workflow filters alone — adding the stack bases would be a change to shared config for the life of one ticket. One caveat if the stack is merged with `gh stack merge`, though: that merges bottom-to-top atomically, so the upper PRs never sit on a `develop` base long enough for CI to fire. Merging them one at a time gives each PR a real CI run, because GitHub retargets the next PR's base to `develop` as soon as the one below lands.

src/background/handlers/unlock-requests.ts
ost-ptk: Took the digest option, in `a62133b5`. `MemoEntry` holds an HMAC of the password instead of the password, compared with `constantTimeEqualHex` — comparison is all `:230` ever needed, so no behaviour changes and residency stops mattering.

One thing worth spelling out, because it decides between HMAC and a plain SHA-256: a bare hash would be a brute-force oracle *cheaper* than the scrypt-derived cipher it guards. So the digest is keyed on a secret generated per process and never persisted, and a test pins that the same password digests differently in a fresh process — a "simplification" back to `sha256(password)` fails it.

That also means neither of your fallbacks was needed: no `memo.clear()` in `lockVaultSaga` (which would have required exporting the map) and no per-entry timer.

Your `:39-42` observation was the load-bearing one — the code already conceded that the TTL bounds replayability rather than residency, and nobody had followed that to its conclusion. The comment now says the entry holds a digest *because* of it.

src/background/workers/scrypt-off-thread.ts
ost-ptk: Fixed in `e3f3db1c`. The worker promise now has a 30s deadline that rejects, terminates the worker and frees the queue slot.

Your reading of `serialise` is exactly right, and it is the part that makes this severe: `derivationQueue = next.catch(...)` handles a rejection but cannot displace a promise that stays pending, so one wedged derivation was permanent until the background restarted — a browser restart on MV2.

The value sits above a real derivation (~350ms measured here) and below the port's own 60s, so the caller gets a derivation failure rather than the transport timeout that `background-port.ts:77-79` re-throws without retrying.

Two tests, not one: the first would be satisfied by any deadline including 0, so a second pins that a merely slow derivation is not failed. Verified by setting the constant to 0 — only that test fails.

Whether a Worker can die without firing `onerror` is still unsettled, as you say. The deadline makes it moot either way.

src/background/handlers/unlock-requests.ts
ost-ptk: Fixed in `5f19655b`, separately from the digest.

Worth flagging: the digest change did **not** subsume this, contrary to what we both expected. `remember` is still called immediately after `serialise`, so the window still started at enqueue. The entry is now re-stamped when the verdict settles, on both paths — the window measures when an answer became available, not what the answer was.

Your point that only a wrong password can double-count is right, and it is why the test drives a wrong-password request queued behind a slow one: without the fix the retry re-derives and produces a second `loginRetryCountIncremented`.

The existing expiry test at `:166-186` had to move to a controlled clock — it pinned the old semantics through a `mockReturnValueOnce` sequence, and the window it describes is now measured from the verdict.

src/libs/layout/unlock-vault/index.tsx
ost-ptk: Took the re-broadcast option, in `031c3b86` — it heals the loop rather than papering over it.

The already-unlocked path now broadcasts state before answering `ok`, so the retry that lands there recovers the page instead of confirming the dead end.

Your trace of why nothing else self-heals was the deciding part: `useUserActivityTracker` attaches only when unlocked, and `unlockVaultSaga`'s puts are upstream of the dropped broadcast. Remounting was indeed the only route, and nothing told the user.

`broadcastToReplicas` moved out of `get-main-store` into its own module so the subscriber and this handler share one definition instead of assembling the same event twice.

Slightly embarrassing detail: the `lockedOut` branch right above already carried a comment about `broadcastToReplicas` swallowing failures. `ok` was left assuming.

src/background/workers/scrypt-worker.ts
ost-ptk: Five of the six are fixed in `4b997d52`; the sixth is deliberately left.

1. `scrypt-worker.ts` now runs for real — the test installs global `postMessage`/`onmessage` and drives it. Both envelope mutations fail: `postMessage(key)` breaks two tests, and breaking the error envelope breaks the third. It also asserts the key matches what `encodePassword` produces, so the two sides of the contract are checked against each other.
2. A derivation failure is now asserted **not** to increment. `.catch(() => false)` fails.
3. The eviction test asserts which key survived — the oldest re-derives, the newest replays. Disabling the `MEMO_LIMIT` guard fails.
4. `ALLOWED_PAGES` is pinned whole plus positive per-page assertions. Removing `SIGNATURE_REQUEST_PAGE` from UNLOCK fails two tests.
5. The anchor mock is observable and `release()` is asserted. Replacing the `finally` body fails.

Each mutation was applied and reverted to confirm.

The sixth — the three page-side response handlers — is not in this pass. `testEnvironment` is node, so they need the SSR harness or e2e rather than a unit test, and that is a bigger change than the rest combined. Noted in the commit message so it does not read as an oversight.

A detail found on the way, in case it bites someone: `convertHexToBytes` truncates invalid hex rather than throwing, so a bad salt cannot be used to force the worker's failure path.

Comp0te (dismissed):

Comp0te (approved):

PR #1505, opened 2026-08-21
fix(background): rank tombstone eviction by ordinal, not key order

ost-ptk:
## Description

`windowRequestResponded` evicted the oldest tombstones by filtering `Object.keys(requests)`, on the stated assumption that *"Insertion order = the order requests were first registered"*.

That is false for the keys this map actually holds. `requestId` is page-generated (`generateRequestId`, `src/content/sdk.ts`), so an integer-like id such as `"42"` is one the wallet accepts — and integer-like keys enumerate ahead of every string key in ascending numeric order, however recently they were written. A dapp numbering its requests could therefore have its **newest** tombstone evicted first while genuinely older ones survived, and `MAX_RESPONDED_TOMBSTONES` stopped bounding what it was meant to bound. Losing a tombstone early means the response dedup no longer drops a late duplicate for that id.

### What changed

- Both arms of `Request` carry a `seq`. The open row is stamped at registration; the tombstone is **restamped at response time** (`nextSeq` over the pre-update map), because the cap's stated policy is eviction by when a request was *answered* — inheriting the registration ordinal would let answering the oldest-registered request evict its own just-written tombstone (review finding, pinned by a dedicated test).
- Eviction ranks on `seq` ascending instead of on key order.
- The ordinal is **derived from the map** (`max + 1`) rather than held in a counter field. A counter would be state that survives nothing: after an MV3 service-worker restart it would resume at zero and re-issue ordinals the map still holds.

### Why not reuse the vault's helpers verbatim

`vault/reducer.ts` solved the same problem for its payload map (WALLET-1418), but its `orderOldestFirst` additionally handles rows written *before* the field existed, ranking unstamped hoisted keys newest. Nothing here can be unstamped — every row is stamped by this reducer at birth — so those branches are not ported. On a file held to 100% coverage they would be unreachable code demanding tests.

The helpers were also deliberately **not** moved into `request-map.ts`: `collectCoverageFrom` matches `src/background/redux/**/reducer.ts` but not `request-map.ts`, so relocating them would silently drop covered code out of the enforced universe for both slices.

### Verification

- `npx jest src/background/redux/windowManagement/ src/background/handlers/ src/background/open-window.test.ts src/background/redux/sagas/vault-sagas.test.ts` — 22 suites, 394 tests pass.
- `npx tsc --noEmit` — clean.
- `windowManagement/reducer.ts` coverage: 100% statements / branches / functions / lines.
- The regression test was proven to fail against the old ranking: with `Object.keys` restored, `"42"` is evicted while the genuinely older `r0` survives.

### Context

This is a prerequisite for the rest of WALLET-1419. The planned fix mirrors `windowManagement`'s request state into `chrome.storage.session` so it survives a service-worker restart; that mirror round-trips through a JSON-shaped structure, which re-applies the same key hoisting, and its write-side row cap needs an age order to drop by. The ordinal has to exist first.

The eviction bug is live on `develop` independently of that work, so this stands on its own.

## Linked tickets

## Checklist

### DIFF
--- a/src/background/redux/windowManagement/reducer.ts
+++ b/src/background/redux/windowManagement/reducer.ts
@@ -9,6 +9,17 @@ import { CancellableMethod, WindowManagementState } from './types';
 // background page never restarts.
 export const MAX_RESPONDED_TOMBSTONES = 50;
 
+// Derived from the map rather than kept in a counter field: a counter is state
+// the session record does not carry, so after a service-worker restart it would
+// resume at zero and re-issue ordinals the restored map still holds.
+const nextSeq = (requests: WindowManagementState['requests']): number => {
+  const stamped = Object.values(requests).flatMap(request =>
+    request == null ? [] : [request.seq]
+  );
+
+  return stamped.length === 0 ? 0 : Math.max(...stamped) + 1;
+};
+
 const initialState: WindowManagementState = {
   windowId: null,
   exportKeysWindowId: null,
@@ -68,7 +79,8 @@ const slice = createSlice({
             origin: action.payload.origin,
             method: action.payload.method,
             windowIds: [],
-            awaitingDeviceConfirmation: false
+            awaitingDeviceConfirmation: false,
+            seq: nextSeq(state.requests)
           }
         }
       };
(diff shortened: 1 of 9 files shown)


src/background/redux/windowManagement/reducer.ts
```diff
@@ -179,22 +191,26 @@ const slice = createSlice({
...
+      const request = getRequest(state.requests, action.payload.requestId);
+
+      if (request?.status !== 'open') {
         return state;
       }
 
       const requests: WindowManagementState['requests'] = {
         ...state.requests,
-        [action.payload.requestId]: { status: 'responded' }
+        [action.payload.requestId]: { status: 'responded', seq: request.seq }
```
Comp0te: `seq: request.seq` gives the tombstone its *registration* ordinal, but eviction sorts ascending and drops the lowest — so answering the oldest-registered request deletes the tombstone this dispatch just wrote, losing response dedup for that id. Not a regression: base picks the same victim. But this PR re-implements the ranking, and its own comments frame it by *when answered*.

Consider `seq: nextSeq(state.requests)`.

`reducer.ts:200-219` writes the responded entry with the open descriptor's `seq`, then sorts ascending and deletes `slice(0, overflow)` — nothing exempts the entry written in this same call. To reproduce: register a request and leave it open (a Ledger request awaiting the device is withheld from window reuse at `open-window.ts:64-69`, so it stays open indefinitely), run 50 register/respond cycles — the supersede path writes those tombstones with no user interaction — then respond to the first: it is `undefined`, while `r0` survives at `seq: 1`.

`selectRequestStatus` then returns `undefined`, so `sdk-response-to-tab.ts:135-168` stops dropping later responses for that id and one reaches `tabs.sendMessage` at `:297`, where `content/sdk.ts:97-110` settles on whichever lands first. A second send is reachable: `sign-transaction/index.tsx:295` and `:315-321` are unguarded in one component instance.

The policy the ranking misses is stated in this PR — `reducer.test.ts:435-436` and `reducer.ts:6-9` both describe the cap by *when a request was answered*. `nextSeq` is `max + 1` over the pre-update map, so monotonicity and "a re-sent id cannot promote itself" both survive; only the literal `seq` assertions change.

</details>

Comp0te (commented):
Reviewed the ordinal-based tombstone eviction and the fixture updates that follow it. The ranking change itself is right — enumeration order was genuinely hoisting integer-like keys — and the tests pin that. One comment below, on what the ordinal is measured from.

src/background/redux/windowManagement/reducer.ts
ost-ptk: Accepted and fixed in 1305f611: the tombstone is now restamped at response time (`seq: nextSeq(state.requests)`), so eviction genuinely ranks by when a request was answered. Your reproduction is pinned as a dedicated test (registered-first-answered-last survives eviction; the earlier-answered tombstone is the evicted one), and the cascading literal expectations were re-derived.

Comp0te (approved):

PR #1518, opened 2026-08-25
docs(background): reframe MV3-restart fallbacks as residual paths

ost-ptk:
## Description

Comment and test-name work only — `git diff` contains no executable-code changes.

The session mirror (previous PR in this stack) makes an MV3 service-worker restart no longer empty `windowManagement.requests` in the ordinary case on Chrome/Edge. Several comments still described that restart as the _primary_ reason a request descriptor can be missing while its approval window is on screen. After the mirror it is a residual reason: a mirror write lost in flight, a row the hydration sanitizer dropped, the deliberate remove-on-rejected-write, an open row dropped by the write cap during a burst, or an approval window surviving a wallet reset. (Firefox and Safari declare `"persistent": true` and never had the restart problem; the mirror is deliberately not active there.)

### What changed

- `sdk-response-to-tab.ts` — the comment above the `request?.origin ?? recoverDappOrigin(sender.url)` fallback now frames it as covering the residual descriptor-less paths. **The fallback itself is kept** — it is load-bearing for exactly those residues.
- `deliver-via-origin.ts` — same rewording for the unscoped-broadcast note.
- `windowManagement/reducer.ts` — two blocks: the tombstone-cap rationale (the Firefox/Safari growth story that justifies the FIFO cap is unchanged and stays; only the "a restart wipes the map" half is retired) and the `windowRequestResponded` transition-guard trigger.
- `vault-sagas.ts` — the saga-body justification for the window-URL half of `reconcileStalePayloadsSaga`'s keep set now rests on the residual paths, not on restarts. The `reconcileStalePayloadsSaga` doc block is deliberately untouched: its stranding story concerns the vault's own payload map (`jsonById`/`eip712ById`), which is not mirrored, and nothing in it became stale.
- Test comments/names that documented the old framing were re-framed onto the state they actually set up ("no descriptor registered"), in `sdk-response-to-tab.test.ts`, `reducer.test.ts`, and `close-windows-on-response.test.ts`. **Every assertion is unchanged.**

### Verification

- `npx jest src/background/handlers/ src/background/redux/windowManagement/ src/background/redux/sagas/` — 29 suites, 537 tests pass (count unchanged from base).
- `npx tsc --noEmit`, eslint and prettier clean on all touched files.
- Zero-behaviour-change reviewed hunk by hunk: every changed line is a comment line or a test-name string.

## Linked tickets

## Checklist

- [x] Include a screenshot or recording if implementing significant UI or user flow change — comment-only, no UI change

### DIFF
--- a/src/background/handlers/close-windows-on-response.test.ts
+++ b/src/background/handlers/close-windows-on-response.test.ts
@@ -537,7 +537,7 @@ describe('handleSdkResponseToTab (WALLET-1416 wiring)', () => {
     expect(removeMock).not.toHaveBeenCalled();
   });
 
-  it('after an MV3 restart it delivers and closes nothing', async () => {
+  it('with no descriptor ever registered it delivers and closes nothing', async () => {
     const store = makeRealStore();
     store.dispatch(
       ledgerNewWindowIdChanged({
--- a/src/background/handlers/deliver-via-origin.ts
+++ b/src/background/handlers/deliver-via-origin.ts
@@ -23,7 +23,8 @@ export async function deliverViaOrigin(
   // wrong one is both a leak and a loss. Only frame 0 is the same frame in
   // every tab, and the origin match already established that tab's top
   // document belongs to the dapp — so a non-zero frameId must not broadcast.
-  // `undefined` (no descriptor — MV3-restart path) keeps broadcasting unscoped.
+  // `undefined` (no descriptor — one of the residual descriptor-less paths)
+  // keeps broadcasting unscoped.
   if (frameId != null && frameId !== 0) {
     console.error(
       'deliverViaOrigin: sub-frame response not broadcast — no other tab can be matched to it',
--- a/src/background/handlers/sdk-response-to-tab.test.ts
+++ b/src/background/handlers/sdk-response-to-tab.test.ts
@@ -686,11 +686,11 @@ describe('handleSdkResponseToTab (background dedupe of SDK responses)', () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
-  it('no descriptor (service worker restarted) → still delivers, dispatches nothing', async () => {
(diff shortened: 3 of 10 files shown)


Comp0te (commented):
Read the rewritten comments against the code each one describes. The reframing holds up — the mirror does survive a worker restart on Chrome/Edge, the residual paths named are real, and the `?origin=` fallback is correctly kept as load-bearing.

One point, on the sweep's reach rather than its content. Three sites still assert the old framing in the present tense at this head. The sharpest is `src/background/redux/sagas/vault-sagas.test.ts:743-745` — "`windowManagement.requests` is not persisted" — the test covering the very saga comment you rewrote at `vault-sagas.ts:519`, and false on Chrome/Edge. Worth folding into the same pass.

<details><summary>The three sites</summary>

- `sagas/vault-sagas.test.ts:743-745` — "…is not persisted, so after a service-worker restart the descriptor is gone"; contradicted by the mirror write at `get-main-store.ts:204` and the flush at `session-store.ts:273-288`.
- `get-main-store.ts:135-137` — "lives only in the worker's memory, and an MV3 restart destroys it", four lines above `requests: requestSession.requests` (`:141`), the `readRequestSession()` output it says cannot exist.
- `vault/reducer.ts:729-732` — "the orphan case (an MV3 restart between registration and the response)", the framing you replaced at `windowManagement/reducer.ts:191`.

Uniform staleness was recoverable; a reader could discount all of it. A mixture is not.
</details>

Comp0te (dismissed):

ost-ptk:
The three remaining present-tense sites from the review are folded in (20faf459, amended into this PR's commit): `vault-sagas.test.ts` and `vault/reducer.ts` now use the residual descriptor-less framing, and `get-main-store.ts`'s comment above the hydration describes the Chrome/Edge mirror vs the Firefox/Safari persistent page instead of contradicting the line below it.

Comp0te (dismissed):

Comp0te (approved):

PR #1519, opened 2026-08-25
fix(background): sweep orphaned request rows after a worker restart

ost-ptk:
## Description

Hardening on top of the session mirror. Durable state is not a durable event: `windows.onRemoved` fires once per window, and two states can freeze into the mirror with no event left to advance them — an init failure consumes the only `onRemoved` while the mirror re-hydrates the row, and a crash inside the 250 ms cancel grace durably persists `{status: 'open', windowIds: []}`, the one shape no window event can ever cancel. A pinned row's id is rejected by the SDK entry guard forever and holds one of the ten `MAX_STORED_PAYLOADS` slots — reopening the leak WALLET-1418 closed.

### The sweep

New module `src/background/handlers/sweep-orphaned-requests.ts`, called fire-and-forget from `getExistingMainStoreSingletonOrInit` after the subscriber attaches (no new `await` between the singleton's null check and assignment). It takes the hydrated snapshot as an argument — never `getState()` after an await — so a request registered after init can never be swept during its own registration→attach gap.

- **Liveness is a window-URL question, never a `windowIds` question.** A hydrated `'open'` row is orphaned iff the enumeration succeeded and no open window's URL claims its id. Empty `windowIds` is the registration→attach shape — the window is on screen — and is never treated as proof of death. A failed enumeration cancels nothing.
- `collectRequestIdsFromOpenWindows` is called **unchanged** — same superset-of-live, fail-toward-keeping polarity `reconcileStalePayloadsSaga` depends on.
- Short-circuits before `windows.getAll` when the snapshot has no `'open'` rows — init runs on every wake, and a user who never touches a dapp must not pay a tab enumeration per wake. Work per wake is bounded to the mirror's write cap.
- `await delay(CANCEL_GRACE_MS)` before cancelling: a Ledger confirmation runs in the window's document, not the worker, so a genuine signed response can be in flight; the vehicle's live re-check then sees `'responded'` and the sweep sends nothing.

### The vehicle, parameterised

`failRequestOnWindowError` gains a `source: SagaErrorSource` parameter (new `'sweep-orphaned-requests'` member of the closed union; `CancelSource`'s comment rewritten):

- **No banner for swept rows.** The `sagaError` dispatch fires only for the `open-window-failed` caller. In close-as-wake the sweep's enumeration resolves inside the grace and delivers the flagship cancel — without this, every ordinary close-after-idle would paint an untranslated error banner, possibly over a live signing prompt. The sweep's policy is console-only, matching the codebase precedent for dapp-triggerable paths.
- **#1484's origin check moved inside the vehicle.** On a navigated-away tab `tabs.sendMessage` _succeeds_ (the content script is on every http(s) page), so a catch-fallback never fires — and a swept row can be arbitrarily old, so navigated-away is the expected state. The check carries #1484's frameId semantics: the origin comparison is gated on `frameId == null || frameId === 0` (`tabs.get` reports the top document's origin) and the send is frame-targeted. **This intentionally changes the existing `open-window-failed` call site's delivery too** — its cancels now get the same navigated-away protection. The cost, stated honestly: the tombstone and the send are now separated by one extra `tabs.get` round-trip for both sources, slightly widening the spec's acknowledged "crash between tombstone and delivery loses that cancel" residual.

### Idempotence

The `getState → filter open → dispatch tombstone` sequence stays one synchronous block — the new origin `await` sits _after_ the tombstone dispatch — so concurrent sweep + window-removed cancels for the same row deliver exactly one cancel in every interleaving (pinned by a test running the real reducer, asserting a single `tabs.sendMessage`).

### What it does not close

A crash between the tombstone write and delivery still loses that cancel permanently: the next wake hydrates `'responded'`, the sweep skips it, and nothing retries. Bounded-staleness by design.

### Verification

- `npx jest src/background/` — 69 suites, 911 tests pass; `npx tsc --noEmit` clean; eslint/prettier clean (the type-only `MainStore` imports use `import type`, so no `import/no-cycle` warnings).
- New-module coverage 100/100/100/100; `src/background/handlers/` floors (85/100/95/95) hold.
- The nine sweep scenarios from the spec are all pinned, including: still-claimed row with empty `windowIds` untouched; genuine response during the grace wins; concurrent sweep + close delivers one cancel; navigated-away tab routed through `deliverViaOrigin`.
- Logging: new logs pass through `redactUrlQuery` only; tests assert no `?…=` substring survives any log the new code emits.

**Manual smoke still owed before merge** (needs a live extension): kill the worker inside the 250 ms grace after closing an approval window, wake it, confirm the pinned row is cancelled and its payload slot reclaimed.

## Linked tickets

## Checklist

### DIFF
--- a/src/background/handlers/cancel-requests.ts
+++ b/src/background/handlers/cancel-requests.ts
@@ -202,8 +202,11 @@ export async function cancelRequestsDisplacedBy(
 // (spec §8.1). Without this the dapp promise hangs until its own timeout
 // (30 min by default).
 //
-// `source` defaults to the original trigger so every existing call site keeps
-// its behaviour unchanged. It doubles as the banner policy: only
+// `source` defaults to the original trigger so the `open-window-failed` call
+// site keeps its BANNER policy unchanged — its delivery does not: #1484's
+// stale-tab check below applies to every source alike, so that call site's
+// delivery is deliberately narrowed too, same as the sweep's. `source` doubles
+// as the banner policy: only
 // `'open-window-failed'` dispatches `sagaError` (a `windows.create` failure is
 // the wallet's own doing, with nothing else to tell the user). Every other
 // source — today just the sweep — is dapp-triggerable and console-only,
@@ -238,14 +241,21 @@ export async function failRequestOnWindowError(
   // the TOP document's origin, so only a top-frame request (no `frameId`, or
   // `0`) is checkable this way; a sub-frame request goes straight to the
   // frame-targeted send, same as today.
-  const staleOrigin =
-    (frameId == null || frameId === 0) &&
-    (await getLiveTabOrigin(tabId)) !== origin;
+  const isTopFrame = frameId == null || frameId === 0;
+  const liveOrigin = isTopFrame ? await getLiveTabOrigin(tabId) : undefined;
+  const staleOrigin = isTopFrame && liveOrigin !== origin;
 
   let delivered = 1;
 
   if (staleOrigin) {
     delivered = await deliverViaOrigin(origin, action, frameId);
+
+    // Identifiers and origins only, matching sdk-response-to-tab's withheld-
+    // response log — never a URL.
+    console.error(
+      `${source}: target tab no longer hosts the requesting origin; response withheld`,
+      { requestId, tabId, expectedOrigin: origin, liveOrigin, delivered }
+    );
   } else {
     try {
       await (frameId == null
@@ -265,6 +275,9 @@ export async function failRequestOnWindowError(
     }
   }
 
+  // The sweep (source === 'sweep-orphaned-requests') knowingly shares this
+  // same tombstone-before-delivery ordering — an accepted residual, not an
+  // oversight specific to the sweep.
   if (source !== 'open-window-failed') {
     console[delivered > 0 ? 'warn' : 'error'](
       `${source}: cancelled an orphaned request`,
--- a/src/background/handlers/cancel-requests.ts
+++ b/src/background/handlers/cancel-requests.ts
@@ -1,8 +1,9 @@
 import { tabs } from 'webextension-polyfill';
 
+import { redactUrlQuery } from '@background/redact-url-query';
 import { sagaError } from '@background/redux/app-events/actions';
 import { SagaErrorSource } from '@background/redux/app-events/types';
-import { MainStore } from '@background/redux/get-main-store';
+import type { MainStore } from '@background/redux/get-main-store';
 import {
   windowDetachedFromRequests,
   windowRequestResponded
@@ -16,6 +17,7 @@ import {
 import { SdkMethod, sdkMethod } from '@content/sdk-method';
 
 import { deliverViaOrigin } from './deliver-via-origin';
+import { getLiveTabOrigin } from './tab-origin';
 
 // Grace before cancelling an abandoned request: lets an in-flight genuine
 // response (a button response, or a Ledger success cleanup that closes this
@@ -28,9 +30,10 @@ const delay = (ms: number) =>
 // The window-driven cancel paths only. Derived from `SagaErrorSource` rather
 // than spelled out again, so the two can never drift — a `source` this module
 // dispatches is by construction one the banner reader knows.
-// `'open-window-failed'` is deliberately NOT here: it is produced by
-// `failRequestOnWindowError`, which is a separate trigger with no grace and no
-// window event behind it.
+// Neither `'open-window-failed'` nor `'sweep-orphaned-requests'` is here: both
+// are produced by `failRequestOnWindowError`, a separate trigger with no
+// window event behind it — and only the first of the two banners at all (see
+// its `source` parameter below).
 export type CancelSource = Extract<
   SagaErrorSource,
   'cancel-on-close' | 'cancel-on-supersede'
@@ -193,12 +196,26 @@ export async function cancelRequestsDisplacedBy(
   await cancelRequests(store, candidates, source, windowId, afterMark);
 }
 
-// `windows.create` rejected, so no window will ever display this request and no
-// `windows.onRemoved` will ever fire for it. Without this the dapp promise hangs
-// until its own timeout (30 min by default).
+// The trigger with no window event behind it — either `windows.create`
+// rejected (no `windows.onRemoved` will ever fire for a window that never
+// existed) or the startup sweep decided a hydrated 'open' row is orphaned
+// (spec §8.1). Without this the dapp promise hangs until its own timeout
+// (30 min by default).
+//
+// `source` defaults to the original trigger so every existing call site keeps
+// its behaviour unchanged. It doubles as the banner policy: only
+// `'open-window-failed'` dispatches `sagaError` (a `windows.create` failure is
+// the wallet's own doing, with nothing else to tell the user). Every other
+// source — today just the sweep — is dapp-triggerable and console-only,
+// matching the precedent in `sdk-methods.ts`'s `reportCapacityRefusal`: a
+// banner mounted route-independently over every approval screen must not
+// fire for a request the user cannot act on, and in close-as-wake the sweep's
+// enumeration resolves inside another cancel's own grace, so it would often
+// paint over an ordinary close or a live signing prompt.
 export async function failRequestOnWindowError(
   store: MainStore,
-  requestId: string
+  requestId: string,
+  source: SagaErrorSource = 'open-window-failed'
 ): Promise<void> {
   const request = selectOpenRequests(store.getState()).find(
     openRequest => openRequest.requestId === requestId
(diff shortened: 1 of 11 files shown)


Comp0te (commented):
Reviewed the startup sweep and the reworked cancel-delivery path. The through-line of these comments is the seam between the two: the tombstone is written before delivery and never revisited, the origin gate withholds without recording that it did, and the new tests around both stay green through the mutations they are named for. One prose comment also states the opposite of what the diff did to the pre-existing call site.

src/background/handlers/cancel-requests.ts
```diff
@@ -210,17 +227,50 @@ export async function failRequestOnWindowError(
...
+        requestId,
+        method,
+        tabId,
+        error: redactUrlQuery(error)
+      });
+      delivered = await deliverViaOrigin(origin, action, frameId);
+    }
+  }
+
+  if (source !== 'open-window-failed') {
```
Comp0te, line 281: The tombstone is dispatched at :228, before any delivery, and this branch keeps it even when `delivered === 0`. Hydration preserves `'responded'` and the sweep collects only `'open'`, so a swept row whose cancel reached nobody is never retried — the dapp hangs to its own timeout.

Is a terminal cancel right here, given a swept row gets another wake?

Confirmed at `cancel-requests.ts:228` and `:268-274`; `get-main-store.ts:198-208` mirrors the tombstone to `storage.session`, `session-store.ts:109-111` preserves it through hydration, `sweep-orphaned-requests.ts:30-35` collects only `status === 'open'`. Unlike a `windows.create` rejection — where delivery failure really is terminal — a swept row is re-hydrated on every later wake and skipped.

Two constraints on a fix. Re-opening the row is not available: `windowManagement/reducer.ts:199-202` replaces it with `{ status: 'responded', seq }`, so `tabId`, `origin`, `method` and `frameId` are gone. And moving the tombstone after delivery re-opens the double-delivery race that `sdk-response-to-tab.ts:228-242` documents as the reason it marks optimistically.

</details>

src/background/handlers/cancel-requests.ts
```diff
@@ -210,17 +227,50 @@ export async function failRequestOnWindowError(
...
+
+  // #1484: verify the tab still hosts the requesting origin BEFORE sending.
+  // On a navigated-away tab `tabs.sendMessage` SUCCEEDS (the content script is
+  // injected on every http(s) page), so the catch → `deliverViaOrigin`
+  // fallback below would never fire — a swept row can be arbitrarily old, so
+  // navigated-away is the expected case, not the exception. `tabs.get` reports
+  // the TOP document's origin, so only a top-frame request (no `frameId`, or
+  // `0`) is checkable this way; a sub-frame request goes straight to the
+  // frame-targeted send, same as today.
+  const staleOrigin =
```
Comp0te: The withhold decision is right and deliberately fail-closed — but silent. When `staleOrigin` is true the direct send is skipped with nothing logged, so for `source === 'open-window-failed'` with `delivered === 0` the event leaves no console trace naming the request or the tab.

Log it with `{ requestId, tabId, expectedOrigin, liveOrigin, delivered }`, as `sdk-response-to-tab.ts:269-279` does.

Fail-closed on an unresolvable origin is the documented contract (`tab-origin.ts:3-7`) and this PR pins it at `fail-request-on-window-error.test.ts:155-164` — that part reads as intended.

What is missing is the record. `deliver-via-origin.ts:28-49` logs only on throw and on the sub-frame refusal, neither of which applies when `frameId` is `null`/`0`; `utils.ts:60-82` logs per-tab send failures, not "no tab matched". The user's only artifact is the `sagaError` at `:281-289`, which names no `requestId` and no `tabId`. On the sweep path `:269-272` reports `delivered: 0` without saying the direct send was withheld, or why.

</details>

src/background/handlers/cancel-requests.ts
```diff
@@ -193,12 +196,26 @@ export async function cancelRequestsDisplacedBy(
...
-// `windows.create` rejected, so no window will ever display this request and no
-// `windows.onRemoved` will ever fire for it. Without this the dapp promise hangs
-// until its own timeout (30 min by default).
+// The trigger with no window event behind it — either `windows.create`
+// rejected (no `windows.onRemoved` will ever fire for a window that never
+// existed) or the startup sweep decided a hydrated 'open' row is orphaned
+// (spec §8.1). Without this the dapp promise hangs until its own timeout
+// (30 min by default).
+//
+// `source` defaults to the original trigger so every existing call site keeps
```
Comp0te: This says every existing call site keeps its behaviour unchanged, but the same hunk changed delivery for the one non-test caller, `open-window.ts:53`. Base did an unconditional `tabs.sendMessage(request.tabId, action)`; head checks the live origin first and, on mismatch or unresolvable, routes via `deliverViaOrigin` instead.

The PR body says the change is deliberate — worth saying it here too.

Sole non-test caller confirmed by grepping both trees in full: `open-window.ts:53` calls `failRequestOnWindowError(store, requestId)` with no `source`, at base and head alike. Base body at `cancel-requests.ts:213-224`, head at `:241-266`.

The divergence is user-visible. A navigated-away tab's content script accepts the base send (delivered = 1 → "the request was cancelled"), while the head path can return 0 → "the site could not be told; the request may still be pending there". Same input, different banner — and this comment sends whoever triages that report away from the routing decision three lines below.

Suggested wording: the default preserves the *banner policy* for that call site, while its delivery is deliberately narrowed to match #1484.

</details>

src/background/handlers/sweep-orphaned-requests.ts
```diff
@@ -0,0 +1,106 @@
...
+    return;
+  }
+
+  // Unchanged, on purpose: a read-only second caller cannot narrow the Set
+  // `reconcileStalePayloadsSaga` consumes, and a second copy would have to
+  // independently re-earn its piecewise URL compare, `REQUEST_BEARING_PATHNAMES`,
+  // `tab.url || tab.pendingUrl`, the per-tab catch and the `Set | null`
+  // contract — then split-brain the moment a new approval page is added. See
+  // spec §8.1.
+  const liveRequestIds = await collectRequestIdsFromOpenWindows();
```
Comp0te, line 50: Five of the sweep's safety properties survive the mutation each is named for. Sharpest is this line's oracle: nothing ties `REQUEST_BEARING_PATHNAMES` to the URLs approval windows open at, so a renamed or added approval page makes the sweep cancel a request whose window is on screen.

Worth closing at least the oracle and the two blind assertions.

<details><summary>The five</summary>

1. **The oracle (this line)** — `open-request-windows.ts:8-12` hardcodes three pathnames; `open-request-windows.test.ts:182` restates the same literals and `:203` a hardcoded negative list. Nothing derives from `getUrlByWindowApp` (`create-open-window.ts:17-45`), and the sweep's tests mock the oracle wholesale (`sweep-orphaned-requests.test.ts:23-25`). The hazard is named at `:44-49` above and unpinned. Assert the **six approval** `WindowApp` values map into the set — not all seven: `ImportAccount` → `import-account-with-file.html` is deliberately excluded (`open-request-windows.test.ts:203-219`).

2. **`sweep-orphaned-requests.test.ts:237-255`** — "logged with identifiers only, never a raw URL" is blind whenever the leaking value is a raw `Error`: `message` and `stack` are non-enumerable, so `JSON.stringify` renders `{}`. Replace `redactUrlQuery(error)` with `error` at `sweep-orphaned-requests.ts:101` and the suite stays 12/12 green. The per-argument walk at `fail-request-on-window-error.test.ts:301-317` — in this same PR — already handles exactly this.

3. **`get-main-store.ts:216`** — the sweep's only production wiring, unobserved. `get-main-store.test.ts`'s polyfill mock (`:22-39`) supplies no `windows`, so `windows.getAll` throws, the enumeration returns `null`, and the sweep returns having cancelled nothing. Deleting the whole call block leaves both suites green (27 tests).

4. **`cancel-requests.ts:269`** — "a suppressed-source delivery failure logs at error level, not warn" (`fail-request-on-window-error.test.ts:251-274`) is already satisfied by the catch at `:258-263`, before `:269` runs. It does catch a collapse to unconditional `error`, but is blind to a collapse to unconditional `warn` — the direction that hides "the dapp was told nothing".

5. **`sweep-orphaned-requests.test.ts:205-220`** — the close path registers `delay(CANCEL_GRACE_MS)` synchronously, before the sweep's microtask resumes and registers its own, so insertion-ordered fake timers always fire it first. Only "close cancels, sweep no-ops" is ever exercised; the reverse guard (`cancel-requests.ts:109-116`) is covered only by `cancel-open-requests-on-close.test.ts:71-80`. The assertion is `toHaveBeenCalledTimes(1)`, with nothing about which path sent it.

</details>

src/background/handlers/cancel-requests.ts
ost-ptk: Kept as-is, knowingly — this is the spec's accepted residual, now stated in the code (8ea76722 adds the note at the branch you flagged). Both constraints you named are exactly why: re-opening is impossible (the tombstone replaced the row's fields) and tombstone-after-delivery re-opens the double-delivery race the optimistic mark exists to prevent. A retry channel would need a third 'cancelled-but-undelivered' state surviving restarts — more machinery than the failure warrants, given the dapp's own timeout still bounds the hang and every earlier compensating mechanism added to this design under review introduced its own critical (see the stack's history). If this residual shows up in practice we'd rather design that state deliberately in a follow-up.

src/background/handlers/cancel-requests.ts
ost-ptk: Fixed in 8ea76722: the withhold now logs `{requestId, tabId, expectedOrigin, liveOrigin, delivered}` at the moment the direct send is skipped, following the sdk-response-to-tab precedent (identifiers and origins only, never a URL). After the reset-PR's extraction the log lives inside `deliverCancelResponse`, so all three callers (open-window, sweep, reset) get the trace.

src/background/handlers/cancel-requests.ts
ost-ptk: Fixed in 8ea76722 with your wording: the default preserves the *banner policy* for the open-window call site, while its delivery is deliberately narrowed to match #1484.

src/background/handlers/sweep-orphaned-requests.ts
ost-ptk: Four of the five closed in 8ea76722: (1) the oracle test derives the six approval WindowApp values through `getUrlByWindowApp` into `REQUEST_BEARING_PATHNAMES` (ImportAccount asserted excluded); (2) the sweep logging test is now a per-argument walk incl. object values (mutation-checked: the raw-error revert fails it); (3) get-main-store's polyfill mock gained `windows.getAll` and a wiring pin, so the production call is observed; (4) the log-level test now pins both collapse directions (message text + the untouched console on each side). (5) — the reverse timer-ordering race — was deliberately left out: the grace-side guard is pinned in cancel-open-requests-on-close.test.ts, and forcing fake-timer insertion order to flip would pin scheduler trivia rather than the property.

Comp0te (approved):

PR #1521, opened 2026-08-25
fix(background): cap concurrently open requests

ost-ptk:
## Description

Hardening on top of the sweep. `sign` and `signTypedData` already refuse at `MAX_STORED_PAYLOADS = 10` before registering a request; `connect`, `switchAccount`, `signMessage` and `decryptMessage` had **no cap at all** — supersede was the only brake, and `CANCEL_GRACE_MS = 250` lets a burst register everything first. With the mirror persisting open rows, an unbounded burst would also be a session-storage write-amplification vector.

The sweep (previous PR) is the prerequisite that makes a cap safe: without it, a pinned-open row would consume a cap slot for the whole browser session, turning frozen states into a slow denial-of-approvals.

### What changed

- `MAX_OPEN_REQUESTS` beside `MAX_RESPONDED_TOMBSTONES` in `windowManagement/reducer.ts`. Steady-state concurrency is ~1–2 (one shared approval window plus the Ledger second window) — the cap only bounds the 250 ms burst.
- `windowRequestOpened` refuses the write at the cap and returns `state` unchanged (reference-identical). Open requests are never evicted, and a refused write consumes no ordinal.
- A silent reducer refusal produces no response by itself, so each of the four capless handlers gets an explicit post-dispatch re-read — the analogue of the existing `getPayload(...) == null` check. On refusal the dapp receives that method's **existing negative shape** and `openWindow` is never called.
- **Cap only — no `errorCode`.** Four of the six response shapes cannot carry one (`connectResponse`/`switchAccountResponse` are bare booleans; `signMessageResponse`/`decryptMessageResponse` are `{cancelled} | {…}`), and their `*Error` creators lack `error: true`, so the SDK would _resolve_ the dapp promise instead of rejecting (the WALLET-1346 masking defect). Widening the wire shapes is a separate ticket.
- The mirror's write cap is tightened from an undocumented headroom to the enforced relation `MAX_SESSION_ROWS = MAX_RESPONDED_TOMBSTONES + MAX_OPEN_REQUESTS`, with a test asserting it.
- `sign`/`signTypedData` (second commit): the same post-dispatch re-read applies to them too — at the cap they refuse with their own `MAX_STORED_PAYLOADS`-style response (`errorCode: tooManyPendingRequests`, a real error channel) and never open a window. Their payload actions are dispatched before `windowRequestOpened`, so on refusal the branch also dispatches `windowRequestResponded({requestId})` — the WALLET-1418 orphan mechanism the vault reducer already keys off — reclaiming the stored payload immediately (the `windowManagement` case no-ops for an id with no open row). Without this, ~20 capless opens plus 10 refused signs would pin every payload slot and kill signing for **all** origins until the next unlock (review finding). The `MAX_STORED_PAYLOADS` refusal itself fires earlier and is byte-unchanged.
- The four capless methods still answer a cap refusal with their negative shapes only — an explicit `errorCode` for them is [WALLET-1436](https://make-software.atlassian.net/browse/WALLET-1436) (wire-shape change + the WALLET-1346 masking fix).

  On the number itself: `MAX_OPEN_REQUESTS = 20` is not arbitrary — the vault already admits `2 × MAX_STORED_PAYLOADS` (deploys + EIP-712 maps, 10 each) concurrently open sign-family requests, so any smaller cap would fire _before_ the payload caps in the all-sign-family case. That floor is necessary, not sufficient — the four capless methods consume open slots outside its quantifier (review finding; the comment and test are scoped accordingly).

### Verification

- `npx jest src/background/` — 936 tests pass; `npx tsc --noEmit` clean; eslint/prettier clean.
- `windowManagement/reducer.ts` stays at 100/100/100/100; handlers/sagas floors hold.
- Pinned by tests: at the cap the write is refused and state returns reference-identical; each refusal produces the method's negative response and no `openWindow` call; a completed request frees a slot; `sign`'s existing refusal still fires first; the `MAX_SESSION_ROWS` relation.

## Linked tickets

## Checklist

[WALLET-1436]: https://make-software.atlassian.net/browse/WALLET-1436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -306,6 +306,28 @@ export async function handleSdkMethod(
         method: 'sign'
       })
     );
+
+    // At `MAX_OPEN_REQUESTS` the reducer refused the write silently, same as
+    // the four capless methods above — but the payload above has ALREADY been
+    // accepted into the map, so it is stranded here rather than refused; it is
+    // reclaimed by `reconcileStalePayloadsSaga` since no descriptor and no
+    // window will ever claim it.
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
+      reportOpenRequestCapacityRefusal(action);
+
+      return {
+        handled: true,
+        response: sdkMethod.signResponse(
+          {
+            cancelled: true,
+            message: CAPACITY_REFUSAL_MESSAGE,
+            errorCode: SdkErrorCode.tooManyPendingRequests
+          },
+          { requestId: action.meta.requestId }
+        )
+      };
+    }
+
     openWindow(store, {
       windowApp: WindowApp.SignatureRequestDeploy,
       searchParams: {
@@ -423,6 +445,29 @@ export async function handleSdkMethod(
         method: 'signTypedData'
       })
     );
+
+    // Same residual as the deploy branch: the payload above is already
+    // accepted, so a refusal here strands it for `reconcileStalePayloadsSaga`
+    // to reclaim rather than refusing the payload write itself.
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
+      reportOpenRequestCapacityRefusal(action);
+
+      return {
+        handled: true,
+        response: sdkMethod.signTypedDataResponse(
+          {
+            cancelled: true,
+            signature: null,
+            digest: null,
+            publicKey: null,
+            error: CAPACITY_REFUSAL_MESSAGE,
+            errorCode: SdkErrorCode.tooManyPendingRequests
+          },
+          { requestId: action.meta.requestId }
+        )
+      };
+    }
+
     openWindow(store, {
       windowApp: WindowApp.SignatureRequestEip712,
       searchParams: {
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -69,6 +69,15 @@ function reportCapacityRefusal(action: SdkMethod) {
   );
 }
 
+// Same rationale as `reportCapacityRefusal`, for `MAX_OPEN_REQUESTS` rather
+// than the payload maps.
+function reportOpenRequestCapacityRefusal(action: SdkMethod) {
+  console.error('sdk-methods: open-request map at capacity, request refused', {
+    requestId: action.meta.requestId,
+    method: action.type
+  });
+}
+
 export async function handleSdkMethod(
   action: SdkMethod,
   sender: Runtime.MessageSender,
@@ -140,6 +149,21 @@ export async function handleSdkMethod(
           method: 'connect'
         })
       );
+
+      // At `MAX_OPEN_REQUESTS` the reducer refused the write silently; this is
+      // the analogue of the `getPayload(...) == null` check above, for a method
+      // with no capacity map of its own to read back.
+      if (
+        selectRequestStatus(store.getState(), action.meta.requestId) == null
+      ) {
+        reportOpenRequestCapacityRefusal(action);
+
+        return {
+          handled: true,
+          response: sdkMethod.connectResponse(false, action.meta)
+        };
+      }
+
       openWindow(store, {
         windowApp: WindowApp.ConnectToApp,
         searchParams: query,
@@ -178,6 +202,16 @@ export async function handleSdkMethod(
         method: 'switchAccount'
       })
     );
+
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
+      reportOpenRequestCapacityRefusal(action);
+
+      return {
+        handled: true,
+        response: sdkMethod.switchAccountResponse(false, action.meta)
+      };
+    }
+
     openWindow(store, {
       windowApp: WindowApp.SwitchAccount,
       searchParams: query,
@@ -307,6 +341,19 @@ export async function handleSdkMethod(
         method: 'signMessage'
       })
     );
+
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
+      reportOpenRequestCapacityRefusal(action);
+
+      return {
+        handled: true,
+        response: sdkMethod.signMessageResponse(
+          { cancelled: true },
+          action.meta
+        )
+      };
+    }
+
     openWindow(store, {
       windowApp: WindowApp.SignatureRequestMessage,
       searchParams: {
@@ -412,6 +459,19 @@ export async function handleSdkMethod(
         method: 'decryptMessage'
       })
     );
+
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
+      reportOpenRequestCapacityRefusal(action);
+
+      return {
+        handled: true,
+        response: sdkMethod.decryptMessageResponse(
+          { cancelled: true },
+          action.meta
+        )
+      };
+    }
+
     openWindow(store, {
       windowApp: WindowApp.DecryptMessageRequest,
       searchParams: {
--- a/src/background/handlers/sdk-methods.ts
+++ b/src/background/handlers/sdk-methods.ts
@@ -28,9 +28,15 @@ import {
   selectIsAccountConnected,
   selectVaultActiveAccount
 } from '@background/redux/vault/selectors';
-import { windowRequestOpened } from '@background/redux/windowManagement/actions';
+import {
+  windowRequestOpened,
+  windowRequestResponded
+} from '@background/redux/windowManagement/actions';
 import { isStorableRequestId } from '@background/redux/windowManagement/request-map';
-import { selectRequestStatus } from '@background/redux/windowManagement/selectors';
+import {
+  selectOpenRequests,
+  selectRequestStatus
+} from '@background/redux/windowManagement/selectors';
 import { emitSdkEventToActiveTabsWithOrigin } from '@background/utils';
 
 import { SiteNotConnectedError, WalletLockedError } from '@content/sdk-errors';
@@ -69,12 +75,21 @@ function reportCapacityRefusal(action: SdkMethod) {
   );
 }
 
-// Same rationale as `reportCapacityRefusal`, for `MAX_OPEN_REQUESTS` rather
-// than the payload maps.
-function reportOpenRequestCapacityRefusal(action: SdkMethod) {
+// Same LOGGING rationale as `reportCapacityRefusal` (this side is not
+// invisible to the wallet even when refused silently) — but not the same dapp
+// half: only the two payload-bearing refusals (`sign`, `signTypedData`) carry
+// `errorCode`. For the four capless methods the refusal is indistinguishable
+// from an ordinary user decline on the dapp side until WALLET-1436 gives it
+// one too. `openCount` is the count AFTER the refused write — it names how
+// full the map the refusal fired against actually was.
+function reportOpenRequestCapacityRefusal(
+  action: SdkMethod,
+  openCount: number
+) {
   console.error('sdk-methods: open-request map at capacity, request refused', {
     requestId: action.meta.requestId,
-    method: action.type
+    method: action.type,
+    openCount
   });
 }
 
@@ -156,7 +171,10 @@ export async function handleSdkMethod(
       if (
         selectRequestStatus(store.getState(), action.meta.requestId) == null
       ) {
-        reportOpenRequestCapacityRefusal(action);
+        reportOpenRequestCapacityRefusal(
+          action,
+          selectOpenRequests(store.getState()).length
+        );
 
         return {
           handled: true,
@@ -204,7 +222,10 @@ export async function handleSdkMethod(
     );
 
     if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
-      reportOpenRequestCapacityRefusal(action);
+      reportOpenRequestCapacityRefusal(
+        action,
+        selectOpenRequests(store.getState()).length
+      );
 
       return {
         handled: true,
@@ -311,9 +332,19 @@ export async function handleSdkMethod(
     // the four capless methods above — but the payload above has ALREADY been
     // accepted into the map, so it is stranded here rather than refused; it is
     // reclaimed by `reconcileStalePayloadsSaga` since no descriptor and no
-    // window will ever claim it.
+    // window will ever claim it. `windowRequestResponded` reclaims it
+    // immediately instead: the vault reducer deletes the payload keyed off
+    // this exact action (the WALLET-1418 orphan mechanism, and it is in the
+    // re-encrypt debounce list so the deletion reaches the cipher), and the
+    // windowManagement case no-ops for an id with no open row.
     if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
-      reportOpenRequestCapacityRefusal(action);
+      reportOpenRequestCapacityRefusal(
+        action,
+        selectOpenRequests(store.getState()).length
+      );
+      store.dispatch(
+        windowRequestResponded({ requestId: action.meta.requestId })
+      );
 
       return {
         handled: true,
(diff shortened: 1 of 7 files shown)


src/background/handlers/sdk-methods.ts
```diff
@@ -272,6 +306,28 @@ export async function handleSdkMethod(
         method: 'sign'
       })
     );
+
+    // At `MAX_OPEN_REQUESTS` the reducer refused the write silently, same as
+    // the four capless methods above — but the payload above has ALREADY been
+    // accepted into the map, so it is stranded here rather than refused; it is
+    // reclaimed by `reconcileStalePayloadsSaga` since no descriptor and no
+    // window will ever claim it.
+    if (selectRequestStatus(store.getState(), action.meta.requestId) == null) {
```
Comp0te, line 340: This branch returns without deleting the payload stored at :269-274, and nothing else can: all four `windowRequestResponded` dispatchers need a live `'open'` row, which the refusal guarantees is absent. The PR body's "~20 concurrently open requests" premise doesn't hold either — the four capless methods fill the cap without touching a payload map. Same branch at :452.

Dispatch `windowRequestResponded({ requestId: action.meta.requestId })` in both refusal branches before returning — `vault/reducer.ts:734-746` is keyed off the action precisely for this orphan case, and the `windowManagement` case no-ops for an id with no open row.

**No reclaim route.** The four non-test dispatchers each need a live `'open'` row: `cancel-requests.ts:120` (from `selectOpenRequests`), `:283` (early-returns at :279-281), `close-windows-on-response.ts:42` (guarded at :29-31), `vault-sagas.ts:552` — the last reachable only via `takeLatest(vaultLoaded.type)` at `:134`, and `vaultLoaded` is put from one place, `unlockVaultSaga:605`. The action is excluded from `FORWARDED_ACTION_TYPES` (`redux-actions.parity.test.ts:121-125`), so no UI route. `sweep-orphaned-requests.ts:30-35` walks only `status === 'open'` rows.

**Reachability.** `signMessage` (:343-365) has no lock gate and no connected-site check and stores no payload; same for `switchAccount` (:196-204), `decryptMessage` (:498-506), `connect` (:137-151). 20 of those fill `openCount`, then each `sign` stores its payload and is refused here. At 10 stored, `vault/reducer.ts:116-121` refuses every later write with no origin check — signing fails for *every* origin, including another tab's dapp, until lock/unlock.

**The regression.** Base `windowManagement/reducer.ts:64-87` has no cap, so the descriptor was always written and supersede reached `windowRequestResponded`, which `vault/reducer.ts:733-761` acts on to delete the payload.

</details>

src/background/redux/windowManagement/reducer.ts
```diff
@@ -9,6 +9,14 @@ import { CancellableMethod, WindowManagementState } from './types';
 // background page never restarts.
 export const MAX_RESPONDED_TOMBSTONES = 50;
 
+// One shared approval window plus the Ledger second window keeps steady-state
+// open-request concurrency at ~1-2; this only needs to bound the 250 ms
+// `CANCEL_GRACE_MS` burst, not any sustained load. Binding floor: this must
+// stay >= 2 * `MAX_STORED_PAYLOADS` (vault/reducer.ts) — below that, this cap
+// would fire before either payload cap on the sign/signTypedData paths, and
+// the payload already accepted by then is stranded with no window to show it.
+export const MAX_OPEN_REQUESTS = 20;
```
Comp0te, line 23: The floor is necessary but not sufficient, and both this comment and `session-store.test.ts:422-427` read as sufficient. Its derivation quantifies only over sign-family traffic: `connect`, `switchAccount`, `signMessage` and `decryptMessage` each take an open slot and no payload slot, so 20 open `signMessage` rows put the next `sign` at the cap with both payload maps empty.

`sdk-methods.ts:310-314` in this PR already concedes the strand happens — so either fix the strand and drop the floor rationale, or say here that the floor bounds only the all-sign-family case.

The conditional as written is true and the invariant is real: at `MAX_OPEN_REQUESTS = 19`, 10 `sign` + 9 `signTypedData` reach the open cap with `eip712ById` at 9, so the 10th eip712 payload is accepted (`vault/reducer.ts:116-121` passes at 9 < 10) and stranded. The gap is "binding" and the test's name — both claim a sufficiency the derivation doesn't establish, because the capless methods consume open slots outside its quantifier.

An assertion that can fail: fill the map with 20 non-payload-bearing `windowRequestOpened` actions through the real reducer, dispatch a `sign`, assert on the payload map.

</details>

src/background/redux/windowManagement/reducer.ts
```diff
@@ -68,6 +76,17 @@ const slice = createSlice({
         return state;
       }
 
+      // At the cap, refuse the write rather than evict — "open requests are
+      // never evicted" is the standing invariant, and a refused write must
+      // consume no ordinal (below `nextSeq` is never reached).
+      const openCount = Object.values(state.requests).filter(
```
Comp0te, line 87: `openCount` is one global tally with no per-origin dimension, though `origin` is right there in the payload at :62-70. Any page can therefore spend another origin's approval capacity: the victim's `connectRequest` returns `connectResponse(false)` at `sdk-methods.ts:163` without reaching `openWindow`. A windowless `false` was already reachable pre-PR via `failRequestOnWindowError`; what's new is that a remote origin triggers it on demand.

Count open requests per `origin` alongside the global bound, so one page cannot spend another's capacity.

The content script matches `http://*/*` and `https://*/*` with `all_frames: true` (`manifest.v3.json:45-51`), so the consuming page is any site or embedded cross-origin frame, not only a connected dapp. The pre-PR windowless route was `failRequestOnWindowError` (`cancel-requests.ts:270-289`) → `deliverCancelResponse` (:209-252), triggered by a `windows.create` rejection at `open-window.ts:50-56` — a wallet-side failure, not something another page could cause.

Saturation is at least visible, and more so than "one shared window" suggests: in a cold burst all 20 `openWindow` calls read `selectWindowId` as `null` (`open-window.ts:64-69`) before `windowIdChanged` lands, so none is a reuse, `cancelRequestsDisplacedBy` never runs, and 20 popups open.

`reducer.test.ts:31-37` hardcodes `origin: 'https://dapp'` and `method: 'sign'`, and all five cap tests at :536-601 build rows from it, so nothing pins mixed-origin behaviour.

</details>

src/background/handlers/sdk-methods.ts
```diff
@@ -140,6 +149,21 @@ export async function handleSdkMethod(
...
+      // the analogue of the `getPayload(...) == null` check above, for a method
+      // with no capacity map of its own to read back.
+      if (
+        selectRequestStatus(store.getState(), action.meta.requestId) == null
+      ) {
+        reportOpenRequestCapacityRefusal(action);
+
+        return {
+          handled: true,
+          response: sdkMethod.connectResponse(false, action.meta)
```
Comp0te, line 181: This `false` is byte-identical to `buildCancelResponse`'s user-cancel at `cancel-requests.ts:49`, and `fetchFromBackground` (`content/sdk.ts:112-116`) rejects only on a top-level `error`, so the dapp promise resolves with the user-rejection value. Four of this file's eight refusal sites are now like this — which breaks the rationale :72-73 borrows from `reportCapacityRefusal`, whose premise opens "the dapp half of a capacity refusal is `errorCode`".

Correct that comment here. The wire shape is WALLET-1436's, but `content/sdk.ts`'s `@returns` at :245 and :321 are undeferred and worth folding in — flagged below with what's already true on the base.

The eight refusal sites: `:285` and `:421` carry `errorCode` (payload cap), as do `:316` and `:453` (open cap). The other four answer `connectResponse(false)` :163, `switchAccountResponse(false)` :211, and `{ cancelled: true }` at :372-375 and :513-516 — matching `buildCancelResponse` at `cancel-requests.ts:49`, `:51`, `:55`, `:68` exactly.

On the JSDoc: `content/sdk.ts` is unchanged by this PR, and `:245`/`:321` were already incomplete on the base, since a windowless `{ cancelled: true }` was reachable through `failRequestOnWindowError`. What changes is the reach — that refusal became one any page can trigger on demand, so the `@returns` gap now matters where before it described a wallet-side window-open failure.

Raising the refusal log to carry the current open count would also make the refusal diagnosable rather than merely present.

</details>

src/background/redux/windowManagement/session-store.test.ts
```diff
@@ -409,6 +412,20 @@ describe('session-store — write path', () => {
     expect(lastWrittenRecord().windowId).toBe(2);
   });
 
+  it('bounds the write cap to exactly what the two request states can hold', () => {
+    // A row is either a tombstone or an open request — no third state — so the
+    // sum of their individual caps is the only correct figure, not headroom
+    // picked separately from the reducer's own open-request bound.
+    expect(MAX_SESSION_ROWS).toBe(MAX_RESPONDED_TOMBSTONES + MAX_OPEN_REQUESTS);
```
Comp0te: This is character-for-character `session-store.ts:31`'s own definition, so it cannot fail while that expression stands, and fails for any deliberate change to it — including a correct one adding headroom. The property it stands in for is untested: `:429-453` and `:455-471` both build states strictly over the cap, and nothing constructs one at exactly 50 + 20.

Replace it with the boundary case: build 50 tombstones and 20 open rows, run `capRows`, assert nothing was dropped.

`session-store.ts:31` is `export const MAX_SESSION_ROWS = MAX_RESPONDED_TOMBSTONES + MAX_OPEN_REQUESTS;` and this line is `expect(MAX_SESSION_ROWS).toBe(MAX_RESPONDED_TOMBSTONES + MAX_OPEN_REQUESTS);` — same two imported symbols (:3, :5), same operator.

The behavioural cap tests use `MAX_RESPONDED_TOMBSTONES + 22` (72) and `+ 21` (71) against a cap of 70; the only other `MAX_SESSION_ROWS` consumer, `sweep-orphaned-requests.test.ts:225`, is also over. The under-cap early return at `session-store.ts:186-188` is exercised by the small-map write tests, so a guard off-by-one would be caught — what nothing catches is the caps drifting so the reducer can hold more rows than the mirror writes, after which an open row is silently dropped and the restart path loses it.

</details>

src/background/redux/windowManagement/reducer.ts
```diff
@@ -68,6 +76,17 @@ const slice = createSlice({
...
+        request => request?.status === 'open'
+      ).length;
+
+      if (openCount >= MAX_OPEN_REQUESTS) {
```
Comp0te, line 91: Separate from the per-origin point above: this refusal counts hydrated session rows the same as live ones. `get-main-store.ts:136-144` injects the mirror as preloaded state, bypassing the reducer, and the read side never applies `capRows`. The sweep that would cancel stale rows is fire-and-forget at :216-222 and reaches `failRequestOnWindowError` only after `await delay(CANCEL_GRACE_MS)`, while `background/index.ts:208` serves the first request immediately.

Could the cap count only what this worker session registered — or the sweep's cancellation pass drain before the first approval request is served?

Ordering, end to end: `readRequestSession` (`session-store.ts:210-236`) sanitises but never caps, since `capRows` runs only from `flush` (:258-263); the sweep is dispatched at `get-main-store.ts:216-222`, before `return storeSingleton` at :228, and reaches `failRequestOnWindowError` at `sweep-orphaned-requests.ts:89-95` only after `await collectRequestIdsFromOpenWindows()` (:50) and `await delay(CANCEL_GRACE_MS)` (:87); meanwhile the listener at `background/index.ts:206-229` awaits the store at :208 and calls `handleSdkMethod` at :225.

So for at least the grace period after a wake, `openCount` includes debris the sweep hasn't cancelled, and the request that woke the worker can be refused and logged at `sdk-methods.ts:74-79` as "open-request map at capacity" — true of the tally, misleading about the cause. How often a worker dies holding 20 open rows in the mirror isn't something I settled; the ordering is.

</details>

Comp0te (commented):
Read the cap end to end — the reducer's refusal path, the six handler read-backs, the session mirror, and the tests — against the merge base. The cap itself is the right shape and the ordinal/no-evict invariants hold. The comments below run along one seam: the refusal is a silent state transition, and the code around it hasn't caught up. One path strands an already-accepted payload; two texts assert the strand can't happen; two tests restate definitions rather than exercise the boundary; and the count that drives the refusal is blind to both origin and provenance.

src/background/handlers/sdk-methods.ts
ost-ptk: Accepted and fixed in ea8f4cf7 exactly as you suggested: both refusal branches dispatch `windowRequestResponded({requestId})` before returning — the vault reducer reclaims the payload off that action (and it is in the re-encrypt debounce list, so the deletion reaches the cipher), while the windowManagement case no-ops for an id with no open row. Tests assert the dispatch on both branches. The PR body's ~20-non-sign premise was also corrected — you're right that the capless methods fill the cap without touching a payload map.

src/background/redux/windowManagement/reducer.ts
ost-ptk: Fixed in ea8f4cf7, your first option plus the scope note: the strand itself is fixed (see the :315 thread), and the floor comment + test are now explicitly scoped to the all-sign-family case, naming the four capless methods as outside the quantifier.

src/background/handlers/sdk-methods.ts
ost-ptk: Fixed in ea8f4cf7: the comment no longer borrows the errorCode premise (it now names the four decline-indistinguishable shapes and WALLET-1436), the refusal log carries the current open count, and `content/sdk.ts`'s two `@returns` note that a negative value can also mean a wallet-side capacity refusal.

src/background/redux/windowManagement/session-store.test.ts
ost-ptk: Replaced in ea8f4cf7 with the boundary case you specified: exactly MAX_RESPONDED_TOMBSTONES tombstones + MAX_OPEN_REQUESTS open rows through `writeRequestSession`, asserting nothing is dropped. The >= 2xMAX_STORED_PAYLOADS floor test stays as the (now correctly scoped) tripwire.

src/background/redux/windowManagement/reducer.ts
ost-ptk: Kept as-is, with the trade-off stated: draining the sweep before serving the first request adds a grace-plus-enumeration latency to EVERY wake to buy correctness only in an already-adversarial corner (a worker dying while holding 20 mirrored open rows), and counting only current-session registrations reopens unbounded growth across restarts. The state self-heals within roughly one sweep pass, and the refusal log now carries the live open count (ea8f4cf7), so the 'at capacity right after a wake' case is diagnosable rather than misleading.

src/background/redux/windowManagement/reducer.ts
ost-ptk: Agreed on the substance; deferring to its own ticket rather than this stack, because the fix is not mechanical: MAX_STORED_PAYLOADS is itself global (2x10), so one dapp can legitimately hold 20 sign-family opens — a per-origin cap below 20 breaks that legal case, and per-origin == global adds nothing. Meaningful isolation needs both caps re-derived together (and possibly per-origin payload budgets). Will link the ticket here once filed.

ost-ptk: Filed as [WALLET-1437](https://make-software.atlassian.net/browse/WALLET-1437) (Low): per-origin isolation for approval-request capacity — the open-request cap and the two payload caps re-derived together, with mixed-origin tests. The ticket records the arithmetic from this thread (per-origin < 20 breaks the legal single-dapp sign burst; per-origin == global adds nothing).

Comp0te (approved):

